PR #2, opened 2024-10-17
Create Genie API wrapper

prithvikannan:
Create the core Genie API wrapper. The wrapper API follows the [Genie convseration APIs ](https://drive.google.com/a/databricks.com/open?id=1377caJzV6T5fXSjkt0DXGNmfDizPEe-Fs-2MjhSGzms)for polling. 

Unit tested.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.9'
+          python-version: '3.8'
       - name: Install dependencies
         run: |
           pip install -r requirements/lint-requirements.txt
@@ -35,7 +35,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.9', '3.10']
+        python-version: ['3.8', '3.9', '3.10']
     timeout-minutes: 20
     steps:
       - name: Checkout code
--- a/README.md
+++ b/README.md
@@ -1,1 +1,1 @@
-# databricks-ai
+# databricks-ai-bridge


pyproject.toml
prithvikannan, line 1: downgrading to match other packages like mlflow

src/databricks_ai_bridge/genie.py
```diff
@@ -0,0 +1,116 @@
...
+        self.space_id = space_id
+        workspace_client = WorkspaceClient()
+        self.genie = workspace_client.genie
+        self.headers = {
+            "Accept": "application/json",
+            "Content-Type": "application/json",
+        }
+
+    def start_conversation(self, content):
+        resp = self.genie._api.do(
```
aravind-segu, line 59: question, why are we using the raw `_api`  request here? Can we not use this method directly? https://databricks-sdk-py.readthedocs.io/en/stable/workspace/dashboards/genie.html#databricks.sdk.service.dashboards.GenieAPI.start_conversation

src/databricks_ai_bridge/genie.py
```diff
@@ -0,0 +1,116 @@
...
+            f"/api/2.0/genie/spaces/{self.space_id}/conversations/{conversation_id}/messages",
+            body={"content": content},
+            headers=self.headers,
+        )
+        return resp
+
+    def poll_for_result(self, conversation_id, message_id):
+        def poll_result():
+            while True:
+                resp = self.genie._api.do(
```
aravind-segu, line 79: Same thing here: https://databricks-sdk-py.readthedocs.io/en/stable/workspace/dashboards/genie.html#databricks.sdk.service.dashboards.GenieAPI.get_message

That way we can use the enums of status instead of using text here.

src/databricks_ai_bridge/genie.py
```diff
@@ -0,0 +1,116 @@
...
+                    "GET",
+                    f"/api/2.0/genie/spaces/{self.space_id}/conversations/{conversation_id}/messages/{message_id}",
+                    headers=self.headers,
+                )
+                if resp["status"] == "EXECUTING_QUERY":
+                    sql = next(r for r in resp["attachments"] if "query" in r)["query"]["query"]
```
aravind-segu, line 85: Are we getting the sql statements here only for the debug statement? Do end customers need this information?

src/databricks_ai_bridge/genie.py
```diff
@@ -0,0 +1,116 @@
...
+                elif state == "RUNNING" or state == "PENDING":
+                    logging.debug("Waiting for query result...")
+                    time.sleep(5)
+                else:
+                    logging.debug(f"No query result: {resp['state']}")
+                    return None
+
+        return poll_result()
+
+    def ask_question(self, question):
```
aravind-segu, line 113: What do you think of adding a timeout feature here, so users are not stuck waiting forever in the polling loop?

src/databricks_ai_bridge/genie.py
```diff
@@ -0,0 +1,116 @@
...
+                    "GET",
+                    f"/api/2.0/genie/spaces/{self.space_id}/conversations/{conversation_id}/messages/{message_id}",
+                    headers=self.headers,
+                )
+                if resp["status"] == "EXECUTING_QUERY":
+                    sql = next(r for r in resp["attachments"] if "query" in r)["query"]["query"]
+                    logging.debug(f"SQL: {sql}")
+                    return poll_query_results()
+                elif resp["status"] == "COMPLETED":
+                    return next(r for r in resp["attachments"] if "text" in r)["text"]["content"]
```
aravind-segu, line 89: For genie, do we only care about the first response in the list?

tests/databricks_ai_bridge/test_genie.py
```diff
@@ -0,0 +1,141 @@
...
+    assert result == pd.DataFrame().to_string()
+
+
+def test_ask_question(genie, mock_workspace_client):
+    mock_workspace_client.genie._api.do.side_effect = [
+        {"conversation_id": "123", "message_id": "456"},
+        {"status": "COMPLETED", "attachments": [{"text": {"content": "Answer"}}]},
+    ]
+    result = genie.ask_question("What is the meaning of life?")
+    assert result == "Answer"
```
aravind-segu, line 77: The answer cant be that simple 🤣

src/databricks_ai_bridge/genie.py
prithvikannan: eventually these should be part of the trace. currently adding extra traces with autologged traces are not supported, but this is coming soon. we'll use the SQL at that time.

src/databricks_ai_bridge/genie.py
prithvikannan: i think the polling mechanisms with the `start_conversation` API does not work. ill asked the genie team and they recommended to use the `_api` for now. we can revisit when fixed.

src/databricks_ai_bridge/genie.py
prithvikannan: at this time there's only one text attachment from genie so this is a safe assumption.

src/databricks_ai_bridge/genie.py
prithvikannan: good question. i wonder if genie requests have some RPC level timeout on that side, but i think a client side timeout also makes sense. will update

prithvikannan: actually genie has a `CANCELED` state that happens in the case of timeout, which will hit the else case and break the loop. i dont think we need the client side timeout.

PR #5, opened 2024-10-22
Create Langchain Genie

prithvikannan:
Create a `GenieAgent` in `databricks_langchain`

### DIFF
--- a/integrations/langchain/src/databricks_langchain/genie.py
+++ b/integrations/langchain/src/databricks_langchain/genie.py
@@ -32,7 +32,7 @@ def _query_genie_as_agent(input, genie_space_id, genie_agent_name):
         return {"messages": [AIMessage(content="")]}
 
 
-def create_genie_agent(genie_space_id, genie_agent_name="Genie"):
+def GenieAgent(genie_space_id, genie_agent_name="Genie", description=""):
     """Create a genie agent that can be used to query the API"""
     from functools import partial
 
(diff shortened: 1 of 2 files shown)


integrations/langchain/src/databricks_langchain/genie.py
```diff
@@ -0,0 +1,47 @@
...
+    # Send the message and wait for a response
+    genie_response = genie.ask_question(message)
+
+    if genie_response:
+        return {"messages": [AIMessage(content=genie_response)]}
+    else:
+        return {"messages": [AIMessage(content="")]}
+
+
+def GenieAgent(genie_space_id, genie_agent_name="Genie", description=""):
```
prithvikannan, line 35: adding `description` to the interface, but TBD on how this will be used.

integrations/langchain/src/databricks_langchain/genie.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+    return concatenated_message
+
+
+def _query_genie_as_agent(input, genie_space_id, genie_agent_name):
+    from langchain_core.messages import AIMessage
+
+    genie = Genie(genie_space_id)
+
+    message = f"I will provide you a chat history, where your name is {genie_agent_name}. Please help with the described information in the chat history.\n"
```
aravind-segu, line 21: Question here: Are we expecting users to use the genie agent name in their questions? Will the questions be something like:

"Use Genie to get me my financial records"

Will this also work when users ask "Get my financial records"

prithvikannan: The ideal end state is that the user can just specify  "Get my financial records". This name is needed so that in a multi-agent (and in the future multi-user) chat array agents can identify who said what.

PR #7, opened 2024-10-24
Update READMEs

prithvikannan:
Update READMEs for `databricks-ai-bridge` and `langchain-databricks`. Currently only scoped to Genie features, but we can extend in the future.

### DIFF
--- a/integrations/langchain/README.md
+++ b/integrations/langchain/README.md
@@ -14,7 +14,10 @@ pip install git+ssh://git@github.com/databricks/databricks-ai-bridge.git#subdire
 
 ## Get started
 
-### Use a Genie space as an agent
+### (Preview) Use a Genie space as an agent
+
+> [!NOTE]
+> Requires Genie API Private Preview. Reach out to your account team for enablement. 
 
 ```python
 from databricks_langchain.genie import GenieAgent


README.md
```diff
@@ -1 +1,14 @@
-# databricks-ai-bridge
+# Databricks AI Bridge library
+
+The Databricks AI Bridge library provides a shared layer of APIs to interact with Databricks AI features, such as [Databricks AI/BI Genie ](https://www.databricks.com/product/ai-bi/genie).
+
+## Installation
+
+```sh
+# install from the source
+pip install git+ssh://git@github.com/databricks/databricks-ai-bridge.git
```
prithvikannan, line 9: note: not possible currently bc the repo is not open source yet

integrations/langchain/README.md
```diff
@@ -0,0 +1,23 @@
...
+# install from the source
+pip install git+ssh://git@github.com/databricks/databricks-ai-bridge.git#subdirectory=integrations/langchain
+```
+
+> [!NOTE]
+> Once this package is published to PyPI, users can install via `pip install databricks-langchain`
+
+## Get started
+
+### Use a Genie space as an agent
```
smurching: Eventually we should flesh this out a bit more - guessing we'd have an example in the actual LangChain Databricks official docs? Also we may want to document somewhere that the API is experimental and requires Genie API PrPr access

smurching: > Also we may want to document somewhere that the API is experimental and requires Genie API PrPr access

(fine to do that in a followup PR)

PR #24, opened 2024-12-05
Add new Databricks Vector Search langchain native tool VectorSearchRetrieverTool

leonbi100:
# What does this PR do?
This PR introduces the class `VectorSearchRetrieverTool` into the `databricks-langchain` package which allows the user to instantiate a langchain native tool that calls Databricks Vector Search when invoked. Under the hood, the core logic is from `DatabricksVectorSearch` in the `langchain-databricks` package.

# How was it tested?
* New unit tests in `test_vector_search_retriever_tool.py` which verifies the new tool object can be instantiated correctly and can be invoked by an arbitrary langchain llm
* Tested e2e in notebook: image
* Can see the populated tool description when no description is provided: image


integrations/langchain/src/databricks_langchain/vector_search.py
```diff
@@ -0,0 +1,24 @@
...
+            columns: Optional[List[str]] = None,
+    ):
+        vector_store = DatabricksVectorSearch(
+            endpoint=endpoint,
+            index_name=index_name,
+            embedding = embedding,
+            text_column = text_column,
+            columns = columns
+        )
+        return vector_store.as_retriever().as_tool()
```
leonbi100, line 24: Looks like .as_retriever() has possible parameters we can pass in. Maybe we consider adding this into the `VectorSearchRetrieverTool` API itself?

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
```diff
@@ -0,0 +1,81 @@
...
+    columns: Optional[List[str]] = Field(None, description="Columns to return when doing the search.")
+    filters: Optional[Dict[str, Any]] = Field(None, description="Filters to apply to the search.")
+    query_type: str = Field("ANN", description="The type of query to run.")
+    tool_name: Optional[str] = Field(None, description="The name of the retrieval tool.")
+    tool_description: Optional[str] = Field(None, description="A description of the tool.")
+    # TODO: Confirm if we can add these two to the API to support direct-access indexes or a delta-sync indexes with self-managed embeddings,
+    text_column: Optional[str] = Field(None, description="If using a direct-access index or delta-sync index, specify the text column.")
+    embedding: Optional[Embeddings] = Field(None, description="Embedding model for self-managed embeddings.")
+    # TODO: Confirm if we can add this endpoint field
+    endpoint: Optional[str] = Field(None, description="Endpoint for DatabricksVectorSearch.")
```
leonbi100: This field was added because of [this restriction](https://github.com/databricks/databricks-ai-bridge/blob/main/integrations/langchain/src/databricks_langchain/vectorstores.py#L239-L242) in databricks-langchain. I felt that if we threw this error without giving the ability for the user to rectify it, it would be a poor user experience. Alternatively maybe we pin databricks-vectorsearch to be >=0.35.

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
```diff
@@ -0,0 +1,81 @@
...
+    index_name: str = Field(..., description="The name of the index to use, format: 'catalog.schema.index'.")
+    num_results: int = Field(10, description="The number of results to return.")
+    columns: Optional[List[str]] = Field(None, description="Columns to return when doing the search.")
+    filters: Optional[Dict[str, Any]] = Field(None, description="Filters to apply to the search.")
+    query_type: str = Field("ANN", description="The type of query to run.")
+    tool_name: Optional[str] = Field(None, description="The name of the retrieval tool.")
+    tool_description: Optional[str] = Field(None, description="A description of the tool.")
+    # TODO: Confirm if we can add these two to the API to support direct-access indexes or a delta-sync indexes with self-managed embeddings,
+    text_column: Optional[str] = Field(None, description="If using a direct-access index or delta-sync index, specify the text column.")
+    embedding: Optional[Embeddings] = Field(None, description="Embedding model for self-managed embeddings.")
```
leonbi100: These two fields are required for [direct-access indexes or delta-sync indexes with self-managed embeddings](https://github.com/databricks/databricks-ai-bridge/blob/main/integrations/langchain/src/databricks_langchain/vectorstores.py#L108-L119). Should we support these additional fields?

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
```diff
@@ -0,0 +1,81 @@
...
+            "text_column": self.text_column,
+            "columns": self.columns,
+        }
+        dbvs = DatabricksVectorSearch(**kwargs)
+        self._vector_store = dbvs
+
+        def get_tool_description():
+            default_tool_description = "A vector search-based retrieval tool for querying indexed embeddings."
+            index_details = IndexDetails(dbvs.index)
+            if index_details.is_delta_sync_index():
```
leonbi100, line 71: direct access indexes don't have an associated source table so we'll just use the default tool description.

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
annzhang-db: I feel like if we support it for DatabricksVectorSearch it makes sense to support it here.

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
annzhang-db: I think it's valid to require databricks-vectorsearch >= 0.35 especially because this is new - that might be the better considering we don't need `endpoint` for any other reason.

smurching: Yeah reasonable to require new versions of other clients!

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
smurching: Yeah, seems reasonable to support these, though I'd say it's worth asking vector search folks how commonly direct access indexes are used, if it's infrequent we could drop this to start with to simplify the API/testing surface

smurching: No need to block this PR on that though, I figure we'll need this eventually anyways, would just be good for us to know

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
smurching: Curious what the existing `langchain-databricks` DatabricksVectorSearch.as_retriever(...).as_tool(...) ends up generating as the tool description

smurching: One way to tell would be to use it as a tool with payload logging enabled & see what the `tools` argument to the LLM API in model serving looks like

smurching: This generally looks reasonable, just curious if we can keep it in sync with the existing behavior/default

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    A utility class to create a vector search-based retrieval tool for querying indexed embeddings.
+    This class integrates with a Databricks Vector Search and provides a convenient interface
+    for building a retriever tool for agents.
+    """
+
+    index_name: str = Field(..., description="The name of the index to use, format: 'catalog.schema.index'.")
+    num_results: int = Field(10, description="The number of results to return.")
+    columns: Optional[List[str]] = Field(None, description="Columns to return when doing the search.")
+    filters: Optional[Dict[str, Any]] = Field(None, description="Filters to apply to the search.")
```
smurching, line 32: QQ, does this get sent to the LLM as the parameter description? If so I wonder if it's worth including examples like the ones in https://docs.databricks.com/api/workspace/vectorsearchindexes/queryindex

smurching: Oh nvm, this is in the init, not in the tool call

smurching: But seems like there is a way we can specify the description of the params for the LLM too: https://chatgpt.com/share/6764d76f-69a0-8009-8a8f-f58977753057

smurching: See also https://python.langchain.com/docs/how_to/custom_tools/#subclass-basetool (we can use `args_schema`)

smurching (commented):
Mostly looks good! Just had some small comments

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
leonbi100: Turns out we already mark the "databricks-vectorsearch>=0.40" as a [dependency here](https://github.com/databricks/databricks-ai-bridge/blob/main/integrations/langchain/pyproject.toml#L15), so I'll just remove this argument.

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
leonbi100: image
Tested it out in a notebook. The default seems to be extremely basic and lacking in content. Does this answer your question?

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
leonbi100: Updated to include VectorSearchRetrieverToolInput as an args_schema

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
leonbi100: [Linking the slack thread I started here](https://databricks.slack.com/archives/C05BKRLUMC4/p1734703211820509)

integrations/langchain/tests/unit_tests/test_vector_search_retriever_tool.py
```diff
@@ -0,0 +1,93 @@
...
+@pytest.mark.parametrize("text_column", [None, "text"])
+@pytest.mark.parametrize("endpoint", [None, "test_endpoint"])
+def test_vector_search_retriever_tool_combinations(
+    index_name: str,
+    columns: Optional[List[str]],
+    tool_name: Optional[str],
+    tool_description: Optional[str],
+    embedding: Optional[Any],
+    text_column: Optional[str],
+    endpoint: Optional[str],
```
annzhang-db: We shouldn't need the `endpoint` argument anymore right?

annzhang-db (approved):
LGTM with one comment - we should update the tests to reflect the most recent changes. 🙇‍♀️

integrations/langchain/tests/unit_tests/test_vector_search_retriever_tool.py
leonbi100: Good catch updated!

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
smurching: Lol yep makes sense, the updated version in this PR is definitely better

integrations/langchain/tests/utils/chat_models.py
```diff
@@ -1,7 +1,8 @@
-import pytest
-
 from typing import Generator
 from unittest import mock
+
```
smurching, line 3: Should we update these tests to also assert that the tool description + args description are properly set? Lmk if it's already done and I missed it (just looking at changes since my last review rn)

smurching (commented):
Will stamp after my testing comment is addressed ,thanks Leon!

smurching (approved):

integrations/langchain/tests/utils/chat_models.py
leonbi100: Added a new test `test_vector_search_retriever_tool_description_generation` to explicitly test these changes.

PR #25, opened 2024-12-05
Inline langchain-databricks components

prithvikannan:
Inline ChatDatabricks, DatabricksEmbeddings, and DatabricksVectorSearch from `langchain-databricks` package and their corresponding unit tests.

Remove `langchain-databricks` as a dependency for `databricks-langchain`.

### DIFF
--- a/integrations/langchain/src/databricks_langchain/__init__.py
+++ b/integrations/langchain/src/databricks_langchain/__init__.py
@@ -1,8 +1,8 @@
 # Import modules from langchain-databricks
 from databricks_langchain.chat_models import ChatDatabricks
 from databricks_langchain.embeddings import DatabricksEmbeddings
-from databricks_langchain.vectorstores import DatabricksVectorSearch
 from databricks_langchain.genie import GenieAgent
+from databricks_langchain.vectorstores import DatabricksVectorSearch
 
 # Expose all integrations to users under databricks-langchain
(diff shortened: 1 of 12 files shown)


leonbi100:
@prithvikannan are we considering porting over the integration tests? I understand that's more complicated since they have their own CI/CD system

integrations/langchain/src/databricks_langchain/__init__.py
```diff
@@ -1,11 +1,8 @@
 # Import modules from langchain-databricks
```
leonbi100: Are we going to remove `langchain-databricks` as a dependency from the `pyproject.toml` file?

prithvikannan:
> @prithvikannan are we considering porting over the integration tests? I understand that's more complicated since they have their own CI/CD system

yes we do plan to. this PR is just the unit tests, but we will move integration tests next.

integrations/langchain/src/databricks_langchain/__init__.py
prithvikannan: good callout, just updated.

leonbi100: I think they're some new test issues with that change could you take a look?

leonbi100 (approved):
lgtm!

PR #27, opened 2024-12-11
Client Arguments for DatabricksVectorSearch

chutch1122:
This PR allows new instances of `DatabricksVectorSearch` to pass custom arguments to the underlying `VectorSearchClient` (imported from `databricks.vector_search.client`).

This allows users to be able to use Service Principal based authentication by passing in `service_principal_client_id` or `service_principal_client_secret` within the new `client_args` parameter.

### DIFF
--- a/integrations/langchain/src/databricks_langchain/vectorstores.py
+++ b/integrations/langchain/src/databricks_langchain/vectorstores.py
@@ -86,6 +86,10 @@ class DatabricksVectorSearch(VectorStore):
                     Make sure the text column specified is in the index.
         columns: The list of column names to get when doing the search.
                 Defaults to ``[primary_key, text_column]``.
+        client_args: Additional arguments to pass to the VectorSearchClient.
+                    Allows you to pass in values like `service_principal_client_id`
+                    and `service_principal_client_secret` for to allow for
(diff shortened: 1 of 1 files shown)


integrations/langchain/src/databricks_langchain/vectorstores.py
```diff
@@ -230,7 +235,7 @@ def __init__(
             ) from e
 
         try:
-            self.index = VectorSearchClient().get_index(
+            self.index = VectorSearchClient(**client_args).get_index(
```
prithvikannan:
```suggestion
            self.index = VectorSearchClient(**(client_args or {})).get_index(
                endpoint_name=endpoint, index_name=index_name
            )
```

prithvikannan (commented):
Thanks @chutch1122 for the contribution. I left one small comment to fix the test failures, good to merge once fixed.

integrations/langchain/src/databricks_langchain/vectorstores.py
chutch1122: Makes sense to me!

prithvikannan (approved):
LGTM

PR #28, opened 2024-12-11
Adding a python http_request wrapper to create external tools

sunishsheth2009:
Adding a python http_request wrapper to create external tools

### DIFF
--- a/src/databricks_ai_bridge/external_tool_request.py
+++ b/src/databricks_ai_bridge/external_tool_request.py
@@ -1,0 +1,57 @@
+import json as js
+from typing import Any, Dict, Optional
+
+import requests
+from databricks.sdk import WorkspaceClient
+
+from databricks_ai_bridge.utils.annotations import experimental
+
+
+@experimental
+def http_request(
+    conn: str,
+    method: str,
+    path: str,
+    *,
+    json: Optional[Any] = None,
+    headers: Optional[Dict[str, str]] = None,
+    params: Optional[Dict[str, Any]] = None,
+) -> requests.Response:
+    """
+    Makes an HTTP request to an external function through the Databricks Workspace.
+
+    Args:
+        conn (str): The connection name to use. This is required to identify the external connection.
+        method (str): The HTTP method to use (e.g., "GET", "POST"). This is required.
+        path (str): The relative path for the API endpoint. This is required.
+        json (Optional[Any]): JSON payload for the request.
+        headers (Optional[Dict[str, str]]): Additional headers for the request.
+                If not provided, only auth headers from connections would be passed.
+        params (Optional[Dict[str, Any]]): Query parameters for the request.
+
+    Returns:
+        requests.Response: The HTTP response from the external function.
+
+    Example Usage:
+        response = http_request(
+            conn="my_connection",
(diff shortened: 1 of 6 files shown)


src/databricks_ai_bridge/external_tool_request.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import Any, Dict, Optional
+
+import requests
+from databricks.sdk import WorkspaceClient
+
+from databricks_ai_bridge.utils.annotations import experimental
+
+
+@experimental
+def http_request(
```
smurching: We should think about the module and API name a little more, maybe something like `databricks_ai_bridge.external_tools.http_request`? Or even `databricks_ai_bridge.external_tools.request`? Just thinking about how this looks in code where you might mix this with generic `requests.request` calls to some public APIs

src/databricks_ai_bridge/utils/annotations.py
```diff
@@ -0,0 +1,65 @@
+import inspect
```
smurching, line 3: Are there tests for these annotations that we should also include + include a backlink to the OSS MLflow code (I'm guessing) that we copied from?

src/databricks_ai_bridge/external_tool_request.py
smurching: Another option is `databricks_ai_bridge.uc_connections.http_request` - I know we wanna present our product as external tools but what we're actually using is the UC connection to make an HTTP request in the process of defining a tool, ratehr than using a tool itself

src/databricks_ai_bridge/external_tool_request.py
```diff
@@ -0,0 +1,57 @@
...
+    conn: str,
+    method: str,
+    path: str,
+    *,
+    json: Optional[Any] = None,
+    headers: Optional[Dict[str, str]] = None,
+    params: Optional[Dict[str, Any]] = None,
+) -> requests.Response:
+    """
+    Makes an HTTP request to an external function through the Databricks Workspace.
```
smurching:
```suggestion
    Makes an HTTP request to a remote API using authentication from a Unity Catalog HTTP connection.
```

smurching (commented):
A few comments but mostly looks great, thanks @sunishsheth2009 !

src/databricks_ai_bridge/utils/annotations.py
sunishsheth2009: There are no tests, but I have added the backlink for ref :) 
good idea

src/databricks_ai_bridge/external_tool_request.py
sunishsheth2009: I have added this to `__init__.py` so for the customers what I was thinking it would be
`from databricks_ai_bridge import http_request`
What do you think? Do we need to include external_tools or un_connections here?

sunishsheth2009: Spoke offline. http_request is generic so adding the product name here would help disambiguate this. 

So changed it to `databricks_ai_bridge.external_tools.http_request`

smurching (approved):
LGTM!

PR #29, opened 2024-12-12
Fix test_chat_databricks_invoke with llama 3.3

prithvikannan:
Update test_chat_databricks_invoke with appropriate values from llama 3.3 on databricks.


prithvikannan:
Got approval on the langchain side PR https://github.com/langchain-ai/langchain-databricks/pull/46, so will merge this one too.

PR #32, opened 2024-12-12
Genie tracing

prithvikannan:
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ dependencies = [
   "pandas",
   "tiktoken>=0.8.0",
   "tabulate",
-  "mlflow-skinny",
+  "mlflow-skinny>=2.19.0",
 ]
 
 [project.license]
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
   "pandas",
   "tiktoken>=0.8.0",
   "tabulate",
+  "mlflow",
```
sunishsheth2009: Can we do mlflow-skinny? That is the one we switched to for databricks-agents as well.

pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
   "pandas",
   "tiktoken>=0.8.0",
   "tabulate",
+  "mlflow-skinny",
```
sunishsheth2009: Do we want to enforce a specific version that has tracing like 2.15 or something?

sunishsheth2009 (approved):

pyproject.toml
prithvikannan: good idea, will do min >2.19 bc we need some recent tracing fixes

PR #37, opened 2024-12-19
v0.1.1 release

prithvikannan:


integrations/langchain/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "databricks-langchain"
-version = "0.1.0.dev0"
+version = "0.1.1.dev0"
 description = "Support for Databricks AI support in LangChain"
 authors = [
     { name="Prithvi Kannan", email="prithvi.kannan@databricks.com" },
```
aravind-segu, line 6: Change the dependencies below to 0.1.1

prithvikannan: we dont need to do this bc we don't depend on anything that changed between bridge 0.1.0 and 0.1.1

aravind-segu (approved):

PR #40, opened 2025-01-03
Add support for OpenAI structured outputs with json_schema

PhilipMathieuIDEXX:
Implements a "json_schema" mode in ChatDatabricks to allow passing a JSON schema to models. Enhances existing options (function calling, JSON mode). Includes updates to test cases to reflect currently-available models and to test the new functionality. (And passes all unit tests.)

This approach is slightly less sophisticated than what the OpenAI integration uses (see https://github.com/langchain-ai/langchain/blob/ff675c11f6814e88aa338dc3668aa84d1ade3d19/libs/partners/openai/langchain_openai/chat_models/base.py#L1142). I personally don't think that the complexity makes sense right now (i.e., when enforced json schemas are not utilized by other Databricks models) but would be willing to implement it if desired by the project maintainers.

@prithvikannan happy to take feedback or comments

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -437,7 +437,9 @@ class ChatDatabricks(BaseChatModel):
         self,
         schema: Optional[Union[Dict, Type]] = None,
         *,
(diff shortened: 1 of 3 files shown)


PhilipMathieuIDEXX:
@prithvikannan I fixed the linting error, should be good to go now

PR #41, opened 2025-01-06
Pull logic out of databricks-langchain to be reused in future integration packages

leonbi100:
### What this PR does
In preparation to create new databricks integrations, this PR pulls out logic from the databricks-langchain package and places it into databricks-ai-bridge so that other integration packages can reuse this logic.

### How was it tested
Existing unit tests still pass


annzhang-db (approved):
lgtm! 😃

PR #44, opened 2025-01-15
Simplify OpenAI VectorSearchRetrieverTool tool execution function

leonbi100:
This PR refactors the OpenAI VectorSearchRetrieverTool#execute_tools function to be VectorSearchRetrieverTool#execute. The goal here is to do less of the response parsing and instead only focus on executing the retriever call, which is more in line with [how the OpenAI SDK documents function calling](https://platform.openai.com/docs/guides/function-calling). 

The new usage pattern is more along the lines of:
```
dbvs_tool = VectorSearchRetrieverTool("index_name")
tools = [dbvs_tool.tool, ...]
response = openai.chat.completions.create(
  model="gpt-4o",
  messages=initial_messages,
  tools=tools,
)
tool_call = response.choices[0].message.tool_calls[0]
args = json.loads(tool_call.function.arguments)
retriever_resp = vector_search_tool.execute(query=args["query"])
```

### DIFF
--- a/integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
+++ b/integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
@@ -27,16 +27,36 @@ class VectorSearchRetrieverTool(VectorSearchRetrieverToolMixin):
     for tool calling using the OpenAI SDK.
 
     Example:
-        dbvs_tool = VectorSearchRetrieverTool("index_name")
-        tools = [dbvs_tool.tool, ...]
-        response = openai.chat.completions.create(
-            model="gpt-4o",
-            messages=initial_messages,
-            tools=tools,
+        messages = [
+                {"role": "system", "content": "You are a helpful assistant."},
+                {
+                    "role": "user",
+                    "content": "Using the Databricks documentation, answer what is Spark?"
+                }
+            ]
+
+        dbvs_tool = VectorSearchRetrieverTool(index_name=index_name, text_column="text", embedding_model_name=embedding_endpoint)
+        tools = [dbvs_tool.tool]
+        response = client.chat.completions.create(
+            model="gpt-4o-mini",
+            messages=messages,
+            tools=tools
         )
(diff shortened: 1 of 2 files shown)


integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -65,136 +87,103 @@ def _validate_tool_inputs(self):
...
+            raise ValueError(
+                "The embedding model name is required for non-Databricks-managed "
+                "embeddings Vector Search indexes in order to generate embeddings for retrieval queries."
+            )
+
         # OpenAI tool names must match the pattern '^[a-zA-Z0-9_-]+$'."
         # The '.' from the index name are not allowed
-        def rewrite_index_name(index_name: str):
-            return index_name.replace(".", "_")
+        def get_tool_name():
```
leonbi100, line 113: [Refactoring this to match how the UC Function Toolkit does this here](https://github.com/unitycatalog/unitycatalog/blob/main/ai/core/src/unitycatalog/ai/core/utils/function_processing_utils.py#L101)

smurching, line 113: We might want to make this into a public API so that we can just officially recommend customers use it to get the name of the vector search tool, but ok to do that in a follow-up

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -65,136 +87,103 @@ def _validate_tool_inputs(self):
...
+            tool_name = self.tool_name or self.index_name.replace(".", "__")
+            if len(tool_name) > 64:
+                _logger.warning(
+                    f"Function name {tool_name} is too long, truncating to 64 characters {tool_name[-64:]}."
```
smurching:
```suggestion
                    f"Tool name {tool_name} is too long, truncating to 64 characters {tool_name[-64:]}."
```

smurching (approved):
LGTM!

PR #51, opened 2025-01-30
[ML-49586] Adding resources to the vectorSearchTool

sunishsheth2009:
Adding resources to the vectorSearchTool

### DIFF
--- a/integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
+++ b/integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
@@ -10,6 +10,7 @@ from langchain_core.embeddings import Embeddings
 from langchain_core.tools import BaseTool
 from pydantic import BaseModel, Field, PrivateAttr, model_validator
 
+from databricks_langchain import DatabricksEmbeddings
 from databricks_langchain.vectorstores import DatabricksVectorSearch
 
 
@@ -51,6 +52,10 @@ class VectorSearchRetrieverTool(BaseTool, VectorSearchRetrieverToolMixin):
         self.name = self.tool_name or self.index_name
         self.description = self.tool_description or self._get_default_tool_description(
             IndexDetails(dbvs.index)
+        )
+        self.resources = self._get_resources(
+            self.index_name,
+            (self.embedding.endpoint if isinstance(self.embedding, DatabricksEmbeddings) else None),
         )
 
         return self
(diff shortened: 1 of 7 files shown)


integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
```diff
@@ -52,6 +53,14 @@ def _validate_tool_inputs(self):
         self.description = self.tool_description or self._get_default_tool_description(
             IndexDetails(dbvs.index)
         )
+        self.resources = self._get_resources(
+            self.index_name,
+            (
+                self.embedding.endpoint
+                if isinstance(self.embedding, DatabricksEmbeddings)
```
sunishsheth2009: For endpoint here I am checking if we are using DatabricksEmbedding since that is the only embedding using databricks resources we support with automatic auth for resources

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -125,6 +125,15 @@ def get_tool_name():
             description=self.tool_description
             or self._get_default_tool_description(self._index_details),
         )
+
+        try:
+            from databricks.sdk import WorkspaceClient
+
+            WorkspaceClient().serving_endpoints.get(self.embedding_model_name)
+            self.resources = self._get_resources(self.index_name, self.embedding_model_name)
```
sunishsheth2009, line 133: Here for resources for embedding, we are checking if the embedding model is a model serving endpoint that we can grant access for. So doing a quick check here. If the resource doesn't exist we fallback to not returning back resources for endpoint here.

src/databricks_ai_bridge/vector_search_retriever_tool.py
```diff
@@ -54,6 +61,9 @@ class VectorSearchRetrieverToolMixin(BaseModel):
     )
     tool_name: Optional[str] = Field(None, description="The name of the retrieval tool.")
     tool_description: Optional[str] = Field(None, description="A description of the tool.")
+    resources: Optional[List[dict]] = Field(
+        None, description="Resources required to log the model."
```
annzhang-db: Nit: This seems a little confusing to me - could we change the description to something like "Resources required to log a model that uses this tool" or something since there isn't really a model here?

annzhang-db (approved):
lgtm! 😄

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
smurching: Curious what exactly that means - is it like the user could be using a different embedding, in which case we should probably log a warning that we're unable to autodetect the resource + they should use manual auth (set `DATABRICKS_TOKEN` etc)?

sunishsheth2009: I don't think we should throw a warning or something. Since if they are using different embedding from Langchain or openai or something that are not databricks-resources, we don't need to generate resources for it. So we just generated resources for the index instead. That is why the code is in try...catch.
What do you think?

PR #56, opened 2025-01-31
Fixing the description to not look for source_table comment for default description of vector_search_tool

sunishsheth2009:
Fixing the description to not look for source_table comment for default description of vector_search_tool

### DIFF
--- a/integrations/langchain/pyproject.toml
+++ b/integrations/langchain/pyproject.toml
@@ -13,8 +13,9 @@ dependencies = [
     "langchain-community>=0.2.0",
     "databricks-vectorsearch>=0.40",
     "databricks-ai-bridge>=0.1.0",
-    "unitycatalog-langchain",
+    "mlflow>=2.20.1",
     "pydantic>2.10.0",
+    "unitycatalog-langchain",
 ]
 
 [project.optional-dependencies]
--- a/integrations/langchain/pyproject.toml
+++ b/integrations/langchain/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "databricks-langchain"
-version = "0.1.1.dev0"
+version = "0.3.0.dev0"
 description = "Support for Databricks AI support in LangChain"
 authors = [
     { name="Prithvi Kannan", email="prithvi.kannan@databricks.com" },
@@ -11,7 +11,6 @@ requires-python = ">=3.9"
 dependencies = [
     "langchain>=0.2.0",
     "langchain-community>=0.2.0",
-    "mlflow",
     "databricks-vectorsearch>=0.40",
     "databricks-ai-bridge>=0.1.0",
     "unitycatalog-langchain",
--- a/integrations/openai/pyproject.toml
+++ b/integrations/openai/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "databricks-openai"
-version = "0.1.0"
+version = "0.2.0.dev"
 description = "Support for Databricks AI support with OpenAI"
 authors = [
     { name="Leon Bi", email="leon.bi@databricks.com" },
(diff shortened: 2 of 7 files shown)


integrations/openai/tests/unit_tests/test_vector_search_retriever_tool.py
```diff
@@ -151,7 +144,9 @@ def test_vector_search_retriever_tool_init(
         if self_managed_embeddings_test.embedding_model_name
         else []
     )
-    assert vector_search_tool.resources == expected_resources
+    assert [res.to_dict() for res in vector_search_tool.resources] == [
```
sunishsheth2009, line 147: Fixing a flaky test

integrations/openai/tests/unit_tests/test_vector_search_retriever_tool.py
```diff
@@ -169,6 +164,8 @@ def test_vector_search_retriever_tool_init(
     trace = mlflow.get_last_active_trace()
     spans = trace.search_spans(name=tool_name or index_name, span_type=SpanType.RETRIEVER)
     assert len(spans) == 1
+    print("SUNISH_SPANS", spans)
```
smurching: Let's drop these :D

src/databricks_ai_bridge/test_utils/vector_search.py
```diff
@@ -135,13 +135,13 @@ def _get_index(
 
 @pytest.fixture(autouse=True)
 def mock_workspace_client() -> Generator:
-    def _get_table_comment(full_name: str) -> MagicMock:
-        table = MagicMock()
-        table.comment = "Mocked table comment"
-        return table
+    def _get_serving_endpoint(full_name: str) -> MagicMock:
```
smurching, line 138: QQ why is this mock needed?

sunishsheth2009: This mock is still needed when fetching the resources for vectorSearch. We check if the embedding endpoint is databricks

integrations/langchain/pyproject.toml
```diff
@@ -11,7 +11,6 @@ requires-python = ">=3.9"
 dependencies = [
     "langchain>=0.2.0",
     "langchain-community>=0.2.0",
-    "mlflow",
```
sunishsheth2009, line 14: We don't need mlflow. We only rely on mlflow-skinny

smurching: If we import any MLflow APIs in our code, it may still make sense to include it, because otherwise we're relying on MLflow being a transitive dependency of other packages, which may change later on. There are some weird dependency conflict issues unfortunately with depending on `mlflow-skinny` in addition to `mlflow`, so if this is true:
>If we import any MLflow APIs in our code


then I'd suggest just keeping `mlflow` here, maybe with a min version that includes the APIs we need

smurching (commented):
One small question/comment, otherwise LGTM!

integrations/langchain/pyproject.toml
```diff
@@ -11,11 +11,11 @@ requires-python = ">=3.9"
...
     "databricks-vectorsearch>=0.40",
     "databricks-ai-bridge>=0.1.0",
-    "unitycatalog-langchain",
+    "mlflow>=2.20.0",
```
smurching:
```suggestion
    "mlflow>=2.20.1",
```
Since 2.20.1 has some bugs in agent authoring APIs

integrations/openai/pyproject.toml
```diff
@@ -12,6 +12,8 @@ dependencies = [
     "databricks-vectorsearch>=0.40",
     "databricks-ai-bridge>=0.1.0",
     "openai>=1.46.1",
+    "pydantic>2.10.0",
+    "mlflow>=2.20.0",
```
smurching:
```suggestion
    "mlflow>=2.20.1",
```

smurching (approved):
LGTM once my suggestions are applied!

PR #64, opened 2025-02-14
Add [databricks] option to unitycatalog in requirements and require python 3.10

bbqiu:
`unitycatalog-*` packages are now using serverless warehouses to execute tool calls, which depends on the package `databricks-connect`, which is only compatible with python>=3.10. to add `unitycatalog-langchain[databricks]` as a core dependency to `databricks-langchain`, we need to bump the min python version for the integration packages.

`databricks-ai-bridge` itself is fine to keep at 3.9.

### DIFF
--- a/integrations/langchain/pyproject.toml
+++ b/integrations/langchain/pyproject.toml
@@ -7,7 +7,7 @@ authors = [
 ]
 readme = "README.md"
 license = { text="Apache-2.0" }
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 dependencies = [
     "langchain>=0.2.0",
     "langchain-community>=0.2.0",
@@ -15,7 +15,7 @@ dependencies = [
     "databricks-ai-bridge>=0.1.0",
     "mlflow>=2.20.1",
     "pydantic>2.10.0",
-    "unitycatalog-langchain",
+    "unitycatalog-langchain[databricks]>=0.1.1",
 ]
 
 [project.optional-dependencies]
--- a/integrations/openai/pyproject.toml
+++ b/integrations/openai/pyproject.toml
@@ -14,7 +14,7 @@ dependencies = [
     "openai>=1.46.1",
     "pydantic>2.10.0",
     "mlflow>=2.20.1",
-    "unitycatalog-openai>=0.1.0",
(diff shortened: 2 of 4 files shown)


integrations/langchain/pyproject.toml
```diff
@@ -15,7 +15,7 @@ dependencies = [
     "databricks-ai-bridge>=0.1.0",
     "mlflow>=2.20.1",
     "pydantic>2.10.0",
-    "unitycatalog-langchain",
+    "unitycatalog-langchain[databricks]>=0.1.1",
```
sunishsheth2009, line 18: Small nit. Do we need the same here: https://github.com/databricks/databricks-ai-bridge/blob/main/integrations/openai/pyproject.toml#L17

sunishsheth2009 (approved):

integrations/langchain/pyproject.toml
bbqiu: added it there too, thanks!

sunishsheth2009 (approved):

integrations/openai/pyproject.toml
```diff
@@ -7,14 +7,14 @@ authors = [
...
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 dependencies = [
     "databricks-vectorsearch>=0.40",
     "databricks-ai-bridge>=0.1.0",
     "openai>=1.46.1",
     "pydantic>2.10.0",
     "mlflow>=2.20.1",
-    "unitycatalog-openai",
+    "unitycatalog-openai>=0.1.0",
```
serena-ruan:
```suggestion
    "unitycatalog-openai[databricks]>=0.1.0",
```

pyproject.toml
```diff
@@ -6,7 +6,7 @@ authors = [
     { name="Databricks", email="agent-feedback@databricks.com" },
 ]
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
```
serena-ruan: Seems like 3.9 is fine with this package, if users use it standalone

bbqiu: sounds good, will leave this at 3.9

PR #66, opened 2025-02-19
[ML-50520] Add disable_notice=True to VectorSearchClient integrations

bbqiu:
https://databricks.atlassian.net/browse/ML-50520

tested, no more warning:
- openai: https://e2-dogfood.staging.cloud.databricks.com/editor/notebooks/1148929843128572?o=6051921418418893
- langchain: https://e2-dogfood.staging.cloud.databricks.com/editor/notebooks/1148929843128550?o=6051921418418893

### DIFF
--- a/integrations/langchain/src/databricks_langchain/vectorstores.py
+++ b/integrations/langchain/src/databricks_langchain/vectorstores.py
@@ -242,7 +242,9 @@ class DatabricksVectorSearch(VectorStore):
             ) from e
 
         try:
-            self.index = VectorSearchClient(**(client_args or {})).get_index(
+            client_args = client_args or {}
+            client_args["disable_notice"] = True
+            self.index = VectorSearchClient(**client_args).get_index(
                 endpoint_name=endpoint, index_name=index_name
             )
         except Exception as e:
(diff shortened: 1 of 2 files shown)


integrations/langchain/src/databricks_langchain/vectorstores.py
```diff
@@ -242,7 +242,9 @@ def __init__(
             ) from e
 
         try:
-            self.index = VectorSearchClient(**(client_args or {})).get_index(
+            client_args = client_args or {}
+            client_args["disable_notice"] = True
```
sunishsheth2009, line 246: Can we do something like this:

```
client_args = {"disable_notice": True, **(client_args or {})}
```

So here we are setting it as default but we let users override it if needed rather than overriding what users pass in?

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -97,7 +97,7 @@ def _validate_tool_inputs(self):
             raise ValueError(
                 f"Index name {self.index_name} is not in the expected format 'catalog.schema.index'."
             )
-        self._index = VectorSearchClient().get_index(index_name=self.index_name)
+        self._index = VectorSearchClient(disable_notice=True).get_index(index_name=self.index_name)
```
sunishsheth2009, line 100: Interesting that we are not allowing users pass `client_args`. But lets not block this PR for that. We can create a ticket to track this later on.

sunishsheth2009 (approved):
Thanks for fixing this quickly :)

PR #67, opened 2025-02-20
Genie agent metadata

manfredcalvo:
In this pull request, I am making the following changes to the GenieAgent:

1) Creating a Python class that extends the RunnableLambda interface to create the GenieAgent instead of having a function to represent the GenieAgent. 
2) Adding a new boolean parameter to the GenieAgent class called "return_metadata." If the flag is True, then the output of the invoke function of the GenieAgent will have two keys: "messages" and "metadata." This metadata contains the direct response from GenieAPI, which is the object GenieResponse.
3) Some minor refactoring of the GenieAgent unit tests. 

If the following screenshots, you can see the output dict and the output trace in both scenarios (return_metadata True and False)

When return_medata is False (this is the default value)

Screenshot 2025-02-19 at 6 48 13 PM

When return_metadata is True

Screenshot 2025-02-19 at 6 49 45 PM

### DIFF
--- a/integrations/langchain/src/databricks_langchain/genie.py
+++ b/integrations/langchain/src/databricks_langchain/genie.py
@@ -1,53 +1,63 @@
 import mlflow
+from langchain_core.messages import AIMessage, BaseMessage
+
 from databricks_ai_bridge.genie import Genie
 
+from langchain_core.runnables import RunnableLambda
 
-@mlflow.trace()
-def _concat_messages_array(messages):
-    concatenated_message = "\n".join(
-        [
-            f"{message.get('role', message.get('name', 'unknown'))}: {message.get('content', '')}"
-            if isinstance(message, dict)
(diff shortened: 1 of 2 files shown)


integrations/langchain/src/databricks_langchain/genie.py
```diff
@@ -1,53 +1,62 @@
...
+from langchain_core.messages import AIMessage, BaseMessage
+
 from databricks_ai_bridge.genie import Genie
 
+from langchain_core.runnables import RunnableLambda
+
+from typing import Dict, Any
+
+
+class GenieAgent(RunnableLambda):
```
FMurray, line 11: Suggestion: add a docstring

manfredcalvo: @FMurray Doc string added to the class.

manfredcalvo:
@bbqiu let me know if this can be merged or I need to follow a different proccess.

PR #72, opened 2025-02-26
Update GenieAgent

bbqiu:
changes in this PR:
- raise a value error when no genie space ID is provided
- return an error message instead of None when Genie does not return anything
- add a timeout message

tested by manual and unit tests

### DIFF
--- a/src/databricks_ai_bridge/genie.py
+++ b/src/databricks_ai_bridge/genie.py
@@ -120,8 +120,12 @@ class Genie:
                     logging.debug("Waiting for query result...")
                     time.sleep(5)
                 else:
-                    logging.debug(f"No query result: {resp['state']}")
-                    return GenieResponse(None, query, description)
+                    return GenieResponse(f"No query result: {resp['state']}", query, description)
+            return GenieResponse(
+                f"Genie query timed out after {MAX_ITERATIONS} iterations of 5 seconds",
+                query,
+                description,
+            )
 
         @mlflow.trace()
         def poll_result():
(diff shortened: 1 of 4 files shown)


src/databricks_ai_bridge/genie.py
```diff
@@ -133,7 +137,7 @@ def poll_result():
                     f"/api/2.0/genie/spaces/{self.space_id}/conversations/{conversation_id}/messages/{message_id}",
                     headers=self.headers,
                 )
-                if resp["status"] == "EXECUTING_QUERY" or resp["status"] == "COMPLETED":
+                if resp["status"] in ["EXECUTING_QUERY", "COMPLETED"]:
```
aravind-segu: nit: use a set instead of list
```suggestion
                if resp["status"] in {"EXECUTING_QUERY", "COMPLETED"}:
```

src/databricks_ai_bridge/genie.py
```diff
@@ -144,12 +148,20 @@ def poll_result():
...
+                elif resp["status"] in ["CANCELLED", "QUERY_RESULT_EXPIRED"]:
+                    return GenieResponse(result=f"Genie query {resp['status'].lower()}.")
+                elif resp["status"] == "FAILED":
+                    return GenieResponse(
+                        result=f"Genie query failed with error: {resp.get('error', 'Unknown error')}"
+                    )
                 else:
                     logging.debug(f"Waiting...: {resp['status']}")
                     time.sleep(5)
+            return GenieResponse(
```
aravind-segu, line 160: Nit: change this message to `Genie result timed out ...` so we can differentiate it from the above time out message for debugging

aravind-segu (approved):
Small nits, but looks good

PR #74, opened 2025-03-04
Update inline for unitycatalog-ai imports in databricks-langchain and databricks-openai

annzhang-db:
- Update inline for unitycatalog-ai imports in databricks-langchain and databricks-openai


integrations/langchain/src/databricks_langchain/__init__.py
```diff
@@ -1,3 +1,7 @@
+from unitycatalog.ai.core.base import set_uc_function_client
+from unitycatalog.ai.core.databricks import DatabricksFunctionClient
+from unitycatalog.ai.langchain.toolkit import UCFunctionToolkit
```
sunishsheth2009, line 3: Removing the existing import would break existing customers. Can we add both if we want to inline it as well? 

cc @bbqiu who might be using it in the notebooks as well.

bbqiu: +1 can we have both and add a deprecation warning when importing from uc_ai? prob won't actually remove it for a few months, but the example notebooks have already been shipped out importing from uc_ai

bbqiu (approved):
would you mind testing the deprecated import behavior in a notebook?

LGTM after that!

PR #75, opened 2025-03-06
Support Credential Strategy in Vector Search Client

aravind-segu:
Add an extra parameter to pass in credential strategy to vector search client. This will ensure that the tool is called with invokers rights. 

Added unit tests, pending testing on Dogfood

### DIFF
--- a/integrations/langchain/src/databricks_langchain/vectorstores.py
+++ b/integrations/langchain/src/databricks_langchain/vectorstores.py
@@ -17,6 +17,7 @@ from typing import (
 )
 
 import numpy as np
+from databricks.sdk import WorkspaceClient
 from databricks_ai_bridge.utils.vector_search import (
     IndexDetails,
     parse_vector_search_response,
@@ -223,6 +224,7 @@ class DatabricksVectorSearch(VectorStore):
         embedding: Optional[Embeddings] = None,
         text_column: Optional[str] = None,
         columns: Optional[List[str]] = None,
+        workspace_client: Optional[WorkspaceClient] = None,
         client_args: Optional[Dict[str, Any]] = None,
     ):
         if not (isinstance(index_name, str) and _INDEX_NAME_PATTERN.match(index_name)):
@@ -235,6 +237,7 @@ class DatabricksVectorSearch(VectorStore):
             from databricks.vector_search.client import (  # type: ignore[import]
                 VectorSearchClient,
             )
+            from databricks.vector_search.utils import CredentialStrategy
         except ImportError as e:
             raise ImportError(
                 "Could not import databricks-vectorsearch python package. "
@@ -244,6 +247,13 @@ class DatabricksVectorSearch(VectorStore):
         try:
             client_args = client_args or {}
             client_args.setdefault("disable_notice", True)
+            if (
+                workspace_client is not None
+                and workspace_client.config.auth_type == "model_serving_user_credentials"
+            ):
+                client_args.setdefault(
+                    "credential_strategy", CredentialStrategy.MODEL_SERVING_USER_CREDENTIALS
+                )
             self.index = VectorSearchClient(**client_args).get_index(
                 endpoint_name=endpoint, index_name=index_name
             )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,7 @@ dependencies = [
   "typing_extensions",
   "pydantic",
   "databricks-sdk>=0.34.0",
+  "databricks-vectorsearch>=0.50",
   "pandas",
   "tiktoken>=0.8.0",
   "tabulate",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,8 +10,7 @@ requires-python = ">=3.9"
 dependencies = [
   "typing_extensions",
   "pydantic",
-  "databricks-sdk>=0.34.0",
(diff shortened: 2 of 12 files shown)


pyproject.toml
```diff
@@ -11,6 +11,7 @@ dependencies = [
   "typing_extensions",
   "pydantic",
   "databricks-sdk>=0.34.0",
+  "databricks-vectorsearch>=0.50",
```
sunishsheth2009: If we are adding this here, can we remove it from other project deps as well so we need to update only in one place rather than multiple projects?

src/databricks_ai_bridge/vector_search_retriever_tool.py
```diff
@@ -2,6 +2,7 @@
 from typing import Any, Dict, List, Optional
 
 import mlflow
+from databricks.vector_search.utils import CredentialStrategy
```
sunishsheth2009: Also I am curious if we should include this globally just for the types. Maybe we can localize the import and no need to include the dep in project.toml for databricks-ai-bridge

src/databricks_ai_bridge/vector_search_retriever_tool.py
```diff
@@ -62,6 +63,7 @@ class VectorSearchRetrieverToolMixin(BaseModel):
     resources: Optional[List[dict]] = Field(
         None, description="Resources required to log a model that uses this tool."
     )
+    credential_strategy: Optional[CredentialStrategy] = Field(default=None, description="When set to MODEL_SERVING_USER_CREDENTIALS, the tool will be authorized with on behalf of user rights")
```
sunishsheth2009: Overall i am curious if we should ask users to pass in the workspace_client instead? So that we can support things like profile, credential_strategy, etc. What do you think? This will be similar to other APIs as well which just cares about passing in the workspaceClient. So if the client is already created we don't need to pass in just the credential_strategy here. Thoughts?

aravind-segu: @smurching I think this is a good idea to have a WorkspaceClient being passed into both VectorSearchTools and UCFunctionToolkit. Internally we could take the workspace client and pass in the credential strategy when we are actually creating the VectorSearchClient since VectorSearchClient does not take in WorkspaceClient.

On the plus side, it would be a consistent CUJ for our users.

on the negative, if users are using the VectorSearchClient directly, then they would have to pass in this credential strategy explicitly. So using VectorSearchClient directly vs using a VectorSearchTool would have different parameters. Is that ok? What do you think

smurching: What's the overall plan for the vector search SDK (it will eventually move to Databricks SDK right?). I think it's fine for us to be consistent with the vector search SDK for now (pass individual params like `credential_strategy` in both places) since to Aravind's point, it may also be confusing to pass a `WorkspaceClient` and find that only the subset of WorkspaceClient attributes that are supported by the VS client are honored in the tool. But definitely long term I'd hope we can pass a WorkspaceClient here

smurching: But if there aren't a lot of config parameters, and the VS SDK is going to merge into Databricks SDK soon, then maybe we should just pass WorkspaceClient upfront. Just don't want us to have to maintain the translation from WorkspaceClient -> VectorSearchClient parameters + have different behavior from the VS client for a long time period

sunishsheth2009: Yeah! that is what i was thinking that in the long run anyway, VS SDK would be deprecated and migrated to workspaceClient. if that happens, there won't be any change in our API.
Plus for users we recommend using UCFunctionToolkit and databricks-ai-bridge for vector search, etc. For all those usecases we can expect workspaceClient then its super consistent for the users

aravind-segu: @smurching @sunishsheth2009 I am not sure if there is a target date for VS Client to migrate to Workspace Client, but I know its going to happen at some point. But if we want to steer our users to use the VS Tool from ai-bridge and not really promote the VectorSearchClient directly, then I think taking in WorkspaceClient here makes sense

integrations/langchain/src/databricks_langchain/vectorstores.py
```diff
@@ -244,6 +247,13 @@ def __init__(
         try:
             client_args = client_args or {}
             client_args.setdefault("disable_notice", True)
+            if (
+                workspace_client is not None
+                and workspace_client.config.auth_type == "model_serving_user_credentials"
+            ):
+                client_args.setdefault(
+                    "credential_strategy", CredentialStrategy.MODEL_SERVING_USER_CREDENTIALS
+                )
```
sunishsheth2009, line 256: What if `workspace_client.config.auth_type` is set to something else? Do we want to check and throw some error/warning for the configs we don't support?

sunishsheth2009 (approved):

integrations/langchain/src/databricks_langchain/vectorstores.py
aravind-segu: We dont want to throw an error. If its not model serving, then we set credential_strategy to none and it will basically pick up the auth from VSClient. Which can be notebook auth, definers auth etc.

PR #76, opened 2025-03-06
[ML-49514] Fix metaclass issue

bbqiu:
Users are running into this error when installing `databricks-langchain`
`metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases`

Reproduced in https://e2-dogfood.staging.cloud.databricks.com/editor/notebooks/2183817319091232?o=6051921418418893#command/8852559416333002.

Left side has the error, right side does not. Upgrading version of LangChain from 0.2.17 --> 0.3.0 fixes this issue, and it's the next release (https://pypi.org/project/langchain/0.3.0/#history).

https://www.diffchecker.com/peml3OxV/

### DIFF
--- a/integrations/langchain/pyproject.toml
+++ b/integrations/langchain/pyproject.toml
@@ -9,8 +9,8 @@ readme = "README.md"
 license = { text="Apache-2.0" }
 requires-python = ">=3.10"
 dependencies = [
-    "langchain>=0.2.0",
-    "langchain-community>=0.2.0",
+    "langchain>=0.3.0",
+    "langchain-community>=0.3.0",
     "databricks-vectorsearch>=0.40",
     "databricks-ai-bridge>=0.1.0",
     "mlflow>=2.20.1",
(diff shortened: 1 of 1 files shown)


integrations/langchain/pyproject.toml
```diff
@@ -9,8 +9,8 @@ readme = "README.md"
 license = { text="Apache-2.0" }
 requires-python = ">=3.10"
 dependencies = [
-    "langchain>=0.2.0",
-    "langchain-community>=0.2.0",
+    "langchain>=0.3.0",
+    "langchain-community>=0.3.0",
```
smurching: Great find - curious in general, do we still have a dependency on LC community now that we've moved all our integration code into this repo? I'd expect we wouldn't need it anymore

smurching (approved):

integrations/langchain/pyproject.toml
sunishsheth2009: +1 I don't see any more imports for `langchain-community` in our code. We should try removing it and see if the tests pass.

bbqiu: removed + all tests passed, thanks!

PR #78, opened 2025-03-07
Add validations on VectorSearchRetrieverTool name

annzhang-db:
**Description of changes**
- Validate tool names and make sure that "." is replaced with "__" for all integrations.

**How is this tested?**
- [x] Unit tests
- [x] Manual test
Before:
Screenshot 2025-03-07 at 11 31 21 AM

After:
Screenshot 2025-03-07 at 11 31 11 AM

### DIFF
--- a/src/databricks_ai_bridge/vector_search_retriever_tool.py
+++ b/src/databricks_ai_bridge/vector_search_retriever_tool.py
@@ -9,6 +9,7 @@ from mlflow.models.resources import (
     Resource,
 )
 from pydantic import BaseModel, Field
+import re
 
 from databricks_ai_bridge.utils.vector_search import IndexDetails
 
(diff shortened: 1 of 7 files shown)


src/databricks_ai_bridge/vector_search_retriever_tool.py
```diff
@@ -78,3 +89,15 @@ def _get_resources(self, index_name: str, embedding_endpoint: str) -> List[Resou
             if embedding_endpoint
             else []
         )
+
+    def _get_tool_name(self) -> str:
+        tool_name = self.tool_name or self.index_name.replace(".", "__")
+
+        # Tool names must match the pattern '^[a-zA-Z0-9_-]+$'."
```
aravind-segu, line 96: should we also check the tool name against this regex. If the tool name they pass in for example has '.' characters, we should throw early instead of letting this propogate and fail when an agent is called.

src/databricks_ai_bridge/vector_search_retriever_tool.py
```diff
@@ -78,3 +89,15 @@ def _get_resources(self, index_name: str, embedding_endpoint: str) -> List[Resou
...
+        # The '.' from the index name are not allowed
+        if len(tool_name) > 64:
+            _logger.warning(
+                f"Tool name {tool_name} is too long, truncating to 64 characters {tool_name[-64:]}."
+            )
+            return tool_name[-64:]
```
aravind-segu, line 108: Any reason for the last 64 chars. The first 64 chars might make more sense?

aravind-segu (approved):
Looks good, just one minor comment

src/databricks_ai_bridge/vector_search_retriever_tool.py
annzhang-db: This was because we are assuming it's going to be `catalog.schema.index` and it makes more sense to keep the index part of that as the tool name and cut off `catalog.schema` instead.

PR #82, opened 2025-03-11
[ML-51228] Genie Client Remove types

bbqiu:
After discussing w/ the Genie team, they are only returning string responses now, so we do not need to cast anymore


bbqiu:
closing because this casting can't hurt but can make some responses more human readable (ex. binary type response)

PR #84, opened 2025-03-13
Add Model Name and System Fingerprint to `llm_output` in `_convert_response_to_chat_result`

jmoreno11:
This PR enhances `_convert_response_to_chat_result` by including the model name and system fingerprint in `llm_output`. The model name is particularly important for observability tools like LangFuse, which rely on `llm_output["model_name"]` for tracking.  

Currently, Databricks **does not provide** the model name in the metadata—only the endpoint name. This change ensures that external models are properly identified, improving observability and logging.  

**Changes:**  
- Added `model_name` to `llm_output`, defaulting to `self.model` if not in the response.  
- Added `system_fingerprint` to `llm_output` for additional metadata.  
- Updated `llm_output` to include `token_usage` separately for better structure.  

**Why?**  
When using external models with Databricks, the returned metadata only includes the endpoint name, not the actual model name. This change allows users to extract the correct model name for logging and monitoring.

**Testing:**  
- Ensured the new metadata fields are correctly populated when present in the response.  
- Maintains backward compatibility by defaulting to `self.model` if `response["model"]` is missing.

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -308,10 +308,11 @@ class ChatDatabricks(BaseChatModel):
             for choice in response["choices"]
         ]
         llm_output = {
-            "token_usage": response.get("usage", {}),
-            "model_name": response.get("model", self.model),
-            "system_fingerprint": response.get("system_fingerprint", ""),
+            k: v for k, v in response.items() if k not in ("choices", "content", "role", "type")
         }
+        if "model" in llm_output and "model_name" not in llm_output:
+            llm_output["model_name"] = llm_output["model"]
+        
         return ChatResult(generations=generations, llm_output=llm_output)
 
     def _stream(
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -18,7 +18,6 @@ from typing import (
     Union,
 )
 
-from databricks_langchain.utils import get_deployment_client
 from langchain_core.callbacks import CallbackManagerForLLMRun
 from langchain_core.language_models import BaseChatModel
 from langchain_core.language_models.base import LanguageModelInput
@@ -54,6 +53,8 @@ from langchain_core.utils.function_calling import convert_to_openai_tool
 from langchain_core.utils.pydantic import is_basemodel_subclass
 from mlflow.deployments import BaseDeploymentClient  # type: ignore
 from pydantic import BaseModel, ConfigDict, Field
+
+from databricks_langchain.utils import get_deployment_client
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -307,8 +307,12 @@ def _convert_response_to_chat_result(self, response: Mapping[str, Any]) -> ChatR
             )
             for choice in response["choices"]
         ]
-        usage = response.get("usage", {})
-        return ChatResult(generations=generations, llm_output=usage)
+        llm_output = {
+            "token_usage": response.get("usage", {}),
```
sunishsheth2009: Can we rename this to usage so it doesn't change the API for existing users?

```suggestion
            "usage": response.get("usage", {}),
```

sunishsheth2009 (commented):
Can we also update tests here: https://github.com/databricks/databricks-ai-bridge/blob/main/integrations/langchain/tests/unit_tests/test_chat_models.py#L84

Thank you for contributing. :)

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -307,8 +307,12 @@ def _convert_response_to_chat_result(self, response: Mapping[str, Any]) -> ChatR
...
+            "model_name": response.get("model", self.model),
+            "system_fingerprint": response.get("system_fingerprint", ""),
+        }
```
bbqiu, line 312: ooc, how did we choose these two fields? I'm seeing from other community integrations that they end up sticking anything that is not part of the traditional message fields into LLM_output. maybe we could just do this?

anthropic ex: https://github.com/langchain-ai/langchain/blob/5ede4248eff56de917ce274a6cc408b0a32a9827/libs/partners/anthropic/langchain_anthropic/chat_models.py#L1064-L1067C9

integrations/langchain/src/databricks_langchain/chat_models.py
jmoreno11: I did as @bbqiu suggested, so `usage` remains intact

integrations/langchain/src/databricks_langchain/chat_models.py
jmoreno11: I took your suggestion and adapt it accordingly, see latest changes.

integrations/langchain/src/databricks_langchain/chat_models.py
jmoreno11: I also created a new unit test, since the existing ones were not altered by this change

integrations/langchain/src/databricks_langchain/chat_models.py
bbqiu: looks great! thanks

bbqiu (approved):
lgtm!

jmoreno11:
Please reapprove, I fixed the linting error (`ruff --fix`)

sunishsheth2009 (approved):

jmoreno11:
Apologies, ruff cli was doing something weird. I created a PR on my forked branch and the linting is now fixed.
image

bbqiu (approved):
approving to unblock merge

PR #88, opened 2025-03-21
Remove default temperature for ChatDatabricks

bbqiu:
- no deprecation warning when temperature is passed in (either None or 0)
  - allow None to be compatible w/ EMs like o1
- deprecation warning when temperature is not passed
  - need to keep default at 0 since FMAPI default is not 0  ([slack msg](https://databricks.slack.com/archives/C03RVD95PU6/p1742588626045899)) ([additional discussion](https://databricks.slack.com/archives/C06KX8NAV8D/p1742598890036829?thread_ts=1742520999.407119&cid=C06KX8NAV8D))

tested here w/ o1 + llama: https://e2-dogfood.staging.cloud.databricks.com/editor/notebooks/1016946506451086?o=6051921418418893#command/7321788187864732

screenshot 2025-03-24-13-59-52-Arc

screenshot 2025-03-24-13-56-54-Arc

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -215,6 +215,7 @@ class ChatDatabricks(BaseChatModel):
     target_uri: str = "databricks"
     """The target URI to use. Defaults to ``databricks``."""
     temperature: Optional[float] = 0.0
+    # TODO: bbqiu log a deprecated warning when temperature is not set
     """Sampling temperature. Higher values make the model more creative."""
     n: int = 1
     """The number of completion choices to generate."""
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -214,7 +214,7 @@ class ChatDatabricks(BaseChatModel):
     """Name of Databricks Model Serving endpoint to query."""
     target_uri: str = "databricks"
     """The target URI to use. Defaults to ``databricks``."""
-    temperature: float
+    temperature: Optional[float] = None
     """Sampling temperature. Higher values make the model more creative."""
     n: int = 1
     """The number of completion choices to generate."""
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -214,7 +214,7 @@ class ChatDatabricks(BaseChatModel):
     """Name of Databricks Model Serving endpoint to query."""
     target_uri: str = "databricks"
     """The target URI to use. Defaults to ``databricks``."""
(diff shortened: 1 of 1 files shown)


integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -214,7 +214,7 @@ class GetPopulation(BaseModel):
     """Name of Databricks Model Serving endpoint to query."""
     target_uri: str = "databricks"
     """The target URI to use. Defaults to ``databricks``."""
-    temperature: float = 0.0
+    temperature: Optional[float] = None
```
sunishsheth2009: This might be a breaking change since now we are not passing a default value as 0.0.

Should we do?
`temperature: Optional[float] = 0.0`

and o1 users can do temperature=None when initializing ChatDatabricks.
I agree that is weird.

But thinking out loud I guess we are ok to make this breaking change. 🤔

sunishsheth2009 (approved):

integrations/langchain/src/databricks_langchain/chat_models.py
bbqiu: i dont think it's technically breaking since the LLM calls will still work, but I agree that the temp being higher now could be confusing, but I don't think setting the temp to None makes a lot of sense for o1 users.

added a brief test in the description to verify that things work + maybe we should discuss this tmrw

integrations/langchain/pyproject.toml
```diff
@@ -15,6 +15,7 @@ dependencies = [
     "mlflow>=2.20.1",
     "pydantic>2.10.0",
     "unitycatalog-langchain[databricks]>=0.2.0",
+    "databricks-connect>=16.1.1"
```
harupy: Can we move this to `dev`?

bbqiu: wouldn't this pip install error happen for our customers as well? why wouldn't we want this to apply to all installs?

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -214,7 +214,8 @@ class GetPopulation(BaseModel):
     """Name of Databricks Model Serving endpoint to query."""
     target_uri: str = "databricks"
     """The target URI to use. Defaults to ``databricks``."""
-    temperature: float = 0.0
+    temperature: Optional[float] = 0.0
+    # TODO: bbqiu log a deprecated warning when temperature is not set
```
smurching:
```suggestion
```

smurching (approved):

bbqiu:
some changes inspired by: https://github.com/langchain-ai/langchain/blob/07823cd41c95ef1594f9075f70e9f7a633433842/libs/partners/openai/langchain_openai/chat_models/base.py#L639

PR #95, opened 2025-04-05
Remove regex validation for `index_name`

harupy:
Clone of https://github.com/langchain-ai/langchain-databricks/pull/44


integrations/langchain/src/databricks_langchain/vectorstores.py
```diff
@@ -227,10 +225,9 @@ def __init__(
         workspace_client: Optional[WorkspaceClient] = None,
         client_args: Optional[Dict[str, Any]] = None,
     ):
-        if not (isinstance(index_name, str) and _INDEX_NAME_PATTERN.match(index_name)):
```
sunishsheth2009, line 230: In the long run, can we check something like this to replace this regex:
`if len(model_name.split(".")) != 3:`
but I get this is solving the problem of allowing special chars. So we can merge this one

sunishsheth2009 (approved):

PR #98, opened 2025-04-09
databricks_langchain 0.4.2 release

sunishsheth2009:
Improve validation for `index_name` for vectorIndex

### DIFF
--- a/integrations/langchain/pyproject.toml
+++ b/integrations/langchain/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "databricks-langchain"
-version = "0.4.1.dev0"
+version = "0.4.2"
 description = "Support for Databricks AI support in LangChain"
 authors = [
     { name="Databricks", email="agent-feedback@databricks.com" },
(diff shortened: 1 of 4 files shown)


integrations/langchain/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "databricks-langchain"
-version = "0.4.1.dev0"
+version = "0.4.2"
```
annzhang-db, line 3: We don't want to commit removing the dev0 here right?

sunishsheth2009: This is against the databricks-ai-bridge-0.4.1 branch. In master yes we want to keep the `.dev0` but in the branch we want to remove it.

Let me know if I am missing something

annzhang-db (approved):
lgtm 😄

integrations/langchain/pyproject.toml
annzhang-db: ah okay, looks good 👍

PR #100, opened 2025-04-09
Support additional retriever fields

annzhang-db:
## What changes are proposed in this pull request?
- Support passing additional retriever fields like `doc_uri` and `chunk_id`
- Supporting fields as provided in `set_retriever_schema` https://mlflow.org/docs/latest/tracing/tracing-schema#retriever-spans and https://mlflow.org/docs/latest/tracing/tracing-schema#retriever-spans
- `primary_key` is mapped to the key `"chunk_id"` in metadata, but calling it `primary_key` in the `RetrieverSchema` to be consistent with the existing behavior of `set_retriever_schema`

## How is this tested?
- Unit tests
- Manual testing
Screenshot 2025-04-10 at 5 13 09 PM
Screenshot 2025-04-10 at 5 13 03 PM

### DIFF
--- a/src/databricks_ai_bridge/utils/vector_search.py
+++ b/src/databricks_ai_bridge/utils/vector_search.py
@@ -70,17 +70,37 @@ class RetrieverSchema:
 
 
 def get_metadata(columns: List[str], result: List[Any], retriever_schema, ignore_cols):
+    """
+    This function constructs a metadata dictionary by mapping column names to their corresponding values
+    from the result row, with special handling for the provided retriever schema.
+
+    Args:
+        columns (List[str]): The list of column names in the result, including a final score column.
+        result (List[Any]): The row of data values corresponding to the columns.
+        retriever_schema: An object that defines which columns represent `doc_uri`, `primary_key`,
+                          and optionally other columns to include.
+        ignore_cols (List[str]): List of column names to exclude from the metadata. Usually contains
+                                 the text column or embedding vector column.
+
+    Returns:
+        Dict[str, Any]: A dictionary containing extracted metadata.
+    """
     metadata = {}
+    
+    # Skipping the last column, which is always the score
     for col, value in zip(columns[:-1], result[:-1]):
         if col == retriever_schema.doc_uri:
             metadata["doc_uri"] = value
         elif col == retriever_schema.primary_key:
             metadata["chunk_id"] = value
         elif col == "doc_uri" and retriever_schema.doc_uri:
+            # Prioritize retriever_schema.doc_uri, don't override with the actual "doc_uri" column
             continue
         elif col == "chunk_id" and retriever_schema.primary_key:
+            # Prioritize retriever_schema.primary_key, don't override with the actual "chunk_id" column
             continue
-        elif col in ignore_cols:  # ignore_cols has precedence over other_columns
+        elif col in ignore_cols:
+            # ignore_cols has precedence over other_columns
             continue
         elif retriever_schema.other_columns is not None:
             if col in retriever_schema.other_columns:
--- a/src/databricks_ai_bridge/utils/vector_search.py
+++ b/src/databricks_ai_bridge/utils/vector_search.py
@@ -76,8 +76,9 @@ def get_metadata(columns: List[str], result: List[], doc_uri, chunk_id, other_co
             metadata["doc_uri"] = value
         elif col == chunk_id:
             metadata["chunk_id"] = value
-        elif other_columns and col in other_columns:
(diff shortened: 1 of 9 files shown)


src/databricks_ai_bridge/utils/vector_search.py
```diff
@@ -60,10 +61,38 @@ def is_databricks_managed_embeddings(self) -> bool:
...
+class RetrieverSchema:
+    text_column: str = None
+    doc_uri: Optional[str] = None
+    primary_key: Optional[str] = None
+    other_columns: Optional[List[str]] = None
+
+
+def get_metadata(columns: List[str], result: List[Any], retriever_schema, ignore_cols):
+    metadata = {}
+    for col, value in zip(columns[:-1], result[:-1]):
```
aravind-segu, line 91: Just Curious, why are we skipping the last element?

src/databricks_ai_bridge/utils/vector_search.py
```diff
@@ -60,10 +61,38 @@ def is_databricks_managed_embeddings(self) -> bool:
...
+        if col == retriever_schema.doc_uri:
+            metadata["doc_uri"] = value
+        elif col == retriever_schema.primary_key:
+            metadata["chunk_id"] = value
+        elif col == "doc_uri" and retriever_schema.doc_uri:
```
aravind-segu, line 96: What case do we need these two if statements

src/databricks_ai_bridge/utils/vector_search.py
```diff
@@ -60,10 +61,38 @@ def is_databricks_managed_embeddings(self) -> bool:
...
+            continue
+        elif col == "chunk_id" and retriever_schema.primary_key:
+            continue
+        elif col in ignore_cols:  # ignore_cols has precedence over other_columns
+            continue
+        elif retriever_schema.other_columns is not None:
+            if col in retriever_schema.other_columns:
+                metadata[col] = value
+        else:
+            metadata[col] = value
```
aravind-segu, line 109: if `retriever_schema` other_columns is specified, then we will never add any remaining columns to metadata. Is that expected behaviour?

tests/databricks_ai_bridge/utils/test_vector_search.py
```diff
@@ -0,0 +1,138 @@
...
+    "result": {
+        "data_array": [
+            ["row 1, column 1", "row 1, column 2", 100, 5.8, 0.673],
+            ["row 2, column 1", "row 2, column 2", 200, 4.1, 0.236],
+        ]
+    },
+}
+
+
+def construct_docs_with_score(
```
aravind-segu, line 19: this test was pretty hard to understand. Can we try to make it simpler or add more comments. 

Maybe:

def construct_docs_with_score(
   page_content,
   columns_to_include: list[str]
 )
 
 and then in the body add the columns to include to metadata with random values because from what i understand the actual values of the column is not tested right?

src/databricks_ai_bridge/utils/vector_search.py
aravind-segu, line 91: Can we add a quick comment here explaining the precedence and hierarchy so its easier to understand in the future.

annzhang-db: The last column is always the score, so we can skip that when getting metadata. I can add a comment explaining this.

src/databricks_ai_bridge/utils/vector_search.py
annzhang-db: Presumably if the user is setting `doc_uri` to some column, they want to get the column they are setting in the retriever schema, so we want to make sure it isn't overriden by some other column that actually has the name `"doc_uri"`.

annzhang-db: Seems not ideal, but not sure what else we could do with the `doc_uri` column if there is an actual column with that name. I decided to prioritize the explicitly set retriever schema here.

src/databricks_ai_bridge/utils/vector_search.py
annzhang-db: Yes, if the user explicitly is setting what other columns they want to include, we are assuming they don't want any other columns.

tests/databricks_ai_bridge/utils/test_vector_search.py
annzhang-db: Ah the actual values of the columns are being tested here, but yeah I can try to simplify this.

aravind-segu (approved):
Thank you for addressing the comments

PR #105, opened 2025-04-30
Support passing arbitrary kwargs through VectorSearchRetrieverTool

annzhang-db:
**Description of Changes**
- Add support for passing arbitrary kwargs through to VS similarity_search
- The user can either pass in kwargs during initialization, which will get passed all the way through, or pass in at query time
- In order for the LLM to be able to pass through arguments at query time, the user will have to update the args schema/tool spec/description appropriately for their use case.

**How is this tested?**
- [x] Unit tests
- [x] Manual test

OpenAI:
Screenshot 2025-05-06 at 2 39 34 PM

LlamaIndex:
Screenshot 2025-05-06 at 2 29 09 PM

Langchain:
Screenshot 2025-05-06 at 3 03 07 PM

### DIFF
--- a/integrations/langchain/src/databricks_langchain/vectorstores.py
+++ b/integrations/langchain/src/databricks_langchain/vectorstores.py
@@ -461,14 +461,16 @@ class DatabricksVectorSearch(VectorStore):
 
         signature = inspect.signature(self.index.similarity_search)
         kwargs = {k: v for k, v in kwargs.items() if k in signature.parameters}
-        kwargs.update({
-            "columns": self._columns,
-            "query_text": query_text,
-            "query_vector": query_vector,
-            "filters": filter,
-            "num_results": k,
-            "query_type": query_type,
-        })
+        kwargs.update(
+            {
+                "columns": self._columns,
+                "query_text": query_text,
+                "query_vector": query_vector,
+                "filters": filter,
+                "num_results": k,
+                "query_type": query_type,
+            }
+        )
         search_resp = self.index.similarity_search(**kwargs)
         return parse_vector_search_response(
             search_resp, self._retriever_schema, document_class=Document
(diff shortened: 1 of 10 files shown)


integrations/langchain/tests/unit_tests/test_vectorstores.py
```diff
@@ -304,24 +304,6 @@ def test_similarity_search_hybrid(index_name: str) -> None:
     assert all(["id" in d.metadata for d in search_result])
 
 
-def test_similarity_search_both_filter_and_filters_passed() -> None:
```
annzhang-db, line 307: Deleting this test since now we are passing `kwargs` through, so both `filter` and `filters` would show up. We are instead trusting that the arguments passed through are meaningful and correct.

integrations/langchain/src/databricks_langchain/vectorstores.py
```diff
@@ -458,14 +459,19 @@ def similarity_search_with_score(
...
             query_vector = self._embeddings.embed_query(query)  # type: ignore[union-attr]
 
-        search_resp = self.index.similarity_search(
-            columns=self._columns,
-            query_text=query_text,
-            query_vector=query_vector,
-            filters=filter,
-            num_results=k,
-            query_type=query_type,
+        signature = inspect.signature(self.index.similarity_search)
```
aravind-segu, line 462: shoudl we do something similar to this in openai and llamaindex as well so we dont pass arbitrary parameters to vector search?

aravind-segu (approved):
LGTM, thanks for the tests. Had a quick question

integrations/langchain/src/databricks_langchain/vectorstores.py
annzhang-db: Ah yeah, I meant to do that 😓 Thanks for catching this!

PR #106, opened 2025-05-01
Update docs to link to Unity Catalog OSS documentation

annzhang-db:
Screenshot 2025-05-01 at 10 08 24 AM
Screenshot 2025-05-01 at 10 08 31 AM


integrations/langchain/src/databricks_langchain/__init__.py
```diff
@@ -1,3 +1,18 @@
+"""
+**Re-exported Unity Catalog Utilities**
```
sunishsheth2009, line 2: Is there no way to just get the docs inlined rather than linking? Just curious. If not this is good as well :) 
cc @smurching since you had the initial idea

annzhang-db: Hmm, I couldn't find a good way to inline the docs but open to any suggestions 🙇‍♀️

sunishsheth2009 (approved):

PR #120, opened 2025-05-16
Optimize Genie Parse Result

aravind-segu:
Previously we were popping the last row and checking if its before the token limit. This is very slow and is causing 5 min wait times when the dataframe result is huge and our truncation limit is small. Use Binary search to speed up the process. Locally this made the pytest finish from 10secs -> 0.35 secs.

Added more unit tests with large cases of data


tests/databricks_ai_bridge/test_genie.py
```diff
@@ -232,6 +234,81 @@ def test_parse_query_result_trims_large_data():
...
+    df.columns = [col.strip() for col in df.columns]
+    df = df.applymap(lambda x: x.strip() if isinstance(x, str) else x)
+
+    # Drop the first column
+    df = df.drop(columns=[df.columns[0]])
+
+    return df
+
+
+@pytest.mark.parametrize("max_tokens", [100, 1000, 2000, 8000, 10000, 15000, 19000, 100000])
```
smurching: If we're doing a binary search, can we test some of the edge cases? e.g. 1 row is too large (MAX_TOKENS=1), odd numbers of rows exceeding the limit, etc

tests/databricks_ai_bridge/test_genie.py
```diff
@@ -232,6 +234,81 @@ def test_parse_query_result_trims_large_data():
...
+    # Re-join cleaned lines and parse
+    cleaned_markdown = "\n".join(lines)
+    df = pd.read_csv(StringIO(cleaned_markdown), sep="|")
+
+    # Strip whitespace from column names and values
+    df.columns = [col.strip() for col in df.columns]
+    df = df.applymap(lambda x: x.strip() if isinstance(x, str) else x)
+
+    # Drop the first column
+    df = df.drop(columns=[df.columns[0]])
```
smurching, line 255: Why do we drop the first column?

src/databricks_ai_bridge/genie.py
```diff
@@ -59,15 +59,33 @@ def _parse_query_result(resp) -> Union[str, pd.DataFrame]:
...
+            hi = mid
+        else:
+            lo = mid + 1
+
+    # Slice to the found limit
+    truncated_df = dataframe.iloc[:lo]
+    truncated_result = truncated_df.to_markdown()
+
+    # Double-check edge case if we overshot by one
+    if _count_tokens(truncated_result) > MAX_TOKENS_OF_DATA:
```
smurching, line 88: Would be nice to have a test for this branch

src/databricks_ai_bridge/genie.py
```diff
@@ -59,15 +59,33 @@ def _parse_query_result(resp) -> Union[str, pd.DataFrame]:
...
-        rows.pop()
-        query_result = pd.DataFrame(rows, columns=header).to_markdown()
-        tokens_used = _count_tokens(query_result)
 
-    return query_result.strip() if query_result else query_result
+    # If the full result fits, return it
+    if tokens_used <= MAX_TOKENS_OF_DATA:
+        return query_result.strip()
+
+    # Binary search to find the maximum number of rows that fit within the token limit
```
smurching: Maybe we can use https://docs.python.org/3/library/bisect.html#bisect.bisect_left to do this? To reduce testing/maintenance burden on our side

src/databricks_ai_bridge/genie.py
aravind-segu: The current tests already test this branch, because sometimes we overshoot by 1.

tests/databricks_ai_bridge/test_genie.py
aravind-segu: The first column is the index column in our markdown. So we are dropping it to compare it with the pd dataframe generated easier

smurching (approved):

PR #129, opened 2025-06-05
Databricks MCP Package

aravind-segu:
Introduce the `databricks-mcp` package with the Oauth Provider

### Usage Guide

```
import asyncio
from datetime import datetime, timedelta
from typing import Optional

from mcp.client.session import ClientSession
from mcp.client.streamable_http import streamablehttp_client
from databricks_mcp import DatabricksOAuthClientProvider
from databricks import sdk

workspace_client = sdk.WorkspaceClient(profile="dogfood")
async def query_mcp():
    async with streamablehttp_client(
        url="https://uc-mcp-server-aravind-6051921418418893.staging.aws.databricksapps.com/mcp/",
        auth=DatabricksOAuthClientProvider(workspace_client),
    ) as (read_stream, write_stream, _):
        async with ClientSession(read_stream, write_stream) as session:
            await session.initialize()
            tools = await session.list_tools()
            print(tools)


async def main():
    await query_mcp()


if __name__ == "__main__":
    asyncio.run(main())
```


mcp/src/databricks_mcp/oauth_provider.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class DatabricksTokenStorage(TokenStorage):
+    def __init__(self, workspace_client):
+        self.workspace_client = workspace_client
+
+    async def get_tokens(self) -> OAuthToken | None:
+        headers = self.workspace_client.config.authenticate()
+        token = headers["Authorization"].split("Bearer ")[1]
+        return OAuthToken(access_token=token, expires_in=60)
```
smurching: Is it guaranteed that these tokens expire in 60 min (is the unit here minutes)? Any way for us to tell? Otherwise picking a very short expiry time is probably best to ensure the token isn't inadvertently cached etc

mcp/src/databricks_mcp/oauth_provider.py
```diff
@@ -0,0 +1,26 @@
...
+from mcp.shared.auth import OAuthToken
+
+
+class DatabricksTokenStorage(TokenStorage):
+    def __init__(self, workspace_client):
+        self.workspace_client = workspace_client
+
+    async def get_tokens(self) -> OAuthToken | None:
+        headers = self.workspace_client.config.authenticate()
+        token = headers["Authorization"].split("Bearer ")[1]
```
smurching: Is this format guaranteed? Maybe worth checking/throwing a better error if it's not

smurching: There are other possible formats (though I think Databricks will usually be bearer auth): https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Authorization so probably a good idea to check

smurching (approved):
A few comments, LGTM once addressed!

mcp/src/databricks_mcp/oauth_provider.py
sunishsheth2009: Usually for Oauth 2.0 spec, `expires_in` should be in seconds. So I am assuming the above is 60 seconds

aravind-segu: yep its in seconds, added a default variable name for more clarity

PR #130, opened 2025-06-08
[WIP] Databricks MCP Client

aravind-segu:


mcp/src/databricks_mcp/mcp.py
```diff
@@ -0,0 +1,130 @@
...
+
+logger = logging.getLogger(__name__)
+
+# MCP URL types
+UC_FUNCTIONS_MCP = "uc_functions_mcp"
+VECTOR_SEARCH_MCP = "vector_search_mcp"
+GENIE_MCP = "genie_mcp"
+
+MCP_URL_PATTERNS = {
+    UC_FUNCTIONS_MCP: r"^/api/2\.0/mcp/functions/[^/]+/[^/]+$",
```
smurching, line 29: Seems this will only match if the string starts with `/api/2.0/...`, we should probably allow for `https://<workspace_hostname>/api/2.0/...`

mcp/src/databricks_mcp/mcp.py
```diff
@@ -0,0 +1,130 @@
...
+    VECTOR_SEARCH_MCP: r"^/api/2\.0/mcp/vector-search/[^/]+/[^/]+$",
+    GENIE_MCP: r"^/api/2\.0/mcp/genie/[^/]+$",
+}
+
+
+class DatabricksMCPClient:
+    def __init__(self, mcp_url: str, client: Optional[WorkspaceClient] = None):
```
smurching:
```suggestion
    def __init__(self, server_url: str, workspace_client: Optional[WorkspaceClient] = None):
```

Probably good to add a docstring as well outlining the expected format of `server_url`

mcp/src/databricks_mcp/mcp.py
```diff
@@ -0,0 +1,130 @@
...
+    def get_databricks_resources(self) -> List[DatabricksResource]:
+        """
+        This method is used to get the Databricks resources from the MCP URL.
+        It will return a list of Databricks resources.
+
+        Note that this method is only for databricks managed MCP URLs. It returns empty list for other MCP URLs
+        """
+        try:
+            if not self._is_valid_databricks_managed_mcp_url():
+                logger.warning(f"Ignoring invalid MCP URL: {self.mcp_url}")
```
smurching:
```suggestion
                logger.warning(f"MCP server URL {self.server_url} did not match any known Databricks-managed MCP server URL. Returning an empty list of resources")
```

smurching: Also, it's probably simpler if we just have an API like `_get_databricks_managed_mcp_url_type` which returns `None` if the type doesn't match any known ones, and then we handle that in the caller (instead of first doing this `_is_valid_databricks_managed_mcp_url` check and then calling `_get_databricks_managed_mcp_url_type`, needing to keep logic in sync across the two methods, etc)

mcp/src/databricks_mcp/mcp.py
```diff
@@ -0,0 +1,130 @@
...
+            if mcp_type == UC_FUNCTIONS_MCP:
+                return [DatabricksFunction(name) for name in normalized]
+            elif mcp_type == VECTOR_SEARCH_MCP:
+                return [DatabricksVectorSearchIndex(name) for name in normalized]
+
+            logger.warning(f"Unknown MCP type: {mcp_type}")
+            return []
+
+        except ValueError as ve:
+            logger.warning(f"Failed to parse MCP URL: {ve}")
```
smurching:
```suggestion
            logger.exception(f"Failed to parse MCP server URL {self.server_url}, returning empty list of resources")
```

This will cause a more detailed/useful stacktrace to be printed

smurching: Though pretty sure we don't need this ValueError handling if we make the change suggested above (handle the case of "this URL doesn't match any known Databricks-managed MCP server URL once)

mcp/src/databricks_mcp/mcp.py
```diff
@@ -0,0 +1,130 @@
...
+            return []
+        except Exception as e:
```
smurching, line 143: Curious why swallow exceptions generically here?

mcp/pyproject.toml
```diff
@@ -11,6 +11,7 @@ requires-python = ">=3.10"
 dependencies = [
     "mcp>=1.9.1",
     "databricks-sdk>=0.49.0",
+    "mlflow==3.1.0rc0"
```
smurching: Curious why do we need MLflow 3.1 RC here? Any chance we can rely on a non-RC version?

smurching: Curious why do we need MLflow 3.1 RC here? Any chance we can rely on a non-RC version?

mcp/src/databricks_mcp/mcp.py
```diff
@@ -0,0 +1,130 @@
...
+GENIE_MCP = "genie_mcp"
+
+MCP_URL_PATTERNS = {
+    UC_FUNCTIONS_MCP: r"^/api/2\.0/mcp/functions/[^/]+/[^/]+$",
+    VECTOR_SEARCH_MCP: r"^/api/2\.0/mcp/vector-search/[^/]+/[^/]+$",
+    GENIE_MCP: r"^/api/2\.0/mcp/genie/[^/]+$",
+}
+
+
+class DatabricksMCPClient:
```
smurching, line 36: Can we annotate this class as experimental using https://github.com/databricks/databricks-ai-bridge/blob/49bc442aebc04a8ad5d4c40d6d9aba8eba42e07a/src/databricks_ai_bridge/utils/annotations.py#L31?

mcp/src/databricks_mcp/mcp.py
```diff
@@ -0,0 +1,130 @@
...
+        """List the tools from the MCP URL."""
+        return asyncio.run(self._get_tools_async())
+
+    def call_tool(self, tool_name: str, arguments: dict[str, Any] | None = None) -> CallToolResult:
+        """Call the tool with the given name and input."""
+        return asyncio.run(self._call_tools_async(tool_name, arguments))
+
+    def get_databricks_resources(self) -> List[DatabricksResource]:
+        """
+        This method is used to get the Databricks resources from the MCP URL.
```
smurching:
```suggestion
        Returns a list of dependent Databricks resources for the current MCP server URL. 
        
          If authoring a custom code agent that runs tools from an MCP server, call this method and pass the returned resources to `mlflow.pyfunc.log_model` when logging your agent, to enable your agent to authenticate to the MCP server and run tools when deployed.
        
        Note that this method only supports detecting resources for Databricks-managed MCP servers. For custom MCP servers or other MCP server URLs, this method returns an empty list        
```

mcp/src/databricks_mcp/mcp.py
```diff
@@ -0,0 +1,130 @@
...
+        if not genie_id:
+            raise ValueError(f"Genie ID not found in: {self.mcp_url}")
+        return genie_id
+
+    def _normalize_tool_name(self, name: str) -> str:
+        """Convert double underscores to dots for compatibility."""
+        return name.replace("__", ".")
+
+    def list_tools(self) -> List[Tool]:
+        """List the tools from the MCP URL."""
```
smurching:
```suggestion
        """List tools for the current MCP server."""
```

smurching (commented):
A few comments but this mostly looks great! We should also get these documented in API docs once they're available (https://api-docs.databricks.com/python/databricks-ai-bridge/latest/databricks_ai_bridge.html)

sunishsheth2009 (commented):
Just looking at the code. IMO, getting the resources and making that easy makes sense. I am not sure if we should wrap MCP specific code into another code. I think it would be easier for users to just read and understand MCP docs. We ideally don't want to create a framework on top of MCP and keep on updating things once MCP updates things IMO. They might add new ways for auth and we don't want to support all of that with streamable http_client IMO. Resources in this PR makes sense to me. What do you guys think?

mcp/src/databricks_mcp/mcp.py
aravind-segu: oh the reason for this is we are performing a `urlparse` and then matching only on the path. So we are omitting the host name

mcp/src/databricks_mcp/mcp.py
aravind-segu: My idea was to not throw an error and instead just return a empty array if it was not a Databricks Managed MCP Server. Lmk if i should just throw an error instead.

Issue #128, opened 2025-06-04
Allow Custom Client to be Passed


PvH-SPC:
Hi,

If a user passes a custom client in via the constructor, it gets overwritten with the 'default' instance.  Could we please enable passing of a custom build constructor. 

https://github.com/databricks/databricks-ai-bridge/blob/b9afb5710a2dff875395bec9825e99a9154c8f2d/integrations/langchain/src/databricks_langchain/chat_models.py#L255

sunishsheth2009:
Hey @PvH-SPC, thank you for reporting this issue. I am curious what client would users want to use that support [predict](https://github.com/databricks/databricks-ai-bridge/blob/b9afb5710a2dff875395bec9825e99a9154c8f2d/integrations/langchain/src/databricks_langchain/chat_models.py#L283) out of the box? 
Also we are happy to take PRs if you have one with example usage as tests :) Thank you

shahrukh-shaik:
Can you see if below helps

```python
from mlflow.deployments import get_deploy_client
client = get_deploy_client("databricks")  # it uses databricks authentication environmental variables for auth, 
# such as DATABRICKS_HOST, DATABRICKS_TOKEN, DATABRICKS_CLIENT_ID/SECRET
chat_model = ChatDatabricks(
    endpoint="databricks-claude-3-7-sonnet",
    client=client,
    temperature=0.1,
    max_tokens=2560,
    extra_params={
        "thinking": {
            "type": "enabled",
            "budget_tokens": 1024
        }
    }
    # See https://python.langchain.com/api_reference/community/chat_models/langchain_community.chat_models.databricks.ChatDatabricks.html for other supported parameters
)
```

PR #131, opened 2025-06-14
Fix: Allow custom client to be passed to ChatDatabricks (Issue #128)

jayshrivastava0:
This commit addresses issue #128 by modifying the `__init__` method of the `ChatDatabricks` class to allow a custom client to be passed in via the constructor.

Previously, any client passed in the constructor would be overwritten by the default client. With this change:
- If a `client` is provided in `kwargs`, that client instance is used.
- Otherwise, the client is initialized using `get_deployment_client(self.target_uri)`.

A new unit test, `test_chat_model_with_custom_client`, has been added to `integrations/langchain/tests/unit_tests/test_chat_models.py` to verify this functionality. The test ensures that the provided custom client is correctly assigned and its methods are called when the `ChatDatabricks` instance makes predictions.

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -252,7 +252,10 @@ class ChatDatabricks(BaseChatModel):
 
     def __init__(self, **kwargs: Any):
         super().__init__(**kwargs)
-        self.client = get_deployment_client(self.target_uri)
+        if "client" in kwargs:
+            self.client = kwargs["client"]
+        else:
+            self.client = get_deployment_client(self.target_uri)
         self.extra_params = self.extra_params or {}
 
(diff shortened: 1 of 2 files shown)


aravind-segu:
@jayshrivastava0 what is the usecase for passing in the client directly? What kind of clients are you passing in here?

jayshrivastava0:
Hey @aravind-segu, great question. We enabled this to give developers more flexibility.

The two main use cases are:

1.  **Testing:** It allows you to inject a **mock client** for unit tests, so you can verify your logic without making actual network calls.
2.  **Custom Configuration:** You can pass in a client that's already configured with specific enterprise needs, like **custom authentication**, proxy settings, or advanced retry logic.

The client just needs to be an object that behaves like MLflow's `BaseDeploymentClient`.

PR #136, opened 2025-06-24
Fix Mlflow Tests

aravind-segu:
Mlflow deprecated `get_last_active_trace`. Fixing the breaking tests which were using this method


annzhang-db:
Ah I just merged a PR which should have fixed this - probably don't need it anymore?

aravind-segu:
> Ah I just merged a PR which should have fixed this - probably don't need it anymore?

Oh Awesome Thank u!!!!

PR #138, opened 2025-06-25
[s] Send Genie Reasoning + SQL Steps on Query

SabhyaC26:
## Changes
Adds `include_context` parameter to the Genie class. When set to true, the `_query_genie_as_agent` will send back the SQL command + reasoning from the Genie API call (which is already retrieved). It is set to false by default to maintain backwards compatibility!

## How it's tested
- Tested with new unit tests.
- Manual tests on local (query_reasoning, query_sql, and query_result are all there)
Screenshot 2025-06-30 at 4 14 25 PM

Manual test file:
```
#!/usr/bin/env python3
from databricks_langchain.genie import GenieAgent

GENIE_SPACE_ID = "01f03c3c97521c9db676ac39ec6a8166"


def test_genie_agent():
    """Test GenieAgent creation and basic usage"""
    print("=== Testing GenieAgent creation ===")
    
    try:
        client = WorkspaceClient()
        
        # Test creating GenieAgent
        agent = GenieAgent(
            genie_space_id=GENIE_SPACE_ID,
            genie_agent_name="TestAgent",
            description="Test agent for manual testing",
            include_context=True,
            client=client
        )
        
        # Test invoking the agent
        test_input = {
            "messages": [
                {"role": "user", "content": "On average, were customers bills higher in 2024 or 2025?"}
            ]
        }

        print("--------------------------------")
        print(f"Test input: {test_input}")
        print("--------------------------------")
        
        result = agent.invoke(test_input)
        print(f"Agent result type: {type(result)}")
        print(f"Agent result: {result}")
        
    except Exception as e:
        print(f"Error: {e}")
        return False
    
    print("✅ GenieAgent test completed successfully\n")
    return True

def main():
    """Run all tests"""
    print("🚀 Starting manual tests for Genie functionality")
    print(f"Using Genie Space ID: {GENIE_SPACE_ID}")
    print("=" * 50)
    
    if GENIE_SPACE_ID == "YOUR_GENIE_SPACE_ID_HERE":
        print("❌ Please replace GENIE_SPACE_ID with your actual Genie space ID")
        return
    
    tests = [
        test_genie_agent,
    ]
    
    passed = 0
    for test in tests:
        try:
            if test():
                passed += 1
        except Exception as e:
            print(f"Test {test.__name__} failed with exception: {e}\n")
    
    print("=" * 50)
    print(f"Tests completed: {passed}/{len(tests)} passed")

if __name__ == "__main__":
    main()
```

### DIFF
--- a/integrations/langchain/tests/unit_tests/test_genie.py
+++ b/integrations/langchain/tests/unit_tests/test_genie.py
@@ -181,45 +181,6 @@ def test_create_genie_agent_with_include_context(MockWorkspaceClient):
     mock_client.genie.get_space.assert_called_once()
 
 
-@patch("databricks.sdk.WorkspaceClient")
-def test_create_genie_agent_with_include_context_false(MockWorkspaceClient):
-    """Test creating a GenieAgent with include_context=False and verify it only returns result"""
-    mock_space = GenieSpace(
-        space_id="space-id",
-        title="Sales Space", 
-        description="description",
-    )
-    mock_client = MockWorkspaceClient.return_value
-    mock_client.genie.get_space.return_value = mock_space
-
-    # Create a proper GenieResponse instance
-    mock_genie_response = GenieResponse(
-        result="It is sunny.",
-        query="SELECT * FROM weather",
-        description="This is the reasoning for the query",
-    )
-
-    # Test with include_context=False (default)
-    agent = GenieAgent("space-id", "Genie", include_context=False, client=mock_client)
-    assert agent.description == "description"
-
-    # Create test input
-    input_data = {"messages": [{"role": "user", "content": "What is the weather?"}]}
-    
-    # Mock the ask_question method on the Genie instance
-    with patch("databricks_ai_bridge.genie.Genie.ask_question", return_value=mock_genie_response):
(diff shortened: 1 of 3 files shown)


integrations/langchain/tests/unit_tests/test_genie.py
```diff
@@ -107,17 +124,40 @@ def test_query_genie_with_client(mock_workspace_client):
...
+        description="description",
+    )
+    mock_client = MockWorkspaceClient.return_value
+    mock_client.genie.get_space.return_value = mock_space
+
+    # Test with include_context=True
+    agent = GenieAgent("space-id", "Genie", include_context=True, client=mock_client)
+    assert agent.description == "description"
+
+    mock_client.genie.get_space.assert_called_once()
```
bbqiu, line 181: nit: should we check that `include_context` is propagated? i could definitely be missing something, but as it stands, this test just makes sure that constructing GenieAgent w/ include_context=True doesn't error out?

SabhyaC26: yeah good point, updating the test

PR #141, opened 2025-07-01
Make ChatDatabricks compatible with ChatAgent and ResponsesAgent models

bbqiu:
needs more work to make fully responses API compatible

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -281,6 +281,9 @@ class ChatDatabricks(BaseChatModel):
     ) -> ChatResult:
         data = self._prepare_inputs(messages, stop, **kwargs)
         resp = self.client.predict(endpoint=self.model, inputs=data)  # type: ignore
+        if "messages" in resp:
+            return self._convert_chatagent_response_to_chat_result(resp)
+
         return self._convert_response_to_chat_result(resp)
 
     def _prepare_inputs(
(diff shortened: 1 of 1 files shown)


integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -228,6 +235,8 @@ class GetPopulation(BaseModel):
     """
     stream_usage: bool = False
     """Any extra parameters to pass to the endpoint."""
+    use_responses_api: bool = False
```
smurching, line 238: QQ does this match how LangChain generally configures use of the responses API? I think you found that it does, just wanted to confirm

smurching: (I would have thought it'd be a bit more extensible to do something like expose an enum here but it's more important to match what other LangChain APIs do)

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -854,3 +1095,86 @@ def _convert_dict_to_message_chunk(
...
+                tool_call_chunk(
+                    name=item.get("name"),
+                    args=item.get("arguments"),
+                    id=item.get("call_id"),
+                )
+            )
+            id = item.get("call_id")
+        elif item_type == "message":
+            id = item.get("id")
+            # handle duplication of text responses that have already been streamed
```
smurching: Curious about this piece, seems like we're assuming here that there's exactly one previous chunk with a `response.output_text.delta` type, and it contains the entire message content? Is it possible to have multiple deltas?

smurching (commented):
Generally LGTM! Just had one question

integrations/langchain/src/databricks_langchain/chat_models.py
bbqiu: https://github.com/langchain-ai/langchain/blob/b03e3262316923ec10505b4b61878c118ec66b5f/libs/partners/openai/langchain_openai/chat_models/base.py#L2027-L2028

yea this is the same way they do it

they also have some auto detection based on tools etc, but we don't have that yet

integrations/langchain/src/databricks_langchain/chat_models.py
bbqiu: discussed offline. we will continue dropping output item done message types, keeping annotations if there are any

PR #146, opened 2025-07-14
Adding databricks-ai-bridge package to databricks_mcp since its missi…

sunishsheth2009:
Adding databricks-ai-bridge package to databricks_mcp since its missing the dep

Its used here: https://github.com/databricks/databricks-ai-bridge/blob/main/databricks_mcp/src/databricks_mcp/mcp.py#L8
But its not installed as a dep.
So fixing that.

### DIFF
--- a/databricks_mcp/pyproject.toml
+++ b/databricks_mcp/pyproject.toml
@@ -11,6 +11,7 @@ requires-python = ">=3.10"
 dependencies = [
     "mcp>=1.9.1",
     "databricks-sdk>=0.49.0",
+    "databricks-ai-bridge>=0.4.2",
     "mlflow>=3.1"
 ]
 


databricks_mcp/pyproject.toml
```diff
@@ -11,6 +11,7 @@ requires-python = ">=3.10"
 dependencies = [
     "mcp>=1.9.1",
     "databricks-sdk>=0.49.0",
+    "databricks-ai-bridge>=0.4.2",
```
annzhang-db, line 14: Do we want to require the newest version of `databricks-ai-bridge>=0.6.0`? Or are we sure that `0.4.2` is good enough?

sunishsheth2009: I guess we just need the @experimental so far which is included from the initial release. I just copied form few integrations that had the 0.4.2 version in the requirements so replicated that

PR #148, opened 2025-07-16
Add DSPy integration in databricks-ai-bridge

chenmoneygithub:
This PR includes:

- Project setup, e.g., pyproject.toml, README, ...
- Add `DatabricksLM` which bypasses the issue that litellm creates a workspaceclient at every LM call, which causes authentication service to throttle.

### DIFF
--- a/integrations/dspy/tests/unit_tests/clients/test_databricks_lm.py
+++ b/integrations/dspy/tests/unit_tests/clients/test_databricks_lm.py
@@ -1,0 +1,52 @@
+from unittest.mock import MagicMock, patch
+
+import pytest
+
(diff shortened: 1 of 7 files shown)


integrations/dspy/tests/unit_tests/clients/test_databricks_lm.py
```diff
@@ -0,0 +1,52 @@
+from unittest.mock import MagicMock, patch
```
annzhang-db, line 1: Can you modify `.github/workflows` so that this test gets run? I think it isn't being run on this PR right now.

PR #149, opened 2025-07-17
LangChain GenieAgent - add message processor param

bbqiu:
- title
- context: https://databricks.slack.com/archives/C096Y4XJ6PQ/p1753864007452189

### DIFF
--- a/integrations/langchain/src/databricks_langchain/genie.py
+++ b/integrations/langchain/src/databricks_langchain/genie.py
@@ -24,26 +24,29 @@ def _query_genie_as_agent(
     genie: Genie,
     genie_agent_name,
     include_context: bool = False,
-    only_pass_last_message: bool = False,
+    only_pass_user_messages: bool = False,
 ):
     from langchain_core.messages import AIMessage
 
-    if only_pass_last_message:
+    if only_pass_user_messages:
         messages = input.get("messages", [])
         if messages:
-            if isinstance(messages[-1], dict):
-                message = messages[-1].get("content", "")
-            else:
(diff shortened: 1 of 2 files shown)


annzhang-db (approved):

anthonyivn2:
@bbqiu Wouldn't this make it so that the GenieAgent only takes in the `last message` passed in when starting a new conversation?

anthonyivn2:
If we are saying that we want to make it so that the GenieAgent is relying on the state of the underlying conversation, then we can't start a new conversation every time a message gets sent

integrations/langchain/src/databricks_langchain/genie.py
```diff
@@ -51,9 +62,49 @@ def GenieAgent(
...
 ):
-    """Create a genie agent that can be used to query the API. If a description is not provided, the description of the genie space will be used."""
+    """Create a genie agent that can be used to query the API. If a description is not provided, the description of the genie space will be used.
+
+    Args:
+        genie_space_id: The ID of the genie space to use
+        genie_agent_name: Name for the agent (default: "Genie")
+        description: Custom description for the agent
+        include_context: Whether to include query reasoning and SQL in the response
+        message_processor: Optional function to process messages before querying.
```
annzhang-db: Can you add something here to say "if message_processor is not defined, default prompt is added ..."

bbqiu:
tests are failing because of https://github.com/openai/openai-python/issues/2553#issuecomment-3177236079, an unintentional change that will be fixed. merging over the failures

PR #152, opened 2025-07-22
Extend the databricks_ai_bridge.genie.ask_question method to allow conversation ids

bradleyjamrozik-ozinga:
This extends the base Genie and GenieResponse class to include the conversation_id returned by the genie sdk. Often a genie space will ask clarifying questions and so the existing conversation needs to be continued. Example:

image

In the above case the genie space asked for my employee number. By providing it along with the conversation id returned from the previous response, the conversation was maintained and the genie could answer correctly.


bradleyjamrozik-ozinga:
Apologies, did not intend to add that last commit.

bradleyjamrozik-ozinga:
Resubmitted under #154

PR #153, opened 2025-07-22
Use OpenAI client to make LLM inference requests in ChatDatabricks

smurching:
Updates `databricks_langchain.ChatDatabricks` to use the OpenAI client to make inference requests. This should:
* Enable use cases like on-behalf-of-user-auth with ChatDatabricks
* Provide better handling for non-english characters
* Improve robustness since we'll benefit from things like the default timeout/retry policy in the OAI client
* Simplify maintenance of our implementation

Updated unit & integration tests to validate the change, and also manually ran [this notebook](https://e2-spog.staging.cloud.databricks.com/editor/notebooks/41269107306404?o=6051921418418893#command/41269107306417) using `ChatDatabricks` from this PR branch to validate changes


integrations/langchain/tests/utils/chat_models.py
```diff
@@ -121,13 +121,86 @@
 
 @pytest.fixture(autouse=True)
 def mock_client() -> Generator:
-    client = mock.MagicMock()
-    client.predict.return_value = _MOCK_CHAT_RESPONSE
-    client.predict_stream.return_value = _MOCK_STREAM_RESPONSE
-    with mock.patch("mlflow.deployments.get_deploy_client", return_value=client):
+    # Mock OpenAI client response objects
+    class MockOpenAIMessage:
```
smurching: Here we should probably just use the actual response objects; any reason we can't construct those directly?

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -827,7 +912,7 @@ def _convert_dict_to_message_chunk(
         return ToolMessageChunk(
             content=content, tool_call_id=_dict["tool_call_id"], id=_dict.get("id")
         )
-    elif role == "assistant":
+    elif role == "assistant" or (role is None and default_role == "assistant"):
```
smurching: Unnecessary change

integrations/langchain/tests/unit_tests/test_chat_models.py
```diff
@@ -350,18 +389,18 @@ def test_convert_message_to_dict_function() -> None:
 def test_convert_response_to_chat_result_llm_output(llm: ChatDatabricks) -> None:
     """Test that _convert_response_to_chat_result correctly sets llm_output."""
 
-    result = llm._convert_response_to_chat_result(_MOCK_CHAT_RESPONSE)
```
smurching, line 374: Why not still call this method?

integrations/llamaindex/tests/unit_tests/test_vector_search_retriever_tool.py
```diff
@@ -19,7 +19,7 @@
     FilterItem,
     VectorSearchRetrieverToolInput,
 )
-from llama_index.core.agent import ReActAgent
+from llama_index.core.agent.workflow import ReActAgent
```
smurching: Llamaindex tests were broken for reasons unrelated to this PR and failing on CI, fixed them here

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -219,8 +218,10 @@ class GetPopulation(BaseModel):
 
     model: str = Field(alias="endpoint")
     """Name of Databricks Model Serving endpoint to query."""
-    target_uri: str = "databricks"
-    """The target URI to use. Defaults to ``databricks``."""
+    target_uri: Optional[str] = None
+    """The target MLflow deployment URI to use. Deprecated: use profile instead."""
```
smurching:
```suggestion
    """The target MLflow deployment URI to use, e.g. "databricks" or "databricks://[profile_name]". Deprecated: use profile instead."""
```

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -290,114 +320,137 @@ def _generate(
...
+            "stream": stream,
             **self.extra_params,  # type: ignore
             **kwargs,
         }
+
+        # Format data based on whether we're using responses API or chat completions
         if self.use_responses_api:
+            # Responses API expects "input" parameter and has different supported parameters
             data["input"] = _convert_lc_messages_to_responses_api(messages)
+            # Responses API only supports temperature, not max_tokens, stop, or n
```
smurching, line 351: stop is supported in the REST API: https://platform.openai.com/docs/api-reference/chat/create#chat_create-stop but seemingly not by the client

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -290,114 +320,137 @@ def _generate(
...
-            item_type = item.get("type")
+        for item in response.output:
+            if item.type == "message":
+                for content in item.content:
+                    if content.type == "output_text":
+                        annotations = []
+                        if content.annotations:
+                            # Convert annotation objects to dictionaries
+                            annotations = [
+                                ann.model_dump() if hasattr(ann, "model_dump") else ann
```
smurching, line 387: Curious if/why this hasattr check is needed

smurching: Could maybe query an MAS or other endpoint to test this branch?

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -429,19 +483,51 @@ def _convert_chatagent_response_to_chat_result(self, response: Mapping[str, Any]
         message = AIMessage(content=response.get("messages"), id=response.get("id"))
         return ChatResult(generations=[ChatGeneration(message=message)])
 
-    def _convert_response_to_chat_result(self, response: Mapping[str, Any]) -> ChatResult:
-        generations = [
-            ChatGeneration(
-                message=_convert_dict_to_message(choice["message"]),
-                generation_info=choice.get("usage", {}),
+    def _convert_response_to_chat_result(self, response: Any) -> ChatResult:
```
smurching, line 488: Question for @bbqiu - seems like we never supported querying ChatAgent endpoints using ChatDatabricks, which is helpful for this PR :D since we don't have to worry about handling that case (the only types of endpoints we can query are OAI-compatible). Also notably it's not possible to do things like pass `custom_inputs` to agent serving endpoints via ChatDatabricks, but that's probably ok for now (it wasn't possible before either, and we could add support by passing a custom httpx client to the OpenAI client later, if needed)

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -68,12 +67,22 @@ class ChatDatabricks(BaseChatModel):
...
                 temperature=0,
                 max_tokens=500,
             )
 
+            # Using a WorkspaceClient instance for custom authentication
+            from databricks.sdk import WorkspaceClient
+
+            workspace_client = WorkspaceClient(host="...", token="...")
+            llm = ChatDatabricks(
+                model="databricks-meta-llama-3-1-405b-instruct",
```
bbqiu: nit: can we update these docstrings to use `databricks-claude-3-7-sonnet` since 405b is no longer hosted

integrations/langchain/src/databricks_langchain/chat_models.py
bbqiu: can use https://e2-dogfood.staging.cloud.databricks.com/ml/endpoints/agents_ml-bbqiu-annotationsv2?o=6051921418418893 to test

bbqiu: i think we should keep the hasattr check -- i think the openai client will try its best to create pydantic objects, but if it's a format it doesn't know, then it'll return the dict (ex. returning annotations that aren't the standard format i think? doesn't break the client, but it won't return the pydantic object)

integrations/langchain/src/databricks_langchain/chat_models.py
bbqiu: for input, we didn't have to do anything different (since it was the same as chatcompletions)
for output parsing, i added `_convert_chatagent_response_to_chat_result`

bbqiu: are we able to use the `extra_body` param to pass `custom_inputs`? I was able to use that to pass extra params to claude: https://e2-dogfood.staging.cloud.databricks.com/editor/notebooks/1118964840082062?o=6051921418418893#command/8070936494348751

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -429,19 +483,51 @@ def _convert_chatagent_response_to_chat_result(self, response: Mapping[str, Any]
...
+                    for tc in choice.message.tool_calls
+                ]
+
+            generation_info = {}
+            if choice.finish_reason:
+                generation_info["finish_reason"] = choice.finish_reason
+
+            generations.append(
+                ChatGeneration(
+                    message=_convert_dict_to_message(message_dict),
```
bbqiu, line 511: instead of manually reconstructing the dict, can we use model dump on the message dict?

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -1098,70 +1200,73 @@ def _convert_dict_to_message_chunk(
...
-            for content_item in item.get("content", []):
-                if content_item.get("type") == "output_text":
+            for content_item in item.content:
+                if content_item.type == "output_text":
                     if skip_duplicate_text:
-                        if content_item.get("annotations"):
-                            content.append({"annotations": content_item.get("annotations")})
+                        if content_item.annotations:
+                            # Convert annotation objects to dictionaries
+                            annotations = [ann.model_dump() for ann in content_item.annotations]
```
bbqiu: similar to above, it may be best to check if it's a pydantic class first

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -456,36 +542,33 @@ def _stream(
...
-                if chunk.get("delta"):
-                    chunk_delta = chunk["delta"]
-                    chunk_message = _convert_dict_to_message_chunk(
-                        chunk_delta, chunk_delta.get("role")
-                    )
-                    chunk = ChatGenerationChunk(message=chunk_message)
-                    yield chunk
-                elif chunk.get("choices"):
-                    choice = chunk["choices"][0]
+            # Use OpenAI client with chat completions API for streaming
```
bbqiu, line 557: i think this won't work with chatagent endpoints parsing-wise. could we test if the openai client throws errors and if not, then we should add back in the parsing logic from https://github.com/databricks/databricks-ai-bridge/pull/153/files#diff-d3df2a94788772d7305c1d27a61e633d133fc5dcb66fec30ce9b918136f67e44L472-L478

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -473,35 +473,43 @@ def _convert_responses_api_response_to_chat_result(self, response: Any) -> ChatR
...
         To accomodate this, we combine the messages into a single message, following LangChain convention.
 
         ex: https://github.com/langchain-ai/langchain/blob/2d3020f6cd9d3bf94738f2b6732b68acc55d9cce/libs/partners/openai/langchain_openai/chat_models/base.py#L3739
         """
-        message = AIMessage(content=response.get("messages"), id=response.get("id"))
+        # ChatAgent response structure may have messages or other fields
+        if hasattr(response, "messages") and response.messages:
+            messages = response.messages
+        elif hasattr(response, "model_dump") and "messages" in response.model_dump():
+            messages = response.model_dump()["messages"]
```
bbqiu: nit: i think none of the cases other than the 1st will happen right? we are only entering this if `messages` is an attr of response

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -456,36 +550,59 @@ def _stream(
...
+            first_chunk_role = None
+            stream = self.client.chat.completions.create(**data)  # type: ignore
+            for chunk in stream:
+                # Handle ChatAgent chunks that don't have choices but have delta
+                if hasattr(chunk, "choices") and chunk.choices is None and hasattr(chunk, "delta"):
+                    # ChatAgent streaming chunk format - has delta directly on chunk
+                    delta = chunk.delta
+
+                    if first_chunk_role is None:
+                        first_chunk_role = delta.get("role", "assistant")
```
bbqiu: nit: i think this is unnecessary since `ChatAgentChunk`'s will always have a role

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -494,27 +611,45 @@ def _stream(
...
+                                    "name": tc.function.name,
+                                    "arguments": tc.function.arguments,
+                                },
+                            }
+                            for tc in chunk_delta.tool_calls
+                        ]
+
                     chunk_message = _convert_dict_to_message_chunk(
-                        chunk_delta, first_chunk_role, usage=usage
+                        chunk_delta_dict, first_chunk_role, usage=usage
```
bbqiu: nit: could we instead follow the above and use `model_dump` rather than manually reconstructing the dict
https://github.com/databricks/databricks-ai-bridge/pull/153/files#diff-d3df2a94788772d7305c1d27a61e633d133fc5dcb66fec30ce9b918136f67e44R508

bbqiu (approved):
LGTM after addressing comments -- thank you for doing such thorough testing!

PR #157, opened 2025-08-05
Genie Deprecation Warning

aravind-segu:
In this release warn users that genie auto truncation will be deprecated. In the next release only truncate data if the flag `truncate_results` is set to True.

### DIFF
--- a/src/databricks_ai_bridge/genie.py
+++ b/src/databricks_ai_bridge/genie.py
@@ -1,6 +1,7 @@
 import bisect
 import logging
 import time
+import warnings
 from dataclasses import dataclass
 from datetime import datetime
 from typing import Optional, Union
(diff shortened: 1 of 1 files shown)


src/databricks_ai_bridge/genie.py
```diff
@@ -88,11 +89,20 @@ def is_too_big(n):
     if _count_tokens(truncated_result) > MAX_TOKENS_OF_DATA:
         truncated_result = truncated_df.iloc[:-1].to_markdown()
 
+    if not truncate_results:
+        warnings.warn(
+            "Automatic data truncation in Genie will be deprecated in a future version. Currently your data is being truncated by default. To preserve this behavior, set truncate_results=True when initializing the Genie class.",
```
smurching:
```suggestion
            "Detected large Genie output, truncating it to better fit in LLM context windows. Automatic result truncation in Genie is deprecated and will be disabled by default in a future release; we recommend truncating large Genie results in your agent code instead, if needed. To keep automatic truncation for large Genie outputs enabled, set truncate_results=True when initializing the Genie class.",
```

smurching (approved):
One comment on wording otherwise LGTM!

PR #158, opened 2025-08-06
Add integration test for databricks-dspy

chenmoneygithub:
The setup is a bit strange, basically:

1. We create a job in workspace https://ai-oss-ecosystem-integration-testing.cloud.databricks.com/, the notebook under which contains the actual testing code.
2. We add an integration test .py file in databricks-ai-bridge, which just triggers the job and check the job status. On any sort of failure in the job, the test will fail.
3. In another internal repo: https://github.com/databricks-eng/ai-oss-integration-tests-runner, we create github action that executes .py file created in step 2. This is necessary because storing the credentials in internal repos is much safer than this public repo.
4. On PRs, we may need to run the test manually though, which is fine IMO. 

This PR handles step 2 specifically.

### DIFF
--- a/integrations/dspy/tests/integration_tests/test_databricks_lm.py
+++ b/integrations/dspy/tests/integration_tests/test_databricks_lm.py
@@ -1,0 +1,45 @@
+import os
+from datetime import timedelta
+
+import pytest
+from databricks.sdk import WorkspaceClient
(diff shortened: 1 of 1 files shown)


chenmoneygithub:
jenkins merge

PR #164, opened 2025-08-11
fix breaking change in openai sdk update

bbqiu:
they removed `ChatCompletionMessageToolCall` and we should use `ChatCompletionMessageFunctionToolCall` instead

bumping min version of openai SDK as well to make sure nothing else breaks


bbqiu:
closing bc tests are failing because of https://github.com/openai/openai-python/issues/2553#issuecomment-3177236079, an unintentional change that will be fixed. merging over the failures

PR #165, opened 2025-08-12
Add timeout and retry params to ChatDatabricks

smurching:
Add standard langchain timeout and retry params to ChatDatabricks. Depends on a Databricks SDK release containing the changes in https://github.com/databricks/databricks-sdk-py/pull/1025

### DIFF
--- a/integrations/langchain/pyproject.toml
+++ b/integrations/langchain/pyproject.toml
@@ -23,7 +23,6 @@ dependencies = [
 dev = [
   "pytest",
   "typing_extensions",
-  "databricks-sdk>=0.63.0",
   "ruff==0.6.4",
 ]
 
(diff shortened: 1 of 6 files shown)


integrations/langchain/pyproject.toml
```diff
@@ -17,13 +17,14 @@ dependencies = [
     "unitycatalog-langchain[databricks]>=0.2.0",
     "databricks-connect>=16.1.1,<16.4",
     "openai>=1.97.1",
+    "databricks-sdk>=0.63.0",
```
smurching: TODO: update this after SDK releases

integrations/langchain/pyproject.toml
```diff
@@ -17,13 +17,14 @@ dependencies = [
...
 ]
 
 [project.optional-dependencies]
 dev = [
   "pytest",
   "typing_extensions",
-  "databricks-sdk>=0.34.0",
+  "databricks-sdk>=0.63.0",
```
bbqiu: if we add it to the normal list of dependencies, i think we can delete it from the dev reqs

PR #166, opened 2025-08-13
Return better MCP Errors for Databricks Apps

aravind-segu:
Currently the errors returned from Databricks MCP Client are not very user friendly. This PR processes some common errors by querying the MCP URL again upon errors and returns a more user friendly message.

### Case 1
Query Databricks MCP App with a PAT Token

#### Before
image

#### After
image

### Case 2
Query Databricks MCP App with a SP where the SP does not have permission to the Databricks App

#### Before
image

#### After
image

### Case 3
Query a valid Databricks MCP App where a mcp app is not hosted

#### Before
image

#### After
image

### Case 4
Databricks MCP App itself returns an error

#### Before
image

#### After
image

### Case 5
Successful Execution
#### Before
image

#### After
image

### DIFF
--- a/databricks_mcp/src/databricks_mcp/mcp.py
+++ b/databricks_mcp/src/databricks_mcp/mcp.py
@@ -1,4 +1,5 @@
 import asyncio
+import json
 import logging
 import re
 from functools import wraps
@@ -35,7 +36,7 @@ MCP_URL_PATTERNS = {
 
 
 def _handle_mcp_errors(func: Callable) -> Callable:
-    """Decorator to handle MCP connection errors consistently for async functions."""
+    """Decorator to handle MCP connection errors for sync wrapper methods."""
 
     def _process_mcp_error(client_instance, error: Exception) -> None:
         """Process and enhance MCP connection errors with better context."""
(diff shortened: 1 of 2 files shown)


databricks_mcp/src/databricks_mcp/mcp.py
```diff
@@ -32,6 +35,76 @@
...
+        # Auth errors to Databricks Apps are redirected to a login page; a 302 often indicates an auth issue
+        if response.status_code == 302:
+            raise PermissionError(
+                "Access denied to the MCP server. When accessing an MCP server hosted on a Databricks App please ensure you are using a valid OAuth token. "
+                "If using a Service Principal, ensure that the service principal has query permissions on the Databricks App. "
+                "For more information refer to the documentation here: "
+                "https://docs.databricks.com/aws/en/generative-ai/mcp/custom-mcp?language=Local+environment#connect-to-the-custom-mcp-server"
+            ) from None
+        # Not finding a `/initialize` endpoint means the MCP server is not running or the endpoint is not correct
+        elif response.status_code == 404:
```
sunishsheth2009, line 89: I think we just return back MCPError for these use cases instead of response.status_code being different? That should help show the errors a bit better? What do you think?

PR #167, opened 2025-08-13
Add DatabricksRM to databricks-dspy

chenmoneygithub:
This is an adapted version from https://github.com/stanfordnlp/dspy/blob/d8b7d1c176eefee9b43302f11cc13b92bcc5e179/dspy/retrievers/databricks_rm.py#L29, and the following changes are made:

- Only keep the path that uses `databricks-sdk` for simplicity, since `databricks-dspy` requires `databricks-sdk` as a hard dependency.
- Support default auth through `.databrickscfg`.
- Some style cleanup.

Tested with unit test internally: 

```
import databricks_dspy

retriever = databricks_dspy.DatabricksRM(
    databricks_index_name="kbqa.moon_fb_managed.index_finance_reports_97g28o",
    text_column_name="chunk_text",
    docs_uri_column_name="doc_uri",
    docs_id_column_name="doc_uri",
    columns=[],
    use_with_databricks_agent_framework=False,
    k=3,
)

retriever(query="financial advice")
```

Produces: 

image

### DIFF
--- a/integrations/dspy/src/databricks_dspy/retrievers/databricks_rm.py
+++ b/integrations/dspy/src/databricks_dspy/retrievers/databricks_rm.py
@@ -1,0 +1,380 @@
+import json
+import os
+from dataclasses import dataclass
+from importlib.util import find_spec
+from typing import Any
+
+import requests
+
+import dspy
+from dspy.primitives.prediction import Prediction
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class Document:
+    page_content: str
+    metadata: dict[str, Any]
+    type: str
+
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "page_content": self.page_content,
+            "metadata": self.metadata,
+            "type": self.type,
+        }
+
+
+class DatabricksRM(dspy.Retrieve):
+    """
+    A retriever module that uses a Databricks Mosaic AI Vector Search Index to return the top-k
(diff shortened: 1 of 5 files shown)


integrations/dspy/src/databricks_dspy/retrievers/databricks_rm.py
```diff
@@ -0,0 +1,377 @@
...
+        self.k = k
+        self.docs_id_column_name = docs_id_column_name
+        self.docs_uri_column_name = docs_uri_column_name
+        self.text_column_name = text_column_name
+        self.use_with_databricks_agent_framework = use_with_databricks_agent_framework
+        if self.use_with_databricks_agent_framework:
+            try:
+                import mlflow
+
+                mlflow.models.set_retriever_schema(
```
annzhang-db, line 145: We were planning on deprecating usage of `set_retriever_schema`, need figure out the details of this but this is something we are not planning to continue using

annzhang-db: I see you are setting primary_key="doc_id", seems like the only significant thing here? (I believe this is instead of "chunk_id")

chenmoneygithub: This part is a bit strange - this code was directly asked by DSPy SAs to get `DatabricksRM` to work well with agent platform's review app.

PR #173, opened 2025-08-22
Remove Databricks Connect Dependency

aravind-segu:
UC AI already restricts the databricks-connect package between the min and max version. We do not need to restrict it again here.


bbqiu (commented):
we can update this after the uc-ai release so that it won't be possible to install the wrong version of uc-ai (which has an outdated version of db connect pinned) when installing the latest db-ai-bridge packages

bbqiu (approved):
lgtm after manual testing: https://e2-dogfood.staging.cloud.databricks.com/browse/folders/2041828871065170?o=6051921418418893

PR #174, opened 2025-08-26
Add workspace client as optional arg to dspy.DatabricksRM

kenctrl:
Agent Bricks is working on moving KA/MA models from the data plane to the control plane. Part of this work involves maintaining a workspace client, which is currently not possible with `dspy.DatabricksRM`.

This PR adds the workspace client as an optional argument + additional verification to fail fast, plus some unit tests.

### DIFF
--- a/integrations/dspy/src/databricks_dspy/retrievers/databricks_rm.py
+++ b/integrations/dspy/src/databricks_dspy/retrievers/databricks_rm.py
@@ -6,6 +6,7 @@ from typing import Any
 
 import dspy
 from dspy.primitives.prediction import Prediction
+from databricks.sdk import WorkspaceClient
 
 logger = logging.getLogger(__name__)
 
@@ -65,6 +66,7 @@ class DatabricksRM(dspy.Retrieve):
             docs_id_column_name="id",
             text_column_name="field2",
             k=3,
+            workspace_client=w,
         )
         ```
 
@@ -90,6 +92,7 @@ class DatabricksRM(dspy.Retrieve):
         docs_uri_column_name: str | None = None,
         text_column_name: str = "text",
         use_with_databricks_agent_framework: bool = False,
+        workspace_client: WorkspaceClient | None = None,
     ):
         """
         Args:
@@ -122,6 +125,8 @@ class DatabricksRM(dspy.Retrieve):
                 containing document text to retrieve.
             use_with_databricks_agent_framework (bool): Whether to use the `DatabricksRM` in a way
                 that is compatible with the Databricks Mosaic Agent Framework.
+            workspace_client (Optional[WorkspaceClient]): The workspace client to use. If not 
+                provided, a new one will be created with default credentials from the environment.
         """
         super().__init__(k=k)
         self.databricks_token = databricks_token or os.environ.get("DATABRICKS_TOKEN")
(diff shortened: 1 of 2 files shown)


integrations/dspy/src/databricks_dspy/retrievers/databricks_rm.py
```diff
@@ -154,6 +159,42 @@ def __init__(
...
+            else:
+                # fallback to default authentication, i.e., using `~/.databrickscfg` file.
+                self.workspace_client = WorkspaceClient()
+                logger.info(
+                    "Creating Databricks workspace client using credentials from `~/.databrickscfg` file."
+                )
+
+        try:
+            # If credentials are invalid, `w.current_user.me()` will throw an error.
+            self.workspace_client.current_user.me()
```
bbqiu: nice touch!

bbqiu (commented):
LGTM, but the owner of this package is technically the DsPy team (they handle releasing this project, so will let them handle stamping)

integrations/dspy/src/databricks_dspy/retrievers/databricks_rm.py
```diff
@@ -65,6 +66,7 @@ class DatabricksRM(dspy.Retrieve):
             docs_id_column_name="id",
             text_column_name="field2",
             k=3,
+            workspace_client=w,
```
chenmoneygithub, line 73: This `w` is not present in the example code, let's create the instance first.

integrations/dspy/tests/unit_tests/retrievers/test_databricks_rm.py
```diff
@@ -49,190 +49,328 @@ def mock_vector_search_response_with_uri():
...
-@patch("databricks.sdk.WorkspaceClient")
-def test_databricks_rm_forward_string_query(mock_workspace_client, mock_vector_search_response):
+def test_databricks_rm_forward_string_query(mock_vector_search_response):
     """Test forward method with string query and ANN search."""
-    mock_client = MagicMock()
-    mock_workspace_client.return_value = mock_client
-    mock_client.vector_search_indexes.query_index.return_value.as_dict.return_value = (
-        mock_vector_search_response
-    )
+    with patch("databricks_dspy.retrievers.databricks_rm.WorkspaceClient") as MockWSClient:
```
chenmoneygithub: Do we require this change? `@patch` was used so that we don't need to patch the specific class inside `databricks_dspy` and reduce the nesting level.

integrations/dspy/tests/unit_tests/retrievers/test_databricks_rm.py
```diff
@@ -49,190 +49,328 @@ def mock_vector_search_response_with_uri():
...
 
     rm = DatabricksRM(
         databricks_index_name="test_index",
-        databricks_client_id="test_client_id",
-        databricks_client_secret="test_client_secret",
+        workspace_client=mock_workspace_client,
     )
 
-    rm("test query")
+    result = rm("test query")
```
chenmoneygithub, line 216: let's add a verification that `current_user.me` is called on this `mock_workspace_client`.

integrations/dspy/tests/unit_tests/retrievers/test_databricks_rm.py
```diff
@@ -49,190 +49,328 @@ def mock_vector_search_response_with_uri():
...
+        MockWSClient.assert_called_once_with(
+            client_id="test_client_id",
+            client_secret="test_client_secret",
+        )
 
 
 def test_databricks_rm_invalid_query_type():
     """Test forward method with invalid query type."""
-    rm = DatabricksRM(databricks_index_name="test_index")
+    with patch("databricks_dspy.retrievers.databricks_rm.WorkspaceClient") as MockWSClient:
```
chenmoneygithub: Same with these patch. This is nonblocking though, if this is required feel free to close the comment.

integrations/dspy/src/databricks_dspy/retrievers/databricks_rm.py
```diff
@@ -154,6 +159,42 @@ def __init__(
                     "library. Please install mlflow via `pip install mlflow`."
                 ) from None
 
+        if workspace_client:
+            self.workspace_client = workspace_client
+        else:
+            if databricks_client_secret and databricks_client_id:
```
chenmoneygithub: Can we keep these inside the actual query method? Ideally we want to keep constructor lightweight, and only authenticate at usage time (lazy auth). Also it's possible that auth is not yet set at construction time.

integrations/dspy/src/databricks_dspy/retrievers/databricks_rm.py
```diff
@@ -191,6 +204,54 @@ def _get_extra_columns(self, item: dict[str, Any]) -> dict[str, Any]:
...
         return extra_columns
 
+    @property
+    def workspace_client(self) -> WorkspaceClient:
+        """Get or create the workspace client lazily."""
+        # If client is already validated, return it
+        if self._workspace_client is not None and self._workspace_client_validated:
+            return self._workspace_client
+
+        if self._workspace_client is not None:
```
chenmoneygithub: nit: `if self.workspace_client:`

integrations/dspy/src/databricks_dspy/retrievers/databricks_rm.py
```diff
@@ -191,6 +204,54 @@ def _get_extra_columns(self, item: dict[str, Any]) -> dict[str, Any]:
             }
         return extra_columns
 
+    @property
+    def workspace_client(self) -> WorkspaceClient:
+        """Get or create the workspace client lazily."""
+        # If client is already validated, return it
+        if self._workspace_client is not None and self._workspace_client_validated:
```
chenmoneygithub: This is actually risky, if the credential expires (rare for PAT, but common for OAuth), this  `_workspace_client_validated` should be reset.

I would drop the `w.current_user.me()` check and `_workspace_client_validated` attribute, instead just let the actual query `self.workspace_client.vector_search_indexes.query_index` handle validation, which emits a clear error message on invalid auth as I remember. We can also remove this property and move the construction of `workspace_client` into `__init__` since that's pretty much one off.

kenctrl: Sounds good, just made the changes to move ws initialization into the constructor and have `query_index` itself throw upon invalid auth.

bbqiu (approved):
stamped to merge

PR #179, opened 2025-09-09
add return_pandas boolean

coltonpeltier-db:
add ability to get pandas dataframe back or markdown when calling `ask_question`. Default to markdown for backwards compatibility.

### DIFF
--- a/src/databricks_ai_bridge/genie.py
+++ b/src/databricks_ai_bridge/genie.py
@@ -29,7 +29,7 @@ class GenieResponse:
 
 
 @mlflow.trace(span_type="PARSER")
-def _parse_query_result(resp, truncate_results) -> Union[str, pd.DataFrame]:
+def _parse_query_result(resp, truncate_results, return_pandas : bool = False) -> Union[str, pd.DataFrame]:
     output = resp["result"]
     if not output:
         return "EMPTY"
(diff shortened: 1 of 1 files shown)


src/databricks_ai_bridge/genie.py
```diff
@@ -93,12 +95,14 @@ def is_too_big(n):
...
-        truncated_result = truncated_df.iloc[:-1].to_markdown()
-    return truncated_result
+    if _count_tokens(truncated_result.to_markdown()) > MAX_TOKENS_OF_DATA:
+        truncated_result = truncated_df.iloc[:-1]
+    
+    final_output = truncated_result if return_pandas else truncated_result.to_markdown().strip()
+    return final_output
 
 
 class Genie:
```
aravind-segu, line 108: Can we add return_pandas as a class level parameter here similar to truncate results.

src/databricks_ai_bridge/genie.py
```diff
@@ -62,16 +62,18 @@ def _parse_query_result(resp, truncate_results) -> Union[str, pd.DataFrame]:
         rows.append(row)
 
     dataframe = pd.DataFrame(rows, columns=header)
+    # Default to dataframe
+    query_result = dataframe
```
aravind-segu: To make the logic easier here, do we need to truncate results when we are returning a dataframe? Tokens doesnt really make sense when we are returning a dataframe and not a string. 

We can just say if(return_pandas): return dataframe

If not then keep the existing truncation logic

src/databricks_ai_bridge/genie.py
coltonpeltier-db: sure, makes sense to me!

src/databricks_ai_bridge/genie.py
coltonpeltier-db: yep, can do so

src/databricks_ai_bridge/genie.py
```diff
@@ -93,17 +98,23 @@ def is_too_big(n):
     if len(truncated_df) == 0:
         return ""
 
-    truncated_result = truncated_df.to_markdown()
```
aravind-segu: Can we undo these changes, there should be no changes here

src/databricks_ai_bridge/genie.py
```diff
@@ -62,13 +64,16 @@ def _parse_query_result(resp, truncate_results) -> Union[str, pd.DataFrame]:
         rows.append(row)
 
     dataframe = pd.DataFrame(rows, columns=header)
+    # Default to dataframe
+    query_result = dataframe
```
aravind-segu, line 69: [nit] can we just do:

if return_pandas:
    return dataframe


This is just to have better code instead of checking if a boolean is False

PR #181, opened 2025-09-12
Create automatic workflow for DSPy release

chenmoneygithub:
Basically the standard workflow will be:

1. Create a release branch for databricks-dspy.
2. Update the project version inside `pyproject.toml` in the release branch.
3. Create a github tag from the release branch. The tag must be of format `'databricks-dspy-v*'` so that this workflow is picked up.
4. The workflow "Build and Release databricks-dspy" will be automatically triggered. It automatically builds and uploads the package to pypi.

Now back to what this workflow is doing:

1. Grab the code from the source branch, i.e., the branch from which we create the github tag.
2. Grab the version from the github tag `'databricks-dspy-v*'`.
3. Verify that the pyproject.toml has the same version as 2.
4. Build the wheel for pypi
5. Change the name locally to `databricks-dspy-test` and appends a unique suffix to the version in pyproject.toml, then build the wheel for testpypi
6. Uplaod wheel to testpypi, and verify everything works.
7. Upload wheel to pypi and wrap up.

See screenshot below for a sample run. Publishing to pypi is skipped because it only gets triggered on the upstream repo.

image

### DIFF
--- a/.github/workflows/databricks-dspy-release.yml
+++ b/.github/workflows/databricks-dspy-release.yml
@@ -59,27 +59,42 @@ jobs:
           fi
           echo "Version validation passed: $PYPROJECT_VERSION"
 
-      - name: Update package name and version for testing
+      - name: Build original package for PyPI
         run: |
+          python -m build
+          # Save original package for PyPI
+          mkdir -p dist-pypi
+          cp dist/* dist-pypi/
+
+      - name: Build test package for TestPyPI
+        run: |
+          # Clean previous build
+          rm -rf dist/*
           # Update package name for TestPyPI
           sed -i 's/^name *= *".*"/name = "databricks-dspy-test"/' pyproject.toml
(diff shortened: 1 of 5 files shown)


.github/workflows/databricks-dspy-release.yml
```diff
@@ -0,0 +1,184 @@
...
+          pip install .[dev]
+
+      - name: Lint with ruff
+        run: |
+          ruff check .
+          ruff format --check .
+
+      - name: Run tests
+        run: |
+          pytest tests/unit_tests
```
bbqiu, line 50: non-blocking, but do we want to check on the integration tests that you have set up in the oss runner repo?

chenmoneygithub: Good call, the integration test cannot be triggered from this repo though. It requires PAT from the AI oss team's staging workspace.

PR #184, opened 2025-09-17
[ML-57490] ChatDatabricks: GPT-OSS workaround

bbqiu:
- because GPT OSS gives back list content items, we need to serialize it into a string to match what downstream consumers of ChatDatabricks runnables expect (ex. output parsers or other lambdas)


requirements/lint-requirements.txt
```diff
@@ -1 +1 @@
-ruff==0.6.4
+ruff==0.12.10
```
bbqiu, line 1: updated to fix lint discrepancies b/w versions

this now matches MLflow's version

sunishsheth2009 (approved):

PR #186, opened 2025-09-18
bump versions after 0.8.0 releases

bbqiu:


integrations/langchain/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "databricks-langchain"
-version = "0.8.0.dev0"
+version = "0.8.1.dev0"
```
sunishsheth2009: Can we bump this to 0.9.0.dev0. This applies to all the integrations and databricks-ai-bridge as well. 

We do minor versions since patch is mostly for hotfix. We ideally don't increment here for hotfix, we can directly create a hotfix branch from `0.8.0.dev0` and cherry-pick commits.

integrations/openai/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "databricks-openai"
-version = "0.6.0.dev0"
```
sunishsheth2009, line 3: 0.7.0.dev0

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "databricks-ai-bridge"
-version = "0.8.0.dev0"
```
sunishsheth2009, line 3: 0.9.0.dev0

PR #187, opened 2025-09-19
Add OBO Debug Mode

aravind-segu:
If `OBO_DEBUG_MODE` environment variable is set then add logging statement for debugging purposes.

Test


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "databricks-ai-bridge"
-version = "0.8.0.dev0"
+version = "0.8.1"
```
sunishsheth2009: Can we do this when we release and cut a branch?

aravind-segu: Oh yes sorry not ready for review yet, just needed this to install it on serving and test

sunishsheth2009 (approved):

PR #192, opened 2025-09-26
ChatDatabricks: turn on stream usage by default

bbqiu:
context: https://databricks.slack.com/archives/C065NC65Q9F/p1758832528624079?thread_ts=1758278104.748729&cid=C065NC65Q9F
from integration test:
screenshot 2025-09-25-21-26-31-Arc

by default, we will now emit an empty AIMessageChunk at the very end with usage info


annzhang-db (approved):
thanks for fixing this!

PR #194, opened 2025-09-30
Higher visibility of Genie API's execution steps in MLflow traces

maxmarcussen-db:
Polls more frequently (from every 5 to every 0.5 seconds) to get [statuses](https://docs.databricks.com/api/workspace/genie/getmessage) of requests to Genie API as they change, and creates new spans for each. The goal is to provide increased traceability into Genie's actual thinking steps as they execute, rather than a generic "poll_result" trace. Added new unit tests to verify this functionality works, and tested locally.

New timeline looks like this:
Screenshot 2025-09-30 at 3 31 15 PM

### DIFF
--- a/src/databricks_ai_bridge/genie.py
+++ b/src/databricks_ai_bridge/genie.py
@@ -10,8 +10,15 @@ import pandas as pd
 from databricks.sdk import WorkspaceClient
 
 MAX_TOKENS_OF_DATA = 20000
-MAX_ITERATIONS = 500
+MAX_ITERATIONS = 2500  # for 250 s total
 ITERATION_FREQUENCY = 0.1  # seconds
+
+TERMINAL_STATES = {
+    "COMPLETED",
+    "FAILED",
+    "CANCELLED",
+    "QUERY_RESULT_EXPIRED",
+}
 
 
 # Define a function to count tokens
@@ -136,6 +143,28 @@ def _truncate_result(dataframe):
     return truncated_result
 
 
+def _end_current_span(client, parent_trace_id, current_span, final_state, error=None):
+    """helper function to safely end a span with exception handling."""
+
+    if current_span is None:
+        return None
+
+    try:
+        attributes = {"final_state": final_state}
+        if error is not None:
+            attributes["error"] = error
+
+        client.end_span(
+            trace_id=parent_trace_id,
+            span_id=current_span.span_id,
+            attributes=attributes,
+        )
+    except mlflow.exceptions.MlflowTracingException as e:
+        logging.warning(f"Failed to end span for {final_state}: {e}")
+
+    return None
+
+
 class Genie:
     def __init__(
         self,
@@ -240,46 +269,40 @@ class Genie:
 
                     # On status change: end previous span, start a new one (excluding terminal states)
                     if current_status != last_status:
-                        # END previous span
-                        if current_span is not None:
-                            client.end_span(
-                                trace_id=parent_trace_id,
-                                span_id=current_span.span_id,
-                                attributes={"final_state": last_status},
-                            )
-                            current_span = None
-
                         # START new span for non-terminal states
-                        if current_status not in {
-                            "COMPLETED",
-                            "FAILED",
-                            "CANCELLED",
-                            "QUERY_RESULT_EXPIRED",
-                        }:
-                            current_span = client.start_span(
-                                name=current_status.lower(),
-                                trace_id=parent_trace_id,
-                                parent_id=parent_span_id,
-                                span_type="CHAIN",
-                                attributes={
-                                    "state": current_status,
-                                    "conversation_id": conversation_id,
-                                    "message_id": message_id,
-                                },
+                        if current_status not in TERMINAL_STATES:
+                            # END previous span
+                            current_span = _end_current_span(
+                                client, parent_trace_id, current_span, last_status
                             )
+                            # START new span
+                            try:
+                                current_span = client.start_span(
+                                    name=current_status.lower(),
+                                    trace_id=parent_trace_id,
+                                    parent_id=parent_span_id,
+                                    span_type="CHAIN",
+                                    attributes={
+                                        "state": current_status,
+                                        "conversation_id": conversation_id,
+                                        "message_id": message_id,
+                                    },
+                                )
+                            except mlflow.exceptions.MlflowTracingException as e:
+                                logging.warning(f"Failed to create span for {current_status}: {e}")
+                                current_span = None
 
                         logging.debug(f"Status: {last_status} → {current_status}")
                         last_status = current_status
 
                     if current_status == "COMPLETED":
                         # close any open child timeline span before ending
-                        if current_span is not None:
-                            client.end_span(
-                                trace_id=parent_trace_id,
-                                span_id=current_span.span_id,
-                                attributes={"final_state": current_status},
-                            )
-                            current_span = None
+                        current_span = _end_current_span(
+                            client,
+                            parent_trace_id,
+                            current_span,
+                            current_status,
+                        )
 
                         attachment = next((r for r in resp["attachments"] if "query" in r), None)
                         if attachment:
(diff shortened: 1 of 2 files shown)


src/databricks_ai_bridge/genie.py
```diff
@@ -213,45 +214,136 @@ def poll_query_results(
...
-                    logging.debug(f"Waiting...: {resp['status']}")
-                    time.sleep(5)
-            return GenieResponse(
-                f"Genie query timed out after {MAX_ITERATIONS} iterations of 5 seconds",
-                conversation_id=conversation_id,
-            )
+                return GenieResponse(
+                    f"Genie query timed out after {MAX_ITERATIONS} iterations of {ITERATION_FREQUENCY} seconds",
+                    conversation_id=conversation_id,
+                )
```
jennsun, line 343: can we consider putting a more human-friendly timeout message here - for example can we add a computed elapsed seconds here so response is like "Genie query timed out after {MAX_ITERATIONS} iterations of {ITERATION_FREQUENCY} seconds (total 5.0 seconds)"

src/databricks_ai_bridge/genie.py
```diff
@@ -213,45 +214,136 @@ def poll_query_results(
...
+                    elif current_status == "FAILED":
+                        # close any open child timeline span before ending
+                        if current_span is not None:
+                            client.end_span(
+                                trace_id=parent_trace_id,
+                                span_id=current_span.span_id,
+                                attributes={
+                                    "final_state": current_status,
+                                    "error": resp.get("error", "Unknown error"),
+                                },
```
jennsun: The span closing logic is used 5x - can we clean this up by extracting into a helper function to help with reusability/maintainability in the future? So we can then use it along the lines of
`_end_current_span(client, parent_trace_id, current_span, final_state, error)`

src/databricks_ai_bridge/genie.py
```diff
@@ -213,45 +214,136 @@ def poll_query_results(
...
+                            current_span = None
+
+                        # START new span for non-terminal states
+                        if current_status not in {
+                            "COMPLETED",
+                            "FAILED",
+                            "CANCELLED",
+                            "QUERY_RESULT_EXPIRED",
+                        }:
+                            current_span = client.start_span(
```
jennsun: can we define the terminal statuses at the module-level where we defined our other constants so 
`TERMINAL_STATES = {"COMPLETED", "FAILED", "CANCELLED", "QUERY_RESULT_EXPIRED"}` and it's clear what these statuses are for, then we can simply do `if current_status not in TERMINAL_STATES` for better readability :)

src/databricks_ai_bridge/genie.py
```diff
@@ -10,7 +10,8 @@
 from databricks.sdk import WorkspaceClient
 
 MAX_TOKENS_OF_DATA = 20000
-MAX_ITERATIONS = 50
+MAX_ITERATIONS = 500
+ITERATION_FREQUENCY = 0.1  # seconds
```
jennsun, line 15: Reminder to double check with Genie team/sid @smurching regarding max iteration increase/0.1 polling frequency which updates from 5 seconds between polls (`time.sleep(5)`) to 0.1 seconds between polls (using value of `ITERATION_FREQUENCY`). This changes the maximum wait time from 50 * 5 = 250 seconds to 500 * 0.1 = 50 seconds. This means our load impact goes from 1 call per 5 seconds (0.2rps) to 1 call per 0.1 seconds (10rps), resulting in 50x higher request rate of APIs per conversation. 

We should let genie know about this increased RPS and see if they have any concerns with load/rate limiting here?

src/databricks_ai_bridge/genie.py
```diff
@@ -213,45 +214,136 @@ def poll_query_results(
...
-                    if resp["status"] == "COMPLETED":
-                        text_content = next(r for r in resp["attachments"] if "text" in r)["text"][
-                            "content"
-                        ]
+
+            # use MLflow client to get parent of any new spans we create from the current active span
+            # (parenting keeps spans in the same trace)
+            client = mlflow.tracking.MlflowClient()
+            with mlflow.start_span(name="genie_timeline", span_type="CHAIN") as parent:
+                parent_trace_id = parent.trace_id if parent else None
```
jennsun, line 251: great - thanks for introducing the `genie_timeline` span here!

src/databricks_ai_bridge/genie.py
```diff
@@ -213,45 +214,136 @@ def poll_query_results(
...
+
+                        # START new span for non-terminal states
+                        if current_status not in {
+                            "COMPLETED",
+                            "FAILED",
+                            "CANCELLED",
+                            "QUERY_RESULT_EXPIRED",
+                        }:
+                            current_span = client.start_span(
+                                name=current_status.lower(),
```
jennsun: can we add try-catch logic/exception handling for Mlflow Tracing exceptions when we start/end spans?
According to the MLflow exception [doc](https://mlflow.org/docs/latest/api_reference/python_api/exceptions/mlflow.exceptions.html), "Tracing logic should not block the main execution flow in general" 

if we do not capture exception handling here, if the user asks a genie query and polling succeeds but the mlflow span fails, then there will be no genie response. We should add exception handling so even if the trace throws an error the user still gets their answer from Genie

Let's do this for areas where we start and end spans and catch the [MlflowTracingException](https://mlflow.org/docs/latest/api_reference/python_api/exceptions/mlflow.exceptions.html#mlflow.exceptions.MlflowTracingException), wrapping span operating in a try-except like something like this:
```
 try:
    current_span = client.start_span(
        name=current_status.lower(),
        trace_id=parent_trace_id,
        parent_id=parent_span_id,
        span_type="CHAIN",
        attributes={
            "state": current_status,
            "conversation_id": conversation_id,
            "message_id": message_id,
        },
    )
except mlflow.exceptions.MlflowTracingException as e:
    logging.warning(f"Failed to create span for {current_status}: {e}")
    current_span = None
logging logic continues here...
```

tests/databricks_ai_bridge/test_genie.py
```diff
@@ -442,3 +454,85 @@ def test_parse_query_result_with_timestamp_formats():
...
+        ]
+
+        genie.poll_for_result("123", "456")
+
+        # Verify span was closed when transitioning to COMPLETED
+        mock_client.end_span.assert_called_once()
+        end_kwargs = mock_client.end_span.call_args[1]
+        assert end_kwargs["trace_id"] == "trace_123"
+        assert end_kwargs["span_id"] == "child_789"
+        assert end_kwargs["attributes"]["final_state"] == "EXECUTING_QUERY"
```
jennsun, line 542: can we add unit tests for more coversage on these behaviors: 
1. verify no new span is created if the same status is returned consecutively (duplicate a set of statuses in the `do.side_effect` but make sure the start_span call count is still 1
2. tests for failed terminal states - cancelled, failed, query_result_expired
3. timeout case -> make sure final state attribute includes "TIMEOUT"
4. exception handling (after adding the try-except from the above comment) -> make sure polling completes successfully even if there are mlflow tracing failures

src/databricks_ai_bridge/genie.py
maxmarcussen-db: resolved with Genie team that we'd do 1 call per 0.5 seconds

src/databricks_ai_bridge/genie.py
```diff
@@ -213,45 +243,123 @@ def poll_query_results(
...
+
+                    if current_status == "COMPLETED":
+                        # close any open child timeline span before ending
+                        current_span = _end_current_span(
+                            client,
+                            parent_trace_id,
+                            current_span,
+                            current_status,
+                        )
+
```
jennsun: this is also a bit misleading for readability because it may sway future readers to believe there is a "COMPLETED" span to close - can you refer to comment above :)

src/databricks_ai_bridge/genie.py
```diff
@@ -213,45 +243,123 @@ def poll_query_results(
...
+                                    name=current_status.lower(),
+                                    trace_id=parent_trace_id,
+                                    parent_id=parent_span_id,
+                                    span_type="CHAIN",
+                                    attributes={
+                                        "state": current_status,
+                                        "conversation_id": conversation_id,
+                                        "message_id": message_id,
+                                    },
+                                )
```
jennsun, line 291: When refactoring to use `_end_current_span`, the span closing got moved inside the `if current_status not in TERMINAL_STATES:` check. This would create incorrect behavior for transitioning to a terminal state. 

Consider the case where we have a status change to a terminal state - for example `EXECUTING -> COMPLETED`. Due to our condition `if current_status not in TERMINAL_STATES:` , wouldn't the previous span not get ended properly/a span would never be started for the TERMINAL state? 

In this case, 
```
last_status = EXECUTING
current_status = COMPLETED

we would skip the # START new span for non-terminal states block
then when we trickle down to the if current_status == "COMPLETED": block
it would call `_end_current_span` where we pass in the `current_status` but not `last_status` - so the "COMPLETED" span would get closed but not the "EXECUTING" span (and the "COMPLETED" span was never created in the first place)


Therefore:
- The status change block skips closing the `executing_query` span (terminal state check fails)
- `executing_query` span remains open
- Terminal handler tries to close it with `final_state="COMPLETED"` (incorrect)
```

We should make sure we handle this terminal transition properly by either:
1. Passing in `last_status` instead of `current_status` for the status we pass into `_end_current_span` under each terminal state conditional blocks (if current_status == "COMPLETED":, elif current_status in {"CANCELLED", "QUERY_RESULT_EXPIRED"}, elif current_status == "FAILED":)
2.  MOVE the end span logic outside of the `if current_status not in TERMINAL_STATES: block `! 

```
if current_status != last_status:
    # END previous span on ALL transitions
    current_span = _end_current_span(
        client, parent_trace_id, current_span, last_status
    )
    
    # START new span only for non-terminal states
    if current_status not in TERMINAL_STATES:
        try:
            current_span = client.start_span(...)
```

Can you double check this with testing?

smurching (approved):
LGTM

PR #197, opened 2025-10-06
Use MMR, hybrid search and self-managed embeddings together in the langchain DatabricksVectorSearch object

sllynn:
This is to address the case where a user is using the DatabricksVectorSearch retriever in langchain using the hybrid-query-with-MMR pattern on an index with self-managed embeddings. It updates the MMR-related methods with the same logic used in the standard similarity search equivalents.


aravind-segu (approved):
Thank you for making this contribution!

PR #199, opened 2025-10-07
add a check for non-delta non-choices usage chunk

jakesteelman-insurica:


WIll-Xu35:
Facing the same issue here

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -688,6 +688,18 @@ def _stream(
                             logprobs=generation_info.get("logprobs"),
                         )
                     yield generation_chunk
+                elif chunk.usage and stream_usage:
```
sunishsheth2009, line 691: Can we add some test here: `integrations/langchain/tests/unit_tests/test_chat_models.py` for this where the last chunk is usage data and those are added to the response correctly?

PR #212, opened 2025-11-07
Add Lakebase Pool Memory/Connection Pooling: LakebasePool and langchain CheckpointSaver

jennsun:
Stateful Agents example: https://docs.databricks.com/aws/en/generative-ai/agent-framework/stateful-agents - currently does not wrap any connection pooling logic in our SDKs. This PR adds connection pooling + lakebase with token caching logic into databricks-ai-bridge and databricks-langchain

in predict_stream, pass in checkpointer by doing:
```
from databricks_langchain import CheckpointSaver
```
then in predict_stream do:
```
with CheckpointSaver(instance_name="instance_name") as checkpointer:
            graph = self._create_graph(checkpointer)
```

Test notebook: https://eng-ml-inference.staging.cloud.databricks.com/editor/notebooks/1475155873800403?o=1653573648247579 

Example successful agent endpoint: https://eng-ml-inference.staging.cloud.databricks.com/ml/endpoints/agents_main-default-j-stateful-sdk-memory?o=1653573648247579

Tested in playground/review app by chatting with agent, then in a new conversation pass in conversation id as a custom input to see if it recalls past thread context:

image

image

Able to see threads and checkpoints stored in lakebase:
image

Unit tests/covered in github workflow automated tests:
```
pytest tests/databricks_ai_bridge/test_lakebase.py
```
```
pytest integrations/langchain/tests/unit_tests/test_checkpointing.py
```

### DIFF
--- a/src/databricks_ai_bridge/lakebase.py
+++ b/src/databricks_ai_bridge/lakebase.py
@@ -62,7 +62,7 @@ class LakebasePool:
             instance = workspace_client.database.get_database_instance(instance_name)
         except Exception as exc:
             raise ValueError(
-                f"Unable to resolve Lakebase host for instance '{instance_name}'. "
+                f"Unable to resolve Lakebase instance '{instance_name}'. "
                 "Ensure the instance name is correct."
             ) from exc
 
@@ -73,7 +73,7 @@ class LakebasePool:
         if not resolved_host:
             raise ValueError(
                 f"Lakebase host not found for instance '{instance_name}'. "
-                "Ensure the instance exposes `read_write_dns` or `read_only_dns` in workspace metadata."
+                "Ensure the instance is running and in AVAILABLE state."
             )
 
         self.workspace_client = workspace_client


src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+class LakebasePool:
+    """Wrapper around a psycopg connection pool with rotating credentials."""
+
+    def __init__(
+        self,
+        *,
+        workspace_client: WorkspaceClient,
+        instance_name: str,
```
smurching, line 52: Is it possible to require only instance name? And have defaults for everything else e.g. use the default workspace client if unspecified?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,259 @@
...
+        host: str,
+        database: str,
+        username: Optional[str] = None,
+        port: int = DEFAULT_PORT,
```
smurching: Probably better to accept **kwargs for everything not lakebase-specific and enable passing these kwargs through to ConnectionPool

src/databricks_ai_bridge/lakebase.py
smurching: Also just to confirm usually this won't the API users use right? It'll be used internally in this package by databricks-langchain and databricks-openai etc, or maybe by users of other agent authoring frameworks where we don't have an integration package

smurching: Thinking if we should keep the API internal and wait for use cases to emerge but I'm ok with it as a public one for this use case.

> maybe by users of other agent authoring frameworks where we don't have an integration package

I think this sort of logic ideally can eventually go into e.g. the apps SDK, since there's nothing AI-specific about it

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,259 @@
...
+        self._finalizer.detach()
+
+    def __enter__(self) -> PooledPostgresSaver:
+        return self
+
+    def __exit__(self, exc_type, exc, tb) -> None:  # type: ignore[override]
+        self.close()
+
+
+def make_checkpointer(pool: Union[ConnectionPool, LakebasePool]) -> PooledPostgresSaver:
```
smurching: This API should be in databricks-langchain code right? Any reason not to provide a `from databricks_langchain import CheckpointSaver` class like in https://docs.google.com/document/d/1gX1XEUdCUvumv-Buks-OcL69DqH3I0P34SVPwgc34Cg/edit?tab=t.0#heading=h.aiy7hr9z0ond?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,259 @@
...
+
+        cred = cls.workspace_client.database.generate_database_credential(
+            request_id=str(uuid.uuid4()),
+            instance_names=[cls.instance_name],
+        )
+        return cred.token
+
+    @classmethod
+    def _get_token(cls) -> str:
+        now = time.time()
```
bbqiu: super nit: can we move this into the context manager below to avoid any edge cases

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,259 @@
...
+    for *N* minutes to avoid repeated API calls.
+    """
+
+    workspace_client: Optional[WorkspaceClient] = None
+    instance_name: Optional[str] = None
+
+    _cache_lock = Lock()
+    _cached_token: Optional[str] = None
+    _cache_ts: Optional[float] = None
+    _cache_duration_sec: int = DEFAULT_CACHE_MINUTES * 60
```
bbqiu: is there an existing design pattern for connections having connection-specific attributes as class attributes?

not sure when this would happen, but if we want to access two separate lakebase instances, we'd run into issues here with conflicting `instance_name` at the minimum (or workspace client, if one is an SP client and the other is a user client)

would it be possible to make these instance variables instead?

(from claude) example usage
```
  class ConnectionFactory:
      def __init__(self, workspace_client, instance_name, cache_duration_sec):
          self.workspace_client = workspace_client
          self.instance_name = instance_name
          # ... token caching logic here

      def create_connection(self, conninfo="", **kwargs):
          kwargs["password"] = self._get_token()
          return psycopg.Connection.connect(conninfo=conninfo, **kwargs)

  Then in LakebasePool:
  self.connection_factory = ConnectionFactory(workspace_client, instance_name, ...)
  self._pool = ConnectionPool(
      connection_factory=self.connection_factory.create_connection,
      # ... other params
  )
```

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,259 @@
...
+
+    @classmethod
+    def _mint_token(cls) -> str:
+        if not cls.workspace_client or not cls.instance_name:
+            raise RuntimeError("RotatingCredentialConnection not initialized.")
+
+        cred = cls.workspace_client.database.generate_database_credential(
+            request_id=str(uuid.uuid4()),
+            instance_names=[cls.instance_name],
+        )
```
bbqiu: how are failures from generating this OAuth token surfaced to the user? is the error message helpful -- something like "<user> does not have access to lakebase instance with name <instance name>"

src/databricks_ai_bridge/lakebase.py
bbqiu: ya agreed that having this as a public API is totally fine even if we introduce a better solution via the Apps SDK in the future

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,259 @@
...
+            conninfo=conninfo,
+            connection_class=RotatingCredentialConnection,
+            min_size=min_size,
+            max_size=max_size,
+            timeout=timeout,
+            open=open_pool,
+            kwargs=default_kwargs,
+        )
+
+        if probe:
```
bbqiu: i'm not the most familiar with the levels of permissions you can be granted for lakebase, but is it possible for you to have read without having write access? is there a more complex probe that we might want to try here?

we're currently running into issues with reading/writing from existing tables if you're not a superuser right?

bbqiu: talking w/ @aravind-segu, there is a "can use" and "can manage" roles on lakebase (where the latter is the superuser that can write/read existing tables), so it might be worthwhile to make this a more nuanced check (since we're labeling this as a generic lakebase connector)

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,259 @@
...
+    def __init__(self, pool: ConnectionPool):
+        self._pool = pool
+        self._conn = pool.getconn()
+        super().__init__(self._conn)
+        self._finalizer = weakref.finalize(self, self._release)
+
+    def _release(self) -> None:
+        if self._conn is not None:
+            try:
+                self._pool.putconn(self._conn)
```
bbqiu: if this fails, can we log an error?

from claude

```python
  def _release(self) -> None:
      if self._conn is not None:
          try:
              self._pool.putconn(self._conn)
          except (psycopg.PoolClosed, psycopg.PoolTimeout) as e:
              # Pool is closed/full - expected in some scenarios
              logger.debug("Pool unavailable for connection return: %s", e)
          except Exception as e:
              # Unexpected error - more serious
              logger.error("Unexpected error returning connection to pool: %s", e)
          finally:
              self._conn = None
```

src/databricks_ai_bridge/lakebase.py
jennsun: discussed offline - there are no clear postgres roles other than `databricks_superuser` (which allows read/write to ALL tables) that will map to queries we'd want to use for the prober. Getting read/write access to lakebase instances requires manually granting them on certain schemas/tables:
https://docs.databricks.com/aws/en/oltp/instances/pg-roles?language=PostgreSQL#create-postgres-roles-and-grant-privileges-for-databricks-identities

The prober check would also not be the most helpful for example, when users are locally testing using their own credentials but eventually deploy their agent which uses a service principal for reading/writing to lakebase instance. We currently do permission checks in other ways (for example, in granting permissions flow in apps/model-serving) so that would cover the majority of customer CUJs

I will remove this probe for now, but include this consideration as part of our conversation with the Lakebase team to improve this CUJ

integrations/langchain/src/databricks_langchain/__init__.py
```diff
@@ -18,6 +18,7 @@
 from unitycatalog.ai.langchain.toolkit import UCFunctionToolkit, UnityCatalogTool
 
 from databricks_langchain.chat_models import ChatDatabricks
+from databricks_langchain.checkpointing import CheckpointSaver
```
smurching:
```suggestion
from databricks_langchain.checkpoint import CheckpointSaver
```

Minor but just to match Langgraph module structure (we'll need to rename `checkpointing.py` to achieve this) looking at https://reference.langchain.com/python/langgraph/checkpoints/#langgraph.checkpoint.base.BaseCheckpointSaver

smurching: It doesn't matter too much since customers should just do `from databricks_langchain import CheckpointSaver`, but might as well be consistent

integrations/langchain/src/databricks_langchain/checkpointing.py
```diff
@@ -0,0 +1,112 @@
...
+                return super().__exit__(exc_type, exc, tb)
+            finally:
+                self._close_pool = previous
+
+    _CHECKPOINT_SAVER_CLASS = _CheckpointSaverImpl
+    globals()["CheckpointSaver"] = _CheckpointSaverImpl
+    return _CHECKPOINT_SAVER_CLASS
+
+
+class CheckpointSaver:  # type: ignore[override]
```
smurching: Why not directly subclass https://reference.langchain.com/python/langgraph/checkpoints/#langgraph.checkpoint.base.BaseCheckpointSaver? We can try-catch import errors of classes from `databricks_ai_bridge` in our `__init__` method to provide a better error if desired, but I think we can probably reduce some of the complexity here (?)

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,352 @@
...
+            if reserved in pool_kwargs:
+                raise TypeError(f"Argument '{reserved}' cannot be overridden.")
+        min_size = int(pool_kwargs.pop("min_size", DEFAULT_MIN_SIZE))
+        max_size = int(pool_kwargs.pop("max_size", DEFAULT_MAX_SIZE))
+        timeout = float(pool_kwargs.pop("timeout", DEFAULT_TIMEOUT))
+        open_flag = pool_kwargs.pop("open", True)
+
+        self.workspace_client = workspace_client
+        self.instance_name = instance_name
+        self.host = host
```
smurching: Is this ever used (outside of logger statements)?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,352 @@
...
+    def __init__(
+        self,
+        *,
+        instance_name: str,
+        workspace_client: WorkspaceClient | None = None,
+        host: str | None = None,
+        database: str | None = None,
+        username: Optional[str] = None,
+        port: Optional[int] = None,
+        sslmode: Optional[str] = None,
```
smurching: Most of this stuff should probably not be passed here right? Instance name is all we need (?)

src/databricks_ai_bridge/lakebase.py
smurching: Oh I see it's used in `conninfo`, but I think the host can be derived from the lakebase instance name, no need to pass in an override

src/databricks_ai_bridge/lakebase.py
smurching: At least host, port, username, and sslmode seem like they should be attributes of the lakebase instance. Optionally specifying the `database` within the lakebase instance to use for checkpointing seems reasonable

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,352 @@
...
+        token_cache_minutes: Optional[int] = None,
```
smurching:
```suggestion
        token_cache_secs: Optional[int] = None,
```

Seems ok to make this seconds to provide more control, any reason to go with minutes?

smurching: Can we add a docstring for this class too to explain what these params do?

src/databricks_ai_bridge/lakebase.py
jennsun: The host is separate from the lakebase name, example instance name is just anything you give it like "lakebase-name" but host is built from the instance uuid, something like `instance-cd864d21-d19f-4180-bddd-eb1bf4da7625.database.staging.cloud.databricks.com`. Are you thinking of using another method to derive the host from the lakebase instance name here?

jennsun: just confirmed the host is used in the connection string, but the lakebase instance name is not. The connection string is composed like:
`jdbc:postgresql://jenny.sun%40databricks.com:${PGPASSWORD}@instance-cd864d21-d19f-4180-bddd-eb1bf4da7625.database.staging.cloud.databricks.com:5432/databricks_postgres?sslmode=require`

We will also need the instance name to generate a database credential. 

In the current CUJ, the lakebase name is passed in the config defined by the user for the purposes of adding the lakebase resource to specify for automatic auth passthrough at deployment time. So we will need both instance name + host

jennsun: discussed offline - we can retrieve the database host name directly from the instance name using https://databricks-sdk-py.readthedocs.io/en/latest/dbdataclasses/database.html#databricks.sdk.service.database.DatabaseInstance.read_write_dns

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -0,0 +1,70 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Dict
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    from databricks.sdk import WorkspaceClient
+else:  # pragma: no cover - runtime fallback when dependency absent
+    WorkspaceClient = Any  # type: ignore
```
bbqiu: nit: do we need to do this safely? the databricks sdk is one of our core dependencies

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def _load_lakebase_dependencies() -> tuple[type, type]:
+    """Import Lakebase dependencies lazily."""
+
+    try:
+        from databricks_ai_bridge.lakebase import LakebasePool, PooledPostgresSaver
+    except ModuleNotFoundError as exc:  # pragma: no cover - surfaced to caller
+        raise ModuleNotFoundError(
+            "CheckpointSaver requires databricks-ai-bridge[lakebase] (psycopg) to be installed."
```
bbqiu: nit: can we make this statement more clear? can also just remove (psycopg) from this statement

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -0,0 +1,70 @@
...
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    from databricks.sdk import WorkspaceClient
+else:  # pragma: no cover - runtime fallback when dependency absent
+    WorkspaceClient = Any  # type: ignore
+
+_CHECKPOINT_SAVER_IMPL: type | None = None
+
+
+def _load_lakebase_dependencies() -> tuple[type, type]:
+    """Import Lakebase dependencies lazily."""
```
bbqiu: nit: since we're only using this for the checkpoint saver, can we rename this to reflect that

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -0,0 +1,70 @@
...
+            try:
+                return super().__exit__(exc_type, exc, tb)
+            finally:
+                self._close_pool = previous
+
+    _CHECKPOINT_SAVER_IMPL = _CheckpointSaver
+    return _CHECKPOINT_SAVER_IMPL
+
+
+class CheckpointSaver:  # type: ignore[override]
```
bbqiu: was there an offline discussion about why we aren't subclassing the checkpointsaver class from langchain?

https://github.com/databricks/databricks-ai-bridge/pull/212#discussion_r2516584466

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+        if workspace_client is None:
+            workspace_client = WorkspaceClient()
+
+        resolved_instance = instance_name or LAKEBASE_NAME
+        if resolved_instance is None:
+            raise ValueError(
+                "Lakebase instance name must be provided. Specify the instance_name argument or set the LAKEBASE_NAME environment variable."
+            )
+
+        resolved_host = DEFAULT_HOST
```
bbqiu: do we need default host if it's always none? do we use this value elsewhere

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+                    "Lakebase host must be provided. Unable to resolve host from workspace metadata."
+                ) from exc
+
+            resolved_host = getattr(instance, "read_write_dns", None) or getattr(
+                instance, "read_only_dns", None
+            )
+
+        if resolved_host is None:
+            raise ValueError(
+                "Lakebase host must be provided. Make sure your Lakebase instance name is correct, set DB_HOST, or ensure the workspace instance metadata exposes read_write_dns."
```
bbqiu: can we revise this error msg since the user isn't providing the lakebase host? in what cases would the host name not get resolved (when is the dns not set on the instance we get from the workspace client)

also how would i "set DB_HOST" (unless it's part of the original comment that just hasn't been updated)?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+        sslmode = DEFAULT_SSLMODE
+        cache_seconds = DEFAULT_CACHE_SECONDS
+
+        self.workspace_client = workspace_client
+        self.instance_name = resolved_instance
+        self.host = resolved_host
+        self.database = database
+        self.port = port
+        self.sslmode = sslmode
+        self.token_cache_seconds = cache_seconds
```
bbqiu: do we need `database` and `DEFAULT_DATABASE`?

ditto for all the other default vars / temp variables we're using here

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+            "keepalives_idle": 30,
+            "keepalives_interval": 10,
+            "keepalives_count": 5,
+        }
+
+        connection_class = _make_rotating_connection_class(
+            workspace_client=workspace_client,
+            instance_name=resolved_instance,
+            cache_duration_sec=cache_seconds,
+        )
```
bbqiu: out of curiosity, why are we not setting this directly to self._connection_class?

src/databricks_ai_bridge/lakebase.py
bbqiu: should we delete this now that it's been moved to the langchain class?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+DEFAULT_TIMEOUT = 30.0
+DEFAULT_SSLMODE = "require"
+DEFAULT_PORT = 5432
+DEFAULT_HOST = None
+LAKEBASE_NAME = None
+DEFAULT_DATABASE = "databricks_postgres"
+DEFAULT_USERNAME = None
+
+
+class RotatingCredentialConnection(psycopg.Connection):
```
bbqiu: can we make it clear that we shouldn't ever really use this directly by making this a private class and not exporting it in the `__all__` list above? we should instead expose the factory function `_make_rotating_connection_class` (should change to a public function and add a docstring explaining usage etc.) as the proper way to create a class to manage creds.

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+
+class LakebasePool:
+    """Wrapper around a psycopg connection pool with rotating Lakehouse credentials.
+
+    instance_name:
+        User-set name on Lakebase Instance
+        (can retrieve from connection details page in Databricks workspace)
+    workspace_client:
+        Optional `WorkspaceClient` to use; default client is created otherwise.
+    token_cache_seconds:
```
bbqiu: can we make this name consistent across our classes here? `token_cache_duration_seconds`

since we're naming it in the docstring, can we also make it an explicit keyword arg in the constructor for discoverability?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+            pool_params.get("min_size"),
+            pool_params.get("max_size"),
+            cache_seconds,
+        )
+
+    @property
+    def pool(self) -> ConnectionPool:
+        return self._pool
+
+    def connection(self):
```
bbqiu, line 169: nit: type hint

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+            ):
+                return cls._cached_token
+
+            token = cls._mint_token()
+            cls._cached_token = token
+            cls._cache_ts = now
+            return token
+
+    @classmethod
+    def connect(cls, conninfo: str = "", **kwargs):
```
bbqiu: nit: type hint

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+
+    def __exit__(self, exc_type, exc, tb) -> None:  # type: ignore[override]
+        self.close()
+
+
+# Build lakebase pool instance with only instance name required
+def build_lakebase_pool(
+    *,
+    instance_name: str,
+    token_cache_seconds: Optional[int] = None,
```
bbqiu: remember to pass this into the LakebasePool constructor

integrations/langchain/tests/unit_tests/test_checkpoint.py
```diff
@@ -0,0 +1,334 @@
+from __future__ import annotations
+
+import sys
+import types
+from pathlib import Path
+from unittest.mock import MagicMock
+
+
+# create lightweight stand-ins for modules for testing
+def _ensure_optional_modules() -> None:
```
bbqiu: instead of doing all of these, we can use pytest.skip if the package isn't in the CI env. then we can modify `.github/workflows/main.yml` to install the lakebase dependencies to run these tests specifically with the real classes. that way we can write unit tests that require us to be specific about which functions we are mocking etc.

(i think some stuff like the databricks sdk / vector search / langchain we don't need to mock at all since theyre required core dependencies)

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+        (3000 seconds).
+    **pool_kwargs:
+        Additional options passed to ``psycopg_pool.ConnectionPool`` (e.g.
+        ``min_size``, ``max_size``).
+    """
+
+    def __init__(
+        self,
+        *,
+        instance_name: str | None = None,
```
bbqiu: from the comment thread with sid earlier, this should be the only required param right?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+        workspace_client: WorkspaceClient | None = None,
+        **pool_kwargs: object,
+    ) -> None:
+        if workspace_client is None:
+            workspace_client = WorkspaceClient()
+
+        resolved_instance = instance_name or LAKEBASE_NAME
+        if resolved_instance is None:
+            raise ValueError(
+                "Lakebase instance name must be provided. Specify the instance_name argument or set the LAKEBASE_NAME environment variable."
```
bbqiu: nit: are we reading this from the env var? this may be an error message that is old like the one below

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -0,0 +1,70 @@
...
+            super().__init__(self._lakebase.pool)
+
+        def close(self) -> None:  # type: ignore[override]
+            super().close()
+            if self._close_pool:
+                self._lakebase.close()
+
+        def __exit__(self, exc_type, exc, tb) -> None:  # type: ignore[override]
+            previous = self._close_pool
+            self._close_pool = False
```
bbqiu: out of curiosity, why do we have to set this to False first? would you mind leaving a comment to explain

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -0,0 +1,70 @@
...
+            )
+            self._close_pool = True
+            super().__init__(self._lakebase.pool)
+
+        def close(self) -> None:  # type: ignore[override]
+            super().close()
+            if self._close_pool:
+                self._lakebase.close()
+
+        def __exit__(self, exc_type, exc, tb) -> None:  # type: ignore[override]
```
bbqiu: nit: do we need this type override?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,327 @@
...
+    def _mint_token(cls) -> str:
+        if not cls.workspace_client or not cls.instance_name:
+            raise RuntimeError("RotatingCredentialConnection not initialized.")
+
+        try:
+            cred = cls.workspace_client.database.generate_database_credential(
+                request_id=str(uuid.uuid4()),
+                instance_names=[cls.instance_name],
+            )
+        except Exception as exc:  # pragma: no cover - surfaced to user
```
bbqiu: are these pragma statements necessary? were we failing lint without them?

integrations/langchain/src/databricks_langchain/checkpoint.py
bbqiu: rereading, i think we need to add this extra to the pyproject.toml file and add langgraph-checkpoint-postgres as a dependency

src/databricks_ai_bridge/lakebase.py
jennsun: actually going to remove the token cache seconds as an adjustable knob - better to have less arguments at first and add if need be, but this is usually abstracted away from user

src/databricks_ai_bridge/lakebase.py
jennsun: I'm hardcoding these for now instead of making them values that users can pass in because this api is specific to facilitating databricks lakebase connections for now:
```
DEFAULT_CACHE_SECONDS = 50 * 60  # Cache token for 50 minutes
DEFAULT_MIN_SIZE = 1
DEFAULT_MAX_SIZE = 10
DEFAULT_TIMEOUT = 30.0
DEFAULT_SSLMODE = "require"
DEFAULT_PORT = 5432
DEFAULT_DATABASE = "databricks_postgres"
```

pyproject.toml
```diff
@@ -37,22 +36,12 @@ packages = ["src/databricks_ai_bridge"]
...
-  "autodoc_pydantic>=2.2.0",
+  "ruff==0.6.4",
+  "databricks-vectorsearch>=0.50",
 ]
-lint = [
-  "ruff==0.12.10",
+
+lakebase = [
+  "psycopg[binary,pool]>=3.1",
 ]
```
bbqiu: just wanted to confirm:
- this not installed by default with the env already?
- this package is heavy enough to warrant it's own extra instead of just adding it as a core dependency?

integrations/langchain/tests/unit_tests/test_checkpoint.py
```diff
@@ -0,0 +1,99 @@
...
+from unittest.mock import MagicMock
+
+import pytest
+from databricks_ai_bridge import lakebase
+
+from databricks_langchain import CheckpointSaver
+
+pytest.importorskip("psycopg")
+pytest.importorskip("psycopg_pool")
+pytest.importorskip("langgraph.checkpoint.postgres")
```
bbqiu, line 12: just as a sanity check, without the lakebase extra, this test is getting run in the normal langchain test without being skipped

Image

integrations/langchain/pyproject.toml
```diff
@@ -17,6 +17,7 @@ dependencies = [
     "unitycatalog-langchain[databricks]>=0.3.0",
     "databricks-sdk>=0.65.0",
     "openai>=1.99.9",
+    "langgraph-checkpoint-postgres==2.0.21",
```
bbqiu: can we pin a min instead of a version

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,256 @@
...
+DEFAULT_SSLMODE = "require"
+DEFAULT_PORT = 5432
+DEFAULT_DATABASE = "databricks_postgres"
+
+
+class _RotatingCredentialConnection(psycopg.Connection):
+    """
+    Base psycopg connection that injects a Lakebase (Postgres) OAuth token at
+    connect-time. Concrete subclasses are generated per pool so that token
+    caches don't leak across pools.
```
bbqiu: can we update the docstring to say that this is not meant to be used? and we should use `create_connection_class` instead to avoid leaking tokens

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,256 @@
...
+            ):
+                return cls._cached_token
+
+            token = cls._mint_token()
+            cls._cached_token = token
+            cls._cache_ts = now
+            return token
+
+    @classmethod
+    def connect(cls: Type["_RotatingCredentialConnection"], conninfo: str = "", **kwargs):
```
bbqiu: nit: type hint

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,256 @@
...
+        if not resolved_host:
+            raise ValueError(
+                f"Lakebase host not found for instance '{instance_name}'. "
+                "Ensure the instance exposes `read_write_dns` or `read_only_dns` in workspace metadata."
+            )
+
+        database = DEFAULT_DATABASE
+        port = DEFAULT_PORT
+        sslmode = DEFAULT_SSLMODE
+        token_cache_duration_seconds = DEFAULT_CACHE_SECONDS
```
bbqiu: nit: can we skip the intermediate assignment since there are no modifications before assigning them to the class attributes?

integrations/langchain/tests/unit_tests/test_checkpoint.py
jennsun: I added the lakebase dependency to langchain_test so currently the expected behavior is test_checkpoint.py should be included in the normal langchain tests

we have tests running in 2 jobs: the new lakebase-test only runs the lakebase-specific tests (test_lakebase.py) and the langchain tests includes `test_checkpoint.py` because it is under the langchain package

Is this behavior different than you expect?

.github/workflows/main.yml
```diff
@@ -50,6 +50,28 @@ jobs:
         run: |
           pytest tests/
 
+  lakebase_test:
```
bbqiu: we can delete this now right?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -16,7 +16,7 @@
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_CACHE_SECONDS = 50 * 60  # Cache token for 50 minutes
+DEFAULT_TOKEN_CACHE_DURATION_SECONDS = 50 * 60  # Cache token for 50 minutes
```
bbqiu, line 18: nit: can we leave a documentation link from lakebase for why we're setting these things as default?

bbqiu (approved):
lgtm, please manually test everything to ensure things continue working

also can get sid's signoff on the final APIs

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -0,0 +1,94 @@
...
+            exit_method = getattr(self._inner, "__exit__", None)
+            if exit_method:
+                exit_method(exc_type, exc, tb)
+        finally:
+            try:
+                self.close()
+            finally:
+                self._close_pool = getattr(self, "_prev_close_pool", True)
+
+    get = _delegate("get")
```
smurching: What is all of this stuff for?

jennsun: BaseCheckpointSaver declares those methods as abstract (https://reference.langchain.com/javascript/classes/_langchain_langgraph-checkpoint.BaseCheckpointSaver.html) and if we don't assign these here, we run into issues like not implemented error when we try to construct CheckpointSaver - so I forwarded all the listed APIs under [BaseCheckpointSaver](https://reference.langchain.com/python/langgraph/checkpoints/#langgraph.checkpoint.base.Checkpoint.updated_channels) 
is there another approach you would recommend here?

jennsun: discussed offline - we should subclass PostgresSaver instead of BaseCheckpointSaver to prevent delegation/forwarding pattern like this. updated and tested!

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,248 @@
...
+DEFAULT_PORT = 5432
+DEFAULT_DATABASE = "databricks_postgres"
+
+
+class _RotatingCredentialConnection(psycopg.Connection):
+    """
+    Note: Don't use - use create_connection_class instead to avoid leaking tokens
+    """
+
+    workspace_client: Optional[WorkspaceClient] = None
```
smurching: Could this be an instance attribute?

smurching:
```suggestion
    def __init__(...):
       super.__init__()
       self._cache_lock = Lock()
```

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,248 @@
...
+        `connection_class`.
+
+    Example
+    -------
+    >>> w = WorkspaceClient()
+    >>> ConnectionClass = create_connection_class(w, "my-lakebase")
+    >>> conn = ConnectionClass.connect("connection-string")
+    """
+
+    return type(
```
smurching: Can construct an instance here instead of creating a new type

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -0,0 +1,63 @@
...
+    ) -> None:
+        self._lakebase = LakebasePool(
+            instance_name=database_instance,
+            workspace_client=workspace_client,
+            **dict(pool_kwargs),
+        )
+        self._pool = self._lakebase.pool
+        self._conn = self._pool.getconn()
+        super().__init__(self._conn)
+        self._close_pool = True
```
smurching: Should be able to just have `__init__`

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -0,0 +1,27 @@
...
+    LangGraph PostgresSaver using a Lakebase connection pool.
+
+    instance_name: Name of Lakebase Instance
+    """
+
+    def __init__(
+        self,
+        *,
+        instance_name: str,
+        workspace_client: WorkspaceClient | None = None,
```
jennsun, line 19: using `instance_name` for name to be consistent with current Databricks Database SDK, example: https://databricks-sdk-py.readthedocs.io/en/latest/workspace/database/database.html#databricks.sdk.service.database.DatabaseAPI.list_database_catalogs

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -0,0 +1,36 @@
...
+        super().__init__(self._lakebase.pool)
+
+    def __enter__(self):
+        """Enter context manager."""
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Exit context manager and close the connection pool."""
+        self._lakebase.close()
+        return False
```
jennsun, line 36: need enter/exit here to add context manager support, otherwise when when setting up checkpoint tables we'll run into `TypeError: 'CheckpointSaver' object does not support the context manager protocol`

pyproject.toml
```diff
@@ -15,6 +15,7 @@ dependencies = [
   "tiktoken>=0.8.0",
   "tabulate>=0.9.0",
   "mlflow-skinny>=2.19.0",
+  "psycopg[binary, pool]>=3.2",
 ]
```
jennsun, line 18: we actually need both psycopg[binary, pool] , saw this was included in the [lakebase connection docs](https://docs.databricks.com/aws/en/oltp/instances/query/notebook#psycopg3) so that's where I initially got it from
additionally, I tried deploying an agent model serving endpoint without binary and the build fails with
```
[j7ppc] [2025-11-18 08:20:05 +0000] ImportError: no pq wrapper available.
[j7ppc] [2025-11-18 08:20:05 +0000] Attempts made:
[j7ppc] [2025-11-18 08:20:05 +0000] - couldn't import psycopg 'c' implementation: No module named 'psycopg_c'
[j7ppc] [2025-11-18 08:20:05 +0000] - couldn't import psycopg 'binary' implementation: No module named 'psycopg_binary'
[j7ppc] [2025-11-18 08:20:05 +0000] - couldn't import psycopg 'python' implementation: libpq library not found
```
the binary package bundles some libraries needed and installs a pre-built wheel file containing the compiled code that the model serving environment doesn't include

src/databricks_ai_bridge/lakebase.py
```diff
@@ -0,0 +1,175 @@
...
+            ) from exc
+
+        resolved_host = getattr(instance, "read_write_dns", None) or getattr(
+            instance, "read_only_dns", None
+        )
+
+        if not resolved_host:
+            raise ValueError(
+                f"Lakebase host not found for instance '{instance_name}'. "
+                "Ensure the instance exposes `read_write_dns` or `read_only_dns` in workspace metadata."
```
smurching: Double check if we need this / if this is actionable?

jennsun: `read_write_dns` is usually present, but during provisioning/stop/error the host may be unset or unusable ([instance cannot be used for read or write operations when stopped](https://docs.databricks.com/aws/en/oltp/instances/create/#behavior-when-stopped)) -- will update this to be more actionable to make sure instance is in "AVAILABLE" state, and also started a thread with lakebase team to confirm if we can refine this messaging further

PR #220, opened 2025-11-18
QoL improvements for langchain Genie implementation

maxmarcussen-db:
Added support for returning Pandas results. Also added support for passing in a Genie conversation ID so that a request can follow up within an existing Genie conversation, rather than creating a new Genie conversation every time.

`
agent = GenieAgent(
    genie_space_id=GENIE_SPACE_ID,
    genie_agent_name="GenieAgent",
    include_context=True,  
    client=client,
    truncate_results=False,
    return_pandas=False  
)

second_query = {
    "messages": [
        {"role": "user", "content": "Follow up on a past conversation?"}
    ],
    "conversation_id": conversation_id  #
}

result2 = agent.invoke(second_query)`

### DIFF
--- a/integrations/langchain/src/databricks_langchain/genie.py
+++ b/integrations/langchain/src/databricks_langchain/genie.py
@@ -101,7 +101,6 @@ def GenieAgent(
     include_context: bool = False,
     message_processor: Optional[Callable] = None,
     client: Optional["WorkspaceClient"] = None,
-    truncate_results: bool = False,
     return_pandas: bool = False,
 ):
     """Create a genie agent that can be used to query the API. If a description is not provided, the description of the genie space will be used.
@@ -115,7 +114,6 @@ def GenieAgent(
                             or LangChain Message objects and return a query string. If not provided, the agent will
                             use the chat history to form the query.
         client: Optional WorkspaceClient instance
-        truncate_results: Whether to truncate results to fit within token limits
         return_pandas: Whether to return results as pandas DataFrames (if False, returns markdown strings)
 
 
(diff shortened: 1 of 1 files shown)


integrations/langchain/src/databricks_langchain/genie.py
```diff
@@ -64,6 +101,8 @@ def GenieAgent(
     include_context: bool = False,
     message_processor: Optional[Callable] = None,
     client: Optional["WorkspaceClient"] = None,
+    truncate_results: bool = False,
```
smurching: Is there a reason to reintroduce this `truncate_results` argument? I believe @aravind-segu removed this since it's now deprecated functionality

smurching: (maybe it's just that this pull request branch is based on an old version of `main` + this isn't intentional?)

smurching: The reason we are trying to deprecate this flag btw is that it's generally better to truncate results in client-side code, not in this API

smurching (commented):
One comment - we should remove truncate_results - but mostly looks reasonable!

integrations/langchain/src/databricks_langchain/genie.py
maxmarcussen-db: removed it -- one note is that it's still exposed in the regular `ai-bridge` implementation [here](https://github.com/databricks/databricks-ai-bridge/blob/main/src/databricks_ai_bridge/genie.py)

PR #221, opened 2025-11-19
Sphinx Fixes for Docs

aravind-segu:
mock packages in sphinx so that we can compile docs

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -309,7 +309,13 @@ latex_documents = [
 
 # Mock torch & fastai imports as per suggestion in
 # https://github.com/sphinx-doc/sphinx/issues/6521#issuecomment-505765893
-autodoc_mock_imports = ["torch", "fastai"]
+autodoc_mock_imports = [
+    "torch",
+    "fastai",
+    "psycopg",  # Required by databricks_ai_bridge.lakebase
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -309,7 +309,13 @@
...
 # Mock torch & fastai imports as per suggestion in
 # https://github.com/sphinx-doc/sphinx/issues/6521#issuecomment-505765893
-autodoc_mock_imports = ["torch", "fastai"]
+autodoc_mock_imports = [
+    "torch",
+    "fastai",
+    "psycopg",  # Required by databricks_ai_bridge.lakebase
+    "psycopg_pool",  # Required by databricks_ai_bridge.lakebase
+    "langgraph.checkpoint.postgres",  # Required by databricks_langchain.checkpoint
+]
```
jennsun, line 318: do we still have to do this if these dependencies/packages are lazily imported? We updated to do so as part of the this fix, so not sure if it automatically has to be imported here? 
https://github.com/databricks/databricks-ai-bridge/pull/224

fanzeyi: I am guessing that we still need to. We are still going to build docs for classes that uses `lakebase` and `psycopg.

fanzeyi (approved):
LGTM

PR #224, opened 2025-11-20
Langchain Dependency Bug Fix: group checkpoint-postgres package with databricks-ai-bridge[memory]

jennsun:
With the newly added `langgraph-checkpoint-postgres>=2.0.0` dependency to langchain, customers are seeing error when `psycopg[binary,pool]>=3.1` is not installed:

```
mlflow.exceptions.MlflowException: Failed to import code model from /model/agent.py. Error: no pq wrapper available.
```

We need to bundle "langgraph-checkpoint-postgres>=2.0.0" and  "psycopg[binary,pool]>=3.1" together - this pr adds them as an optional dep of databricks-langchain package so they are only installed when user wants to access memory features

testing via notebook:

without [memory]: https://eng-ml-inference.staging.cloud.databricks.com/editor/notebooks/657935421316087?o=1653573648247579#command/657935421316094 

can import but not instantiate:
image

with [memory]: https://eng-ml-inference.staging.cloud.databricks.com/editor/notebooks/657935421316132?o=1653573648247579
can import and instantiate:
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -86,7 +86,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          pip install .
+          pip install .[memory]
           pip install "integrations/langchain[dev, memory]"
       - name: Run tests
         run: |
--- a/integrations/langchain/tests/unit_tests/test_checkpoint.py
+++ b/integrations/langchain/tests/unit_tests/test_checkpoint.py
@@ -9,6 +9,7 @@ pytest.importorskip("psycopg_pool")
 pytest.importorskip("langgraph.checkpoint.postgres")
 
 from databricks_ai_bridge import lakebase
+
 from databricks_langchain import CheckpointSaver
 
 
--- a/src/databricks_ai_bridge/lakebase.py
+++ b/src/databricks_ai_bridge/lakebase.py
@@ -12,11 +12,11 @@ try:
     import psycopg
     from psycopg.rows import dict_row
     from psycopg_pool import ConnectionPool
-except ImportError:
-    raise RuntimeError(
+except ImportError as e:
+    raise ImportError(
         "LakebasePool requires databricks-ai-bridge[memory]. "
         "Please install with: pip install databricks-ai-bridge[memory]"
-    ) from None
+    ) from e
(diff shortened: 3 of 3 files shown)


integrations/langchain/pyproject.toml
```diff
@@ -34,6 +33,11 @@ integration = [
   "anyio",
 ]
 
+memory = [
+  "langgraph-checkpoint-postgres>=2.0.0",
```
bbqiu, line 37: we should just list databricks-ai-bridge[memory] as an extra here - example: https://github.com/unitycatalog/unitycatalog/blob/main/ai/integrations/langchain/pyproject.toml#L22-L26

make sure to pin this to be >=0.10.0

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -1,8 +1,20 @@
...
+    raise RuntimeError(
+        "psycopg is needed to enable checkpoint feature. Please install with databricks-langchain[memory]"
+    ) from None
+
+try:
+    from langgraph.checkpoint.postgres import PostgresSaver
+except ImportError:
+    raise RuntimeError(
+        "langgraph-checkpoint-postgres is needed to enable checkpoint feature. Please install with databricks-langchain[memory]"
+    ) from None
```
bbqiu: nit: we can combine this into one block and say that this `CheckpointSaver` requires databricks-langchain[memory]

integrations/langchain/pyproject.toml
smurching: +1, seems like we should have that extra which will pull in `"psycopg[binary,pool]>=3.1",` and additionally list `"langgraph-checkpoint-postgres>=2.0.0"` here

integrations/langchain/pyproject.toml
```diff
@@ -11,13 +11,12 @@ requires-python = ">=3.10"
 dependencies = [
     "langchain>=1.0.0",
     "databricks-vectorsearch>=0.50",
-    "databricks-ai-bridge>=0.4.2",
+    "databricks-ai-bridge>=0.10.0",
```
bbqiu: nit: we don't need to update this

PR #228, opened 2025-11-21
Langchain MCP Adapters

aravind-segu:
Introduce Langchain MCP Adapters to make the MCP client easier to use.

Example Notebook: https://e2-spog.staging.cloud.databricks.com/editor/notebooks/367820108530247?o=6051921418418893#command/8418062200632198

### DIFF
--- a/integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
+++ b/integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
@@ -21,6 +21,7 @@ class MCPServer(BaseModel):
         - httpx_client_factory: Callable - Custom httpx client factory
         - terminate_on_close: bool - Terminate connection on close
         - session_kwargs: dict - Additional session kwargs
+        - handle_tool_error: bool | str | Callable - Error handling strategy
 
     Example:
         ```python
@@ -48,11 +49,9 @@ class MCPServer(BaseModel):
         default=None,
         exclude=True,
         description=(
-            "How to handle errors raised by tools from this server. Options:\n"
-            "- None/False: Raise the error\n"
-            "- True: Return error message as string\n"
-            "- str: Return this string when errors occur\n"
-            "- Callable: Function that takes error and returns error message string"
+            "If True, return the error message as the output. If False, raise the error. "
+            "If a string, return the string as the error message. "
+            "If a callable, return the result of the callable as the error message."
         ),
     )
 
@@ -105,9 +104,9 @@ class DatabricksMCPServer(MCPServer):
         exclude=True,
     )
 
-    def model_post_init(self, __context: Any) -> None:
+    def model_post_init(self, context: Any) -> None:
         """Initialize DatabricksServer with auth setup."""
-        super().model_post_init(__context)
+        super().model_post_init(context)
 
         # Set up Databricks OAuth authentication after initialization
         if self.workspace_client is None:


integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
```diff
@@ -0,0 +1,223 @@
...
+        client = DatabricksMultiServerMCPClient(
+            [
+                # Databricks server with automatic OAuth - just pass params as kwargs!
+                DatabricksMCPServer(
+                    name="databricks-prod",
+                    url="https://your-workspace.databricks.com/mcp",
+                    workspace_client=WorkspaceClient(),
+                    timeout=30.0,
+                    sse_read_timeout=60.0,
+                    handle_tool_error=True,  # Return errors as strings instead of raising
```
smurching, line 156: QQ is this `handle_tool_error` param standard/necessary to start with?

aravind-segu: We are starting with this for bricks. Bricks is going to use it to handle tool errors and then show the tool call confirmations correctly.

bbqiu: i think we should still include this since it's a base langchain tool concept, but won't the bricks folks encounter it at `get_tools` time?

smurching: Just to help usability, is there a langchain doc page we can link to to explain? I saw the docstring above that keyword parameters like `hanlde_tool_error` get passed through to langchain's connection object, but IDK what that is :D. Maybe in DatabricksMCPServer's docstring we can link the relevant LangChain docs for LangChain's connection object?

smurching: Googling I did find https://reference.langchain.com/python/langgraph/agents/#langgraph.prebuilt.tool_node.ToolNode:~:text=Configuration%20for%20error%20handling%20during%20tool%20execution.%20Supports%20multiple%20strategies%3A but that seems different

integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
```diff
@@ -0,0 +1,223 @@
...
+        default=None,
+        description="Databricks WorkspaceClient for authentication. If None, will be auto-initialized.",
+        exclude=True,
+    )
+
+    def __init__(self, **data):
+        """Initialize DatabricksServer with auth setup."""
+        super().__init__(**data)
+
+        # Set up Databricks OAuth authentication after initialization
```
smurching, line 111: According to ChatGPT, it might be better to do this in `model_post_init`: https://chatgpt.com/share/6926bd22-9194-8009-a58e-fae3261f338b

integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
aravind-segu: Oh didnt know there were two more types, will add that in. I was trying to have the same parameter here as langchain

aravind-segu: I think the Langgraph toolNode is different. In Langchain we return BaseTool defined here: https://github.com/langchain-ai/langchain/blob/0a6d01e61df86b70d5b9afdb1ba68e7b2943e787/libs/core/langchain_core/tools/base.py#L483

But I cant find documentation for it

integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
```diff
@@ -0,0 +1,223 @@
...
+        ```
+    """
+
+    workspace_client: WorkspaceClient | None = Field(
+        default=None,
+        description="Databricks WorkspaceClient for authentication. If None, will be auto-initialized.",
+        exclude=True,
+    )
+
+    def model_post_init(self, __context: Any) -> None:
```
bbqiu: should this be `context` instead of `__context`?

https://docs.pydantic.dev/latest/concepts/models/#defining-a-custom-__init__

bbqiu (approved):
LGTM with a nit! please also remove or document the handle_tool_error param before merging

integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
bbqiu: if this is too ambiguous to document clearly and it's not going to help the bricks folks right now (since the error will be on get instead of execute), so we can skip including it for now?

PR #233, opened 2025-12-04
Timeout Fix

aravind-segu:
Looks like the `client_session_timeout_seconds` default to 5 seconds was stopping our original timeout from being respected. Github issue: https://github.com/openai/openai-agents-python/issues/1026

Updated the code to pass along the same timeout to cleint_session_timeout_seconds if its not explicitly specified.

### Tests
Added Unit Tests
Successfully ran a 15 second tool
image


bbqiu (approved):
LGTM, thanks

PR #236, opened 2025-12-10
Adding DCR support to create UC Connections for external clients for MCP

sunishsheth2009:
Adding DCR support to create UC Connections for external clients for MCP

Tested:
result = register_mcp_client_via_dcr("test_notion_connection", "https://mcp.notion.com/mcp")
print(result)

Connection created:
https://e2-dogfood.staging.cloud.databricks.com/explore/connections/test_notion_connection?o=6051921418418893&activeTab=overview

### DIFF
--- a/databricks_mcp/src/databricks_mcp/connector.py
+++ b/databricks_mcp/src/databricks_mcp/connector.py
@@ -1,0 +1,615 @@
+"""MCP OAuth Discovery and Dynamic Client Registration for Databricks.
+
+This module provides functionality for discovering OAuth metadata and performing
+Dynamic Client Registration (DCR) for MCP (Model Context Protocol) connections.
+"""
+
+import json
+import logging
+import re
+from typing import Any, Dict, Optional, Tuple
+from urllib.parse import urljoin, urlparse
+
+import requests
+from databricks.sdk import WorkspaceClient
+from databricks.sdk.service import catalog
+from databricks_ai_bridge.utils.annotations import experimental
+
+# Configure logging
+logger = logging.getLogger(__name__)
+
+# Constants
+DEFAULT_TIMEOUT = 10
+DEFAULT_PORT = "443"
+WWW_AUTHENTICATE_HEADER = "WWW-Authenticate"
+CONTENT_TYPE_JSON = "application/json"
+
+# ---------------- Helpers ----------------
+
+
+def _fetch_json(url: str, description: str) -> Dict[str, Any]:
+    """Fetch and parse JSON from a URL.
+
+    Args:
+        url: The URL to fetch from
+        description: Human-readable description for error messages
+
+    Returns:
+        Parsed JSON as a dictionary
+
+    Raises:
+        RuntimeError: If the request fails or JSON parsing fails
+    """
+    logger.debug(f"Fetching {description} from {url}")
+
+    try:
+        resp = requests.get(url, timeout=DEFAULT_TIMEOUT)
+        resp.raise_for_status()
+    except requests.exceptions.Timeout as e:
+        raise RuntimeError(f"Timeout while fetching {description} from {url}") from e
+    except requests.exceptions.RequestException as e:
+        raise RuntimeError(f"Failed to fetch {description} from {url}: {e}") from e
+
+    try:
+        data = resp.json()
+        logger.debug(f"Successfully fetched {description}")
+        return data
+    except ValueError as e:
+        raise RuntimeError(
+            f"Failed to parse {description} from {url} as JSON: {e}. Response: {resp.text[:200]}"
+        ) from e
+
+
+def _parse_www_authenticate(header_value: str) -> Dict[str, str]:
+    """Parse WWW-Authenticate header value into key-value parameters.
+
+    Parses headers like:
+        'Bearer resource_metadata="https://example.com/meta", scope="read write"'
+
+    Args:
+        header_value: The WWW-Authenticate header value
+
+    Returns:
+        Dictionary of parsed parameters from the header (quotes are removed from values)
+
+    Example:
+        >>> _parse_www_authenticate('Bearer scope="read write" realm="api"')
+        {'scope': 'read write', 'realm': 'api'}
+    """
+    if not header_value:
+        return {}
+
+    # Split off the auth scheme (e.g., "Bearer") from parameters
+    parts = header_value.split(" ", 1)
+    params_part = parts[1] if len(parts) == 2 else parts[0]
+    params = {}
+
+    # Extract key="value" pairs using regex
+    # Pattern: (\w+) captures parameter name, ([^"]*) captures value between quotes
+    for m in re.finditer(r'(\w+)\s*=\s*"([^"]*)"', params_part):
+        # Store without quotes: group(1) is key, group(2) is value
+        params[m.group(1)] = m.group(2)
+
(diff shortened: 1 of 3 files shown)


databricks_mcp/src/databricks_mcp/connector.py
```diff
@@ -0,0 +1,572 @@
...
+        Dictionary of parsed parameters from the header
+    """
+    if not header_value:
+        return {}
+
+    parts = header_value.split(" ", 1)
+    params_part = parts[1] if len(parts) == 2 else parts[0]
+    params = {}
+
+    for m in re.finditer(r'(\w+)\s*=\s*"([^"]*)"', params_part):
```
aravind-segu, line 89: Nit can we add a comment on what this regex does so its easier to maintain this code later

databricks_mcp/src/databricks_mcp/connector.py
```diff
@@ -0,0 +1,572 @@
...
+    Args:
+        header_value: The WWW-Authenticate header value
+
+    Returns:
+        Dictionary of parsed parameters from the header
+    """
+    if not header_value:
+        return {}
+
+    parts = header_value.split(" ", 1)
```
aravind-segu, line 83: Maybe add a comment that the header might look like this:

Bearer realm="OAuth", error="invalid_token", error_description="Missing or invalid access token"

databricks_mcp/src/databricks_mcp/connector.py
```diff
@@ -0,0 +1,572 @@
...
+    params_part = parts[1] if len(parts) == 2 else parts[0]
+    params = {}
+
+    for m in re.finditer(r'(\w+)\s*=\s*"([^"]*)"', params_part):
+        params[m.group(1)] = m.group(2)
```
aravind-segu, line 91: Looks like we are capturing the `"` marks in the regex above, can we remove them when adding to the dictionary?

databricks_mcp/src/databricks_mcp/connector.py
```diff
@@ -0,0 +1,572 @@
...
+        return params["resource_metadata"]
+
+    # Some servers use variations
+    for key in ("authorization_uri", "resource", "as_uri", "auth_uri"):
+        if key in params:
+            logger.debug(f"Found {key} in WWW-Authenticate header")
+            return params[key]
+
+    # Very last resort: first URL in the header
+    m = re.search(r"https?://[^\s,>]+", www_auth_header)
```
aravind-segu, line 126: Is this a valid spec? Its not documented [here](https://datatracker.ietf.org/doc/html/rfc9728#section-5.1)

databricks_mcp/src/databricks_mcp/connector.py
```diff
@@ -0,0 +1,572 @@
...
+        logger.debug(f"Using issuer: {as_base}")
+    else:
+        raise RuntimeError(
+            "Could not determine authorization server URL from Protected Resource metadata. "
+            "Expected one of: authorization_servers, authorization_server, or issuer. "
+            f"Available keys: {list(resource_meta.keys())}"
+        )
+
+    # Build candidate well-known URLs
+    candidates = [
```
aravind-segu: This makes sense for issues without path components, lets also add in the stuff for the path components specified here: https://modelcontextprotocol.io/specification/draft/basic/authorization#authorization-server-metadata-discovery
image

databricks_mcp/src/databricks_mcp/connector.py
```diff
@@ -0,0 +1,572 @@
...
+        >>> print(result["client_id"])
+    """
+    if not connection_name:
+        raise ValueError("connection_name cannot be empty")
+    if not mcp_url:
+        raise ValueError("mcp_url cannot be empty")
+
+    logger.info(f"Starting MCP client registration for connection: {connection_name}")
+    logger.info(f"MCP URL: {mcp_url}")
+
```
aravind-segu, line 578: just curious on DCR, how does this work with duplicates? If a user runs this twice will we create two difference dynamic apps and then two different UC Connections? Should we first check whether a uc connection exist before firing of creating a new one?

databricks_mcp/src/databricks_mcp/connector.py
```diff
@@ -0,0 +1,572 @@
...
+        "redirect_uris": [redirect_uri],
+        "grant_types": ["authorization_code"],
+        "response_types": ["code"],
+        "token_endpoint_auth_method": "none",
+    }
+    logger.debug(f"Registration payload: {json.dumps(payload, indent=2)}")
+
+    # Perform registration
+    try:
+        resp = requests.post(
```
aravind-segu, line 418: same question on duplicates, can we have two different dcr apps with Databricks as Client Name? Is that ok?

databricks_mcp/src/databricks_mcp/connector.py
```diff
@@ -0,0 +1,572 @@
...
+    reg["registration_method"] = "dcr"
+    reg["scope"] = selected_scope  # Store selected scope for UC connection
+
+    return reg
+
+
+def create_uc_connection(
+    mcp_url: str,
+    connection_name: str,
+    as_meta: Dict[str, Any],
```
aravind-segu: nit: lets rename this to dcr_result, i was confused reading this

databricks_mcp/src/databricks_mcp/connector.py
```diff
@@ -0,0 +1,572 @@
...
+        # Step 3: Perform Dynamic Client Registration
+        dcr_result = perform_dynamic_client_registration(
+            as_meta, prm, www_auth_header, workspace_client
+        )
+
+        # Step 4: Create Unity Catalog Connection
+        create_uc_connection(mcp_url, connection_name, dcr_result, workspace_client)
+
+        logger.info("Successfully completed MCP client registration")
+        return dcr_result
```
aravind-segu: Not sure if this is good to return with all the client id and client secret information. Maybe we should return the [url](https://databricks-sdk-py.readthedocs.io/en/latest/dbdataclasses/catalog.html#databricks.sdk.service.catalog.ConnectionInfo) of the connection which is returned from `create_uc_connection`

aravind-segu (commented):
Super cool, added some comments. Had an overall question on duplicates and how we can prevent someone from running this twice

databricks_mcp/src/databricks_mcp/connector.py
```diff
@@ -0,0 +1,615 @@
...
+    logger.info(f"MCP URL: {mcp_url}")
+
+    # Check if connection already exists to prevent duplicates
+    w = workspace_client or WorkspaceClient()
+    try:
+        w.connections.get(connection_name)
+        logger.info(f"Connection '{connection_name}' already exists, skipping DCR")
+        # Construct workspace URL for the existing connection
+        workspace_id = w.get_workspace_id()
+        connection_url = f"{w.config.host}/explore/connections/{connection_name}?o={workspace_id}&activeTab=overview"
```
aravind-segu, line 586: small nit: connection info contains a url can we just return that instead? Same thing in the create_uc_connection above

aravind-segu (approved):
2 nits, but lgtm

databricks_mcp/src/databricks_mcp/connector.py
sunishsheth2009: Discussed offline that the url in the connection is:
image

which is the combination of host_name + base_path + port
and not the URL to the connection.

So manually crafting would help :)

PR #238, opened 2025-12-12
Fixes #211 with_structured_output for json_schema method

### LINKED ISSUE #211: missing "name" for ChatDatabricks.with_structured_output(method="json_schema")
Hi,

When using ChatDatabricks.with_structured_output(method="json_schema"), there is this error:

```
{
  "external_model_provider":"openai",
  "external_model_error": {
    "error":{
      "message":"Missing required parameter: 'response_format.json_schema.name'.",
      "type":"invalid_request_error",
      "param":"response_format.json_schema.name",
      "code":"missing_required_parameter"
    }
  }
}
```

The problem is within this code block:
https://github.com/databricks/databricks-ai-bridge/blob/main/integrations/langchain/src/databricks_langchain/chat_models.py#L973

"response_format" should have another key named "name", per GPT's request.

WIll-Xu35:
Fixes #211 

Add "name" key for response_format. It can optional input with key "schema_name". Otherwise default to "json_schema"

Change schema value type from tuple to dict.

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -975,6 +975,7 @@ class ChatDatabricks(BaseChatModel):
             response_format = {
                 "type": "json_schema",
                 "json_schema": {
+                    "name": kwargs.get("schema_name", "json_schema"),
                     "strict": True,
                     "schema": schema.model_json_schema() if is_pydantic_schema else schema,  # type: ignore[union-attr]
                 },


integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -975,10 +975,9 @@ class AnswerWithJustification(BaseModel):
             response_format = {
                 "type": "json_schema",
                 "json_schema": {
+                    "name": kwargs.get("schema_name", "json_schema"),
```
jennsun, line 978: the name should be derived and not hardcoded - here you should use [convert_to_openai_function()](https://github.com/langchain-ai/langchain/blob/202d7f6c4a2ca8c7e5949d935bcf0ba9b0c23fb0/libs/core/langchain_core/utils/function_calling.py#L333C5-L333C31) ,(add an additional import from the langchain core function calling class for `from langchain_core.utils.function_calling import convert_to_openai_function` . we should also do this to address the other cascading errors (see my overall review comment) 

can refer to how the partner library is doing this as a reference which calls convert_to_openai_function: 
https://github.com/langchain-ai/langchain/blob/202d7f6c4a2ca8c7e5949d935bcf0ba9b0c23fb0/libs/partners/openai/langchain_openai/chat_models/base.py#L1449

suggested fix may look something like this, but please be sure to test!

```
from langchain_core.utils.function_calling import (
      convert_to_openai_function,  # add this import
      convert_to_openai_tool,
  )

  # ... in with_structured_output, json_schema branch:

  elif method == "json_schema":
      if schema is None:
          raise ValueError(
              "schema must be specified when method is 'json_schema'. Received None."
          )
      function = convert_to_openai_function(
          pydantic_schema if pydantic_schema else schema, strict=True
      )
      function["schema"] = function.pop("parameters")
      response_format = {"type": "json_schema", "json_schema": function}
      llm = self.bind(response_format=response_format)
```

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -975,10 +975,9 @@ class AnswerWithJustification(BaseModel):
...
                     "strict": True,
-                    "schema": (
-                        schema.model_json_schema() if is_pydantic_schema else schema  # type: ignore[union-attr]
-                    ),
+                    "schema": schema.model_json_schema() if is_pydantic_schema else schema,  # type: ignore[union-attr]
                 },
```
jennsun, line 981: the pr is also missing unit test coverage - we have an existing test at `tests/unit_tests/test_chat_models.py` (test_chat_model_with_structured_output) 

in which this line
`assert bind["response_format"]["json_schema"]["schema"] == JSON_SCHEMA` would likely fail. 

can you update this test and include asserts for the name/strict/schema fields? may look something like:
```
elif method == "json_schema":
      js = bind["response_format"]["json_schema"]
      assert js["name"] == "AnswerWithJustification"
      assert js["strict"] is True
      assert js["schema"]["additionalProperties"] is False
```

can you also add the follow unit tests:
- pydantic schema produces a valid response_format
- raw dict schema produces a valid response_format
- nested Pydantic model to validate $ref dereferencing?

PR #239, opened 2025-12-12
Initial draft of LangChain TypeScript Integration for Databricks Model Serving

eLysgaard:
This PR introduces an initial version of @databricks/langchain-ts, a TypeScript package that integrates Databricks Model Serving with the LangChain JS SDK. This breaks ground on adding a TypeScript equivalent of the existing Python databricks-langchain integration.

Supports both Chat Completions and Responses.

I've added integration tests that run against FMAPI using both of these API schemas.

  ## Summary

  Adds `@databricks/langchain-ts`, a LangChain TypeScript integration for Databricks Model Serving.

  ## Features

  - `ChatDatabricks` class compatible with LangChain's `BaseChatModel` interface
  - Supports streaming responses via `stream()`
  - Supports tool/function calling via `bindTools()`
  - Three endpoint types: Chat Completions, Chat Agent, and Responses
  - Authentication via Databricks SDK

  ## Usage

  ```typescript
  import { ChatDatabricks } from "@databricks/langchain-ts";

  const model = new ChatDatabricks({
    endpoint: "databricks-meta-llama-3-3-70b-instruct",
  });

  // Simple invocation
  const response = await model.invoke("What is the capital of France?");

  // Streaming
  const stream = await model.stream("Tell me a story");
  for await (const chunk of stream) {
    process.stdout.write(chunk.content as string);
  }

  // Tool calling
  const modelWithTools = model.bindTools([myTool]);
  const response = await modelWithTools.invoke("What's the weather?");
  ```
   ## Implementation

  - Uses @databricks/ai-sdk-provider internally for API communication
  - Uses @databricks/sdk for authentication
  - Message conversion utilities for LangChain ↔ AI SDK formats
  - Tool conversion supporting OpenAI format, Zod schemas, and plain objects

    ## Files

  | File                  | Description                            |
  |-----------------------|----------------------------------------|
  | src/chat_models.ts    | Main ChatDatabricks implementation     |
  | src/utils/messages.ts | Message format conversion              |
  | src/utils/tools.ts    | Tool format conversion                 |
  | examples/             | Basic and advanced tool usage examples |
  | tests/                | Unit and integration tests             |

### DIFF
--- a/integrations/langchainjs/README.md
+++ b/integrations/langchainjs/README.md
@@ -51,7 +51,6 @@ const model = new ChatDatabricks({
 });
 ```
 
-
 ### Responses
 
 Rich output with reasoning, citations, and function calls.
@@ -81,10 +80,11 @@ const model = new ChatDatabricks({
 ```
 
 ### Environment Variables
+
 The following environment variables are supported
 
 | Variable                          | Description                                          |
-|-----------------------------------|------------------------------------------------------|
+| --------------------------------- | ---------------------------------------------------- |
 | DATABRICKS_HOST                   | Workspace or account URL                             |
 | DATABRICKS_TOKEN                  | Personal access token                                |
 | DATABRICKS_CLIENT_ID              | OAuth client ID / Azure client ID                    |
--- a/integrations/langchainjs/src/chat_models.ts
+++ b/integrations/langchainjs/src/chat_models.ts
@@ -73,10 +73,10 @@ export interface ChatDatabricksInput extends BaseChatModelParams {
 
   /**
    * Whether to use the Responses API or Chat Completions API
-   * 
+   *
    * - Chat Completions: See https://docs.databricks.com/aws/en/machine-learning/foundation-model-apis/api-reference#chat-completions-api
    * - Responses: See https://docs.databricks.com/aws/en/machine-learning/foundation-model-apis/api-reference#responses-api
-   * 
+   *
    * @default false
    */
   useResponsesApi?: boolean;
--- a/integrations/langchainjs/tests/chat_models.int.test.ts
+++ b/integrations/langchainjs/tests/chat_models.int.test.ts
@@ -253,20 +253,20 @@ describe.skipIf(!hasCredentials)("ChatDatabricks Integration Tests", () => {
         // Responses API doesn't support stop sequences
         // See https://docs.databricks.com/aws/en/machine-learning/foundation-model-apis/api-reference#responses-api-request
         .skipIf(useResponsesApi)("respects stop sequences", async () => {
-          const modelWithStop = new ChatDatabricks({
-            endpoint,
-            useResponsesApi,
-            maxTokens: 100,
-            stop: ["5"],
-          });
-
-          const response = await modelWithStop.invoke([
-            new HumanMessage("Count from 1 to 10, one number per line"),
-          ]);
-
-          const content = response.content as string;
-          expect(content).not.toContain("6");
+        const modelWithStop = new ChatDatabricks({
+          endpoint,
+          useResponsesApi,
+          maxTokens: 100,
+          stop: ["5"],
         });
+
+        const response = await modelWithStop.invoke([
+          new HumanMessage("Count from 1 to 10, one number per line"),
+        ]);
+
+        const content = response.content as string;
+        expect(content).not.toContain("6");
+      });
     });
 
     describe("Error Handling", () => {


integrations/langchain-ts/.env.example
```diff
@@ -0,0 +1,11 @@
...
+# Copy this file to .env.local and fill in your values
+
+# Your Databricks workspace URL
+DATABRICKS_HOST="e2-dogfood.staging.cloud.databricks.com"
+
+# Your Databricks personal access token
+DATABRICKS_TOKEN="dapi..."
+
+# (Optional) Model serving endpoint to use for examples/tests
+TEST_ENDPOINT_NAME="databricks-meta-llama-3-3-70b-instruct"
```
bbqiu: btw i think this is getting deprecated soon, might want to switch to somethign like claude sonnet

integrations/langchain-ts/src/chat_models.ts
```diff
@@ -0,0 +1,411 @@
...
+  extraParams?: Record<string, unknown>;
+}
+
+/**
+ * ChatDatabricks - Chat model integration for Databricks Model Serving
+ *
+ * Supports three endpoint types:
+ * - FMAPI: OpenAI-compatible foundation models
+ * - ChatAgent: Databricks agent endpoints
+ * - ResponsesAgent: Rich output with reasoning and citations
```
bbqiu: i think we can revise this language to be responses API / responsesagent where possible since this code can be reused to query Responses API LLMs, not just responsesagent models (same API though!)

integrations/langchain-ts/README.md
```diff
@@ -0,0 +1,338 @@
...
+```
+
+### Responses Agent
+
+Rich output with reasoning, citations, and function calls.
+
+```typescript
+const model = new ChatDatabricks({
+  endpoint: "my-responses-agent",
+  endpointType: "responses-agent",
```
bbqiu: instead of this, is it possible to follow the logic from the python chatdatabricks where we have a singular flag `use_responses_api`? this follows the official langchain `ChatOpenAI` client API. you send the same input for chat-agent and fmapi chat completions, so they can be consolidated into one case, you'll just have to parse the output differently within the ai-sdk (you can see the code in python chatdatabricks here, just search for "chatagent": https://github.com/databricks/databricks-ai-bridge/blob/8ce93d3418466e98581ecdb565b131b5fd8f6686/integrations/langchain/src/databricks_langchain/chat_models.py#L541)

https://github.com/search?q=repo:langchain-ai/langgraph+OR+repo:databricks/databricks-ai-bridge+OR+repo:langchain-ai/langchain+OR+repo:unitycatalog/unitycatalog+OR+repo:mlflow/mlflow+use_responses_api&type=code

bbqiu (commented):
overall lgtm -- please test against a responses agent (like one hosted on apps) and a responses API LLM that is served on db

integrations/langchain-ts/README.md
eLysgaard: Hmm we have to specify which AI SDK language model to use. We don't do duck typing when parsing the output.

bbqiu: ah i just mean to match the API params of langchain JS more closely. example: https://docs.langchain.com/oss/javascript/integrations/chat/openai#code-interpreter

more generic responses API docs: https://docs.langchain.com/oss/javascript/integrations/chat/openai#responses-api

so we could have somethign like

```
new ChatDatabricks(
model: "db_endpoint_name",
useResponsesAPI: true
)
```

integrations/langchainjs/.env.example
```diff
@@ -0,0 +1,12 @@
+# Databricks Authentication
+# Copy this file to .env.local and fill in your values
+
+# Your Databricks workspace URL
+DATABRICKS_HOST="e2-dogfood.staging.cloud.databricks.com"
```
bbqiu: nit: can we recommend users use their oauth token and `DATABRICKS_CONFIG_PROFILE` instead of host / token

integrations/langchain-ts/README.md
bbqiu: this means we could also make the `endpointAPI` param non-required (will default to chat completions, matching the official langchain client)

integrations/langchainjs/src/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @databricks/langchain-ts - LangChain integration for Databricks Model Serving
```
bbqiu: ai nit: are we going w/ this name or databricks/langchainjs

eLysgaard: Ah missed this! I decided to rename the project `@databricks/langchainjs` to match the name of https://github.com/langchain-ai/langchainjs

integrations/langchainjs/.env.example
eLysgaard: Sure!

eLysgaard: Look like this env var isn't supported when using the TS SDK (yet?). The full set of supported env vars is here https://www.npmjs.com/package/@databricks/sdk-experimental#environment-variables

integrations/langchain-ts/README.md
eLysgaard: Makes sense. I'll change this to a boolean control which flips between chat completions (as default) and responses

PR #241, opened 2025-12-15
Pass in embeddings to Databricks Store for Vector Search Semantic Retrieval

jennsun:
Adds semantic search support to DatabricksStore by enabling the index configuration when creating the underlying [PostgresStore](https://reference.langchain.com/javascript/classes/_langchain_langgraph-checkpoint-postgres.store.PostgresStore.html). LangGraph's PostgresStore supports semantic search through an [index parameter](https://reference.langchain.com/javascript/interfaces/_langchain_langgraph-checkpoint-postgres.store.PostgresStoreConfig.html) that configures an embedding model to vectorize stored memories, allowing store.search() to find items by meaning rather than just listing by key match. Previously, our DatabricksStore wrapper did not expose this capability, limiting users to basic key-value retrieval. 

This change adds embedding_endpoint, embedding_dims, and embedding_fields parameters to `DatabricksStore.__init__`, which internally creates a DatabricksEmbeddings instance and passes it to PostgresStore via the index={"embed": embeddings, "dims": dims, "fields": fields} configuration. This allows Databricks users to leverage their existing Model Serving embedding endpoints (e.g., databricks-gte-large-en) for semantic memory search in LangGraph agents https://docs.databricks.com/aws/en/machine-learning/foundation-model-apis/supported-models#gte-large-en (used in other Databricks examples like vector search https://docs.databricks.com/aws/en/vector-search/create-vector-search)

Also added bug fix to embeddings logic since the Model Serving API returns JSON, where:
- 0 is parsed as int and 0.5 is parsed as float. PostgreSQL/pgvector requires homogeneous arrays, so mixed types cause an error like:
`DataError: cannot dump lists of mixed types; got: float, int`
patching `embeddings.py` because the type annotation already says List[List[float]] - the fix makes the code match its contract, float(0) = 0.0 is mathematically identical, and all downstream code expects floats anyway

Usage:
1. using embedding_endpoint:
```
store = DatabricksStore(
    instance_name="my-lakebase-instance",
    embedding_endpoint="databricks-gte-large-en",
    embedding_dims=1024,
)
```
2. using pre-configured embeddings instance:
```
embeddings = DatabricksEmbeddings(endpoint="databricks-gte-large-en")

store = DatabricksStore(
    instance_name="my-lakebase-instance",
    embeddings=embeddings,
    embedding_dims=1024,
)
``` 
3. custom embedding fields:
```
store = DatabricksStore(
    instance_name="my-lakebase-instance",
    embedding_endpoint="databricks-gte-large-en",
    embedding_dims=1024,
    embedding_fields=["text"],  # Only embed the "text" field, not entire JSON
)
```
rest of store api is unchanged, use like:
```
store.put(namespace, memory_key, memory_data)
self.store.search(namespace, query=query, limit=5)
```

Example notebook: https://eng-ml-inference-team-eu-central-1.cloud.databricks.com/editor/notebooks/1322463895224394?o=5763476629593792

Tested storage, retrieval, deletion in lakebase work
image
image
image
image
image
image

example endpoint: https://eng-ml-liteswap-us-east-1-3.staging.cloud.databricks.com/ml/endpoints/agents_main-default-long-term-memory-agent-embeddings?o=1929317688015595
image

### DIFF
--- a/integrations/langchain/src/databricks_langchain/store.py
+++ b/integrations/langchain/src/databricks_langchain/store.py
@@ -43,6 +43,21 @@ class DatabricksStore(BaseStore):
         embeddings: Optional[DatabricksEmbeddings] = None,
         **pool_kwargs: Any,
     ) -> None:
+        """Initialize DatabricksStore with embedding support.
+
+        Args:
+            instance_name: The name of the Lakebase instance to connect to.
+            workspace_client: Optional Databricks WorkspaceClient for authentication.
+            embedding_endpoint: Name of the Databricks Model Serving endpoint for embeddings
+                (e.g., "databricks-gte-large-en"). If provided, enables semantic search.
+            embedding_dims: Dimension of the embedding vectors (e.g., 1024 for gte-large-en,
+                1536 for OpenAI-compatible models). Required if embedding_endpoint is set.
+            embedding_fields: List of JSON paths to vectorize. Defaults to ["$"] which
+                vectorizes the entire JSON value.
+            embeddings: Optional pre-configured DatabricksEmbeddings instance. If provided,
+                takes precedence over embedding_endpoint.
+            **pool_kwargs: Additional keyword arguments passed to LakebasePool.
+        """
         if not _store_imports_available:
             raise ImportError(
                 "DatabricksStore requires databricks-langchain[memory]. "


integrations/langchain/src/databricks_langchain/embeddings.py
```diff
@@ -59,7 +59,8 @@ def _embed(self, texts: List[str], params: Dict[str, str]) -> List[List[float]]:
                 endpoint=self.endpoint,
                 inputs={"input": txt, **params},  # type: ignore[arg-type]
             )
-            embeddings.extend(r["embedding"] for r in resp["data"])
+            for r in resp["data"]:
+                embeddings.append([float(v) for v in r["embedding"]])
         return embeddings
```
jennsun, line 64: bug fix for embeddings logic: 
0 is parsed as int and 0.5 is parsed as float. PostgreSQL/pgvector requires homogeneous arrays, so mixed types cause an error like:
`DataError: cannot dump lists of mixed types; got: float, int`
type annotation for this class says List[List[float]] - the fix makes the code match its contract, float(0) = 0.0 is mathematically identical, and all downstream code expects floats

integrations/langchain/src/databricks_langchain/store.py
```diff
@@ -47,13 +54,41 @@ def __init__(
...
+            # Use pre-configured embeddings instance
+            self.embeddings = embeddings
+            if embedding_dims is None:
+                raise ValueError("embedding_dims is required when providing an embeddings instance")
+            self.index_config = {
+                "dims": embedding_dims,
+                "embed": self.embeddings,
+                "fields": embedding_fields or ["$"],
+            }
+        elif embedding_endpoint is not None:
```
fanzeyi, line 94: I wonder if we should issue a warning when both `embedding_endpoint` and `embeddings` are specified.

jennsun: good point - let me add a warning here clarifying we prioritize using `embeddings`

integrations/langchain/src/databricks_langchain/store.py
```diff
@@ -33,6 +37,10 @@ def __init__(
         *,
         instance_name: str,
         workspace_client: Optional[WorkspaceClient] = None,
+        embedding_endpoint: Optional[str] = None,
```
bbqiu, line 40: nit: can we add a docstring to this function to better describe how to use this API?

PR #245, opened 2025-12-18
Langchain MCP Adapters Fix

aravind-segu:
Currently in langchain mcp adapters, the DatabricksOAuthClientProvider is defined at initialization and is reused in all future calls. With our current architecture, this gets the same token on every invocation of the request.

Therefore this pr introduces a new httpx_async_client_factory. Now on every initialization, a new DatabricksOAuthClientProvider is initialized prompting a new "get_tokens" call resulting in a refreshed token

### DIFF
--- a/integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
+++ b/integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
@@ -16,6 +16,11 @@ class DatabricksMcpHttpClientFactory(McpHttpClientFactory):
         auth: httpx.Auth | None = None,
     ) -> httpx.AsyncClient:
         if isinstance(auth, DatabricksOAuthClientProvider) and auth.workspace_client is not None:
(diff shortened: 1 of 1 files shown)


integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
```diff
@@ -1,11 +1,34 @@
...
+        self,
+        headers: dict[str, str] | None = None,
+        timeout: httpx.Timeout | None = None,
+        auth: httpx.Auth | None = None,
+    ) -> httpx.AsyncClient:
+        if isinstance(auth, DatabricksOAuthClientProvider) and auth.workspace_client is not None:
+            return httpx.AsyncClient(
+                headers=headers,
+                timeout=timeout,
+                auth=DatabricksOAuthClientProvider(auth.workspace_client),
```
bbqiu, line 27: discussed offline, this ends up making sense, but can we explain the call path for this?

like when does this actuallyl get called and why do we have to reinitialize it each time?

PR #246, opened 2025-12-20
typing: fixing type errors in databricks_ai_bridge

fanzeyi:
[Click here to review](https://ghstack.zeyi.fan/databricks/databricks-ai-bridge/pull/246)

as titled

```
$ uv run --exact --extra memory ty check
All checks passed!
```

### DIFF
--- a/src/databricks_ai_bridge/lakebase.py
+++ b/src/databricks_ai_bridge/lakebase.py
@@ -5,7 +5,7 @@ import logging
 import time
 import uuid
 from threading import Lock
-from typing import Optional
+from typing import Any
 
 from databricks.sdk import WorkspaceClient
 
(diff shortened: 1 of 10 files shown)


src/databricks_ai_bridge/lakebase.py
```diff
@@ -127,7 +142,7 @@ def __init__(
         instance_name: str,
         workspace_client: WorkspaceClient | None = None,
         token_cache_duration_seconds: int = DEFAULT_TOKEN_CACHE_DURATION_SECONDS,
-        **pool_kwargs: object,
+        **pool_kwargs: dict[str, Any],
```
jennsun, line 135: Can you elaborate on the decision to use `dict[str, Any]` here instead of something like `Any` or `object`? Since these kwargs are forwarded directly to Conncetion pool and looks like it may be what's causing the `# type: ignore[invalid-argument-type]` below

fanzeyi: copying from slack: `dict[str, Any]` is the best we could do for typing `kwargs` for now.

src/databricks_ai_bridge/lakebase.py
```diff
@@ -74,14 +84,16 @@ def __init__(
         self.host = resolved_host
         self.username = self._infer_username()
 
-        self._cached_token: Optional[str] = None
-        self._cache_ts: Optional[float] = None
+        self._cached_token = None
```
bbqiu: interesting that the optional notation is against ty rules? isn't it better to have a hint as to what this attribute should contain vs getting nothing?

fanzeyi: This was me changing the style of annotation but it turned out there is a difference. GPT explained it better:

https://chatgpt.com/share/69669f87-b538-8009-a0e1-5945ec2fcab8

Reverting it back to `self.var: type = ...` style.

PR #248, opened 2025-12-23
typing: fixing type errors in databricks_openai

fanzeyi:
[Click here to review](https://ghstack.zeyi.fan/databricks/databricks-ai-bridge/pull/248)

```
[openai]$ uv run ty check
All checks passed!
```

### DIFF
--- a/integrations/openai/src/databricks_openai/agents/mcp_server.py
+++ b/integrations/openai/src/databricks_openai/agents/mcp_server.py
@@ -108,14 +108,14 @@ class McpServer(MCPServerStreamableHttp):
         self.workspace_client = workspace_client
 
         if params is None:
-            params = MCPServerStreamableHttpParams()
+            params = MCPServerStreamableHttpParams(url="")
 
-        if url is None and params.get("url") is None:
+        if url is None and not params.get("url"):
             raise ValueError(
                 "Please provide the url of the MCP Server when initializing the McpServer. McpServer(url=...)"
             )
 
-        if url is not None and params.get("url") is not None and url != params.get("url"):
+        if url is not None and params.get("url") and url != params.get("url"):
             raise ValueError(
                 "Different URLs provided in url and the MCPServerStreamableHttpParams. Please provide only one of them."
             )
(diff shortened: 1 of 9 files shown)


integrations/openai/src/databricks_openai/agents/mcp_server.py
```diff
@@ -108,14 +109,14 @@ def __init__(
         self.workspace_client = workspace_client
 
         if params is None:
-            params = MCPServerStreamableHttpParams()
+            params = MCPServerStreamableHttpParams(url="")
 
-        if url is None and params.get("url") is None:
+        if url is None and not params.get("url"):
```
bbqiu, line 113: this change isn't equivalent right? i'm fine with it since the URL still shouldn't be an empty string

integrations/openai/src/databricks_openai/agents/mcp_server.py
```diff
@@ -286,16 +287,41 @@ def create_streams(
...
+        auth = DatabricksOAuthClientProvider(self.workspace_client)
+
+        timeout_value = self.params.get("timeout", 5)
+        if isinstance(timeout_value, timedelta):
+            timeout: float = float(timeout_value.total_seconds())
+        else:
+            timeout = float(timeout_value)
+
+        sse_read_timeout_value = self.params.get("sse_read_timeout", 60 * 5)
+        if isinstance(sse_read_timeout_value, timedelta):
```
bbqiu: is timedelta the default type that is passed inside the mcp library? or is the consideration that users may pass timedelta values instead

integrations/openai/src/databricks_openai/agents/mcp_server.py
fanzeyi: Yeah. This is intentional because `url` in params cannot be `None` according to the definiton

https://github.com/openai/openai-agents-python/blob/c06a8bb42a24edd18e944a40f8a89cc4f2d36eb4/src/agents/mcp/server.py#L577

I assumed our intent here is to check if the url is not empty.

integrations/openai/src/databricks_openai/agents/mcp_server.py
fanzeyi: I now can't remember why I decided to do this.  It was the case that `streamablehttp_client` only supported `float` type in older version but it's no longer the case.

I don't really see a reason of doing this. Will simplify.

PR #249, opened 2025-12-23
typing: fixing type errors in databricks_langchain

fanzeyi:
```
[langchain]$ uv run ty check
      Built databricks-langchain @ file:///Users/zeyi.f/GitHub/databricks-ai-bridge/integrations/langchain
Uninstalled 1 package in 0.96ms
Installed 1 package in 1ms
All checks passed!
```

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -3,6 +3,7 @@
 import json
 import logging
 import warnings
+from functools import cached_property
 from operator import itemgetter
 from typing import (
     Any,
@@ -16,8 +17,10 @@ from typing import (
     Sequence,
     Type,
     Union,
+    cast,
 )
 
+from databricks import sdk
 from langchain_core.callbacks import CallbackManagerForLLMRun
 from langchain_core.language_models import BaseChatModel
 from langchain_core.language_models.base import LanguageModelInput
@@ -51,7 +54,11 @@ from langchain_core.runnables import Runnable, RunnableMap, RunnablePassthrough
 from langchain_core.tools import BaseTool
 from langchain_core.utils.function_calling import convert_to_openai_tool
 from langchain_core.utils.pydantic import is_basemodel_subclass
+from openai import OpenAI, Stream
+from openai.types.chat import ChatCompletion, ChatCompletionChunk
+from openai.types.responses import Response, ResponseStreamEvent
 from pydantic import BaseModel, ConfigDict, Field
+from typing_extensions import override
 
 from databricks_langchain.utils import get_openai_client
 
@@ -76,10 +83,10 @@ class ChatDatabricks(BaseChatModel):
                 max_retries=3,  # Maximum number of retries
             )
 
-            # Using a WorkspaceClient instance for custom authentication
-            from databricks.sdk import WorkspaceClient
+            # Using a sdk.WorkspaceClient instance for custom authentication
+            from databricks.sdk import sdk.WorkspaceClient
 
-            workspace_client = WorkspaceClient(host="...", token="...")
+            workspace_client = sdk.WorkspaceClient(host="...", token="...")
             llm = ChatDatabricks(
                 model="databricks-claude-3-7-sonnet",
                 workspace_client=workspace_client,
@@ -260,7 +267,7 @@ class ChatDatabricks(BaseChatModel):
     """Name of Databricks Model Serving endpoint to query."""
     target_uri: Optional[str] = None
     """The target MLflow deployment URI to use. Deprecated: use workspace_client instead."""
-    workspace_client: Optional[Any] = Field(default=None, exclude=True)
+    workspace_client: Optional[sdk.WorkspaceClient] = Field(default=None, exclude=True)
     """Optional WorkspaceClient instance to use for authentication. If not provided, uses default authentication."""
     temperature: Optional[float] = None
     """Sampling temperature. Higher values make the model more creative."""
@@ -282,7 +289,6 @@ class ChatDatabricks(BaseChatModel):
     """Timeout in seconds for the HTTP request. If None, uses the default timeout."""
     max_retries: Optional[int] = None
     """Maximum number of retries for failed requests. If None, uses the default retry count."""
-    client: Optional[object] = Field(default=None, exclude=True)  #: :meta private:
 
     @property
     def endpoint(self) -> str:
@@ -304,6 +310,18 @@ class ChatDatabricks(BaseChatModel):
         )
         self.model = value
 
+    @cached_property
+    def client(self) -> OpenAI:
+        # Always use OpenAI client (supports both chat completions and responses API)
+        # Prepare kwargs for the SDK call
+        openai_kwargs = {}
+        if self.timeout is not None:
+            openai_kwargs["timeout"] = self.timeout
+        if self.max_retries is not None:
+            openai_kwargs["max_retries"] = self.max_retries
+
+        return get_openai_client(workspace_client=self.workspace_client, **openai_kwargs)
+
     def __init__(self, **kwargs: Any):
         super().__init__(**kwargs)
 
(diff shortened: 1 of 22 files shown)


integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -363,11 +369,12 @@ def _generate(
 
         if self.use_responses_api:
             # Use OpenAI client with responses API
-            resp = self.client.responses.create(**data)  # type: ignore
+            resp = cast(OpenAI, self.client).responses.create(**data)
+            # TODO: streaming support
```
bbqiu: what is this todo from?

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -363,11 +369,12 @@ def _generate(
 
         if self.use_responses_api:
             # Use OpenAI client with responses API
-            resp = self.client.responses.create(**data)  # type: ignore
+            resp = cast(OpenAI, self.client).responses.create(**data)
```
bbqiu, line 374: is there a way for us to get around type hint issues without casting here?

if not, then can we just do the normal pattern of None | OpenAI and then continue to initialize the client during the init?

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -706,19 +723,24 @@ def _stream(
...
+                            total_tokens=final_usage["total_tokens"],
+                        ),
+                    )
                 )
                 usage_chunk_emitted = True
 
     def bind_tools(
         self,
-        tools: Sequence[Union[Dict[str, Any], Type[BaseModel], Callable, BaseTool]],
+        tools: Sequence[dict[str, Any] | type | Callable | BaseTool],
```
bbqiu: is type accurate here? feels like it should be basemodel instead

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -706,19 +723,24 @@ def _stream(
...
         *,
-        tool_choice: Optional[
-            Union[dict, str, Literal["auto", "none", "required", "any"], bool]
-        ] = None,
+        tool_choice: str | dict | bool | None = None,
         **kwargs: Any,
-    ) -> Runnable[LanguageModelInput, BaseMessage]:
+    ) -> Runnable[LanguageModelInput, AIMessage]:
```
bbqiu, line 745: i believe this can produce multiple kinds of messages (think server side execution)

integrations/langchain/src/databricks_langchain/genie.py
```diff
@@ -170,5 +170,6 @@ def custom_processor(messages):
 
     runnable = RunnableLambda(partial_genie_agent)
     runnable.name = genie_agent_name
-    runnable.description = description or genie.description
+    # TODO: `description` field does not exist on `RunnableLambda`
```
bbqiu, line 173: this is probably bad practice, but i think i added a description attribute always

integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
```diff
@@ -43,7 +44,7 @@ class MCPServer(BaseModel):
 
     Common optional parameters:
         - headers: dict[str, str] - Custom HTTP headers
-        - timeout: float - Request timeout in seconds
+        - timeout: timedelta - Request timeout in seconds
```
bbqiu: same comment as the one in the openai PR -- can we leave these as floats?

integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
```diff
@@ -253,7 +258,7 @@ def model_post_init(self, context: Any) -> None:
         # Store the auth provider internally
         self._auth_provider = DatabricksOAuthClientProvider(self.workspace_client)
 
-    def to_connection_dict(self) -> dict[str, Any]:
+    def to_connection_dict(self) -> StreamableHttpConnection:
```
bbqiu, line 263: i'm p sure these functions just return dicts that are in the shape of a streamablehttpconnection, is this correct?

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
```diff
@@ -84,7 +88,7 @@ def _validate_tool_inputs(self):
         return self
 
     @vector_search_retriever_tool_trace
-    def _run(self, query: str, filters: Optional[List[FilterItem]] = None, **kwargs) -> str:
+    def _run(self, query: str, filters: list[FilterItem] | None = None, **kwargs) -> list[Document]:
```
bbqiu: just an fyi i'm gonna change this here: https://github.com/databricks/databricks-ai-bridge/pull/271

integrations/langchain/src/databricks_langchain/vectorstores.py
```diff
@@ -713,16 +711,16 @@ async def amax_marginal_relevance_search(
...
         fetch_k: int = 20,
         lambda_mult: float = 0.5,
-        filter: Optional[Any] = None,
+        filter: Any | None = None,
         *,
         query_type: Optional[str] = None,
         reranker: Optional[Reranker] = None,
         **kwargs: Any,
-    ) -> List[Document]:
+    ) -> list[Document]:  # ty:ignore[invalid-method-override]: this shouldn't have happened but fixing this may break user's code
```
bbqiu, line 723: would you mind explaining what's happening here

bbqiu (approved):
approving to unblock merge, but please address all comments

integrations/langchain/src/databricks_langchain/chat_models.py
fanzeyi: oh remember when I asked you if we don't support streaming in openai. that's for this. I should remove it now.

integrations/langchain/src/databricks_langchain/chat_models.py
fanzeyi: The limitation here is Python type checker couldn't see through `@cached_property`:

https://play.ty.dev/7c8f223b-d6d0-4514-a816-2da2b195ac10
https://mypy-play.net/?mypy=latest&python=3.12&gist=b24cddd715787ec9afbc47054d29c49c

:grrr:

`cast` should be safe and noop here.

integrations/langchain/src/databricks_langchain/chat_models.py
fanzeyi: This follows langchain-core type hints:

https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/language_models/chat_models.py#L1521-L1527

though specifically for this, I think ty has improved its support for type refinement / covariance and it's able to accept `Type[BaseModel]` here as the overriden signature.

integrations/langchain/src/databricks_langchain/chat_models.py
fanzeyi: ditto above :P

integrations/langchain/src/databricks_langchain/genie.py
fanzeyi: yeah, I couldn't decide what to do with this. Is it used anywhere? I think I can still `setattr` it (or just make ty ignore this).

integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
fanzeyi: Yes. `StreamableHttpConnection` is `TypedDict`. Effectively `dict` but with better type hints for each attribute.

integrations/langchain/src/databricks_langchain/multi_server_mcp_client.py
fanzeyi: This is because `StreamableHttpConnection` expects a `timedelta` for `timeout`

https://github.com/langchain-ai/langchain-mcp-adapters/blob/f15d1b63356896bc98ccd3757152069038bcb980/langchain_mcp_adapters/sessions.py#L154-L155

Though this is backward incompatible. I will convert it in `to_connection_dict`.

integrations/langchain/src/databricks_langchain/vectorstores.py
fanzeyi: We were not doing the correct overriding here. Upstream `langchain-core` has signature:

```python
    def max_marginal_relevance_search_by_vector(
        self,
        embedding: list[float],
        k: int = 4,
        fetch_k: int = 20,
        lambda_mult: float = 0.5,
        **kwargs: Any,
    ) -> list[Document]:
```

https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/vectorstores/base.py#L723-L730

and not only we added new parameters but we also did not respect the parameter order (`query`).

Therefore it is an invalid method override, but fixing this means we will break existing code. :/

integrations/langchain/src/databricks_langchain/vector_search_retriever_tool.py
fanzeyi: oh ok. I will change this to a suppression and you can remove the directive in your PR.

PR #251, opened 2025-12-23
typing: fixing type errors in databricks_dspy

fanzeyi:
[Click Here to Review](https://ghstack.zeyi.fan/databricks/databricks-ai-bridge/pull/251)

```
[dspy] $ uv run ty check
All checks passed!
```

### DIFF
--- a/integrations/dspy/src/databricks_dspy/clients/databricks_lm.py
+++ b/integrations/dspy/src/databricks_dspy/clients/databricks_lm.py
@@ -95,13 +95,16 @@ class DatabricksLM(dspy.LM):
                 "for how to set up the authentication."
             ) from e
 
-        self.create_pt_endpoint = create_pt_endpoint
-        self.pt_entity = pt_entity
+        self.create_pt_endpoint: bool = create_pt_endpoint
+        self.pt_entity: PtServedModel | None = pt_entity
 
         if create_pt_endpoint:
             self.pt_endpoint = self._create_pt_endpoint()
 
     def _create_pt_endpoint(self):
+        if self.pt_entity is None:
+            raise ValueError("pt_entity must be provided when create_pt_endpoint is True")
+
         # Create the provisioned throughput endpoint configuration
         config = PtEndpointCoreConfig(served_entities=[self.pt_entity])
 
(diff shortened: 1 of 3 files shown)


integrations/dspy/src/databricks_dspy/clients/databricks_lm.py
```diff
@@ -103,7 +105,7 @@ def __init__(
 
     def _create_pt_endpoint(self):
         # Create the provisioned throughput endpoint configuration
-        config = PtEndpointCoreConfig(served_entities=[self.pt_entity])
+        config = PtEndpointCoreConfig(served_entities=[self.pt_entity] if self.pt_entity else [])
```
bbqiu: should this throw an error if the pt_entity field isn't filled in?

doesn't make sense to try and create this pt endpoint with an empty array right

fanzeyi: While I think it doesn't make a lot sense but that would be a pretty big API behavior change

https://github.com/fanzeyi/databricks-ai-bridge/blob/61e2511371bc3bfe7cafb7f10f42947073c0cb03/integrations/dspy/src/databricks_dspy/clients/databricks_lm.py#L74

we do allow constructing `DatabricksLM` without setting `pt_entity` and we were incorrectly passing `[None]` to `PtEndpointCoreConfig` in the past. I think skipping endpoint creation is a bigger behavior change than passing an empty array.

fanzeyi: oh I guess it is already raising an error now:

```
File ~/GitHub/databricks-ai-bridge/integrations/dspy/src/databricks_dspy/clients/databricks_lm.py:117, in DatabricksLM._create_pt_endpoint(self)
    112     return w.serving_endpoints.create_provisioned_throughput_endpoint_and_wait(
    113         name=model_name_without_databricks_prefix,
    114         config=config,
    115     )
    116 except Exception as e:
--> 117     raise RuntimeError(
    118         f"Failed to create provisioned throughput endpoint: {e}\n\n"
    119         "`create_pt_endpoint=True` is only supported in Databricks notebooks now."
    120     ) from e

RuntimeError: Failed to create provisioned throughput endpoint: None
```

PR #255, opened 2025-12-26
Add Async Stateful Classes for Langgraph memory: AsyncLakebasePool, AsyncCheckpointSaver, AsyncDatabricksStore

jennsun:
This PR adds async versions of the Lakebase connection pool (AsyncLakebasePool), LangGraph checkpoint saver (AsyncCheckpointSaver) and LangGraph PostgresStore(AsyncDatabricksStore) so they can be used in async workflows such as async agent servers and streaming applications

Changes
**Core Library (databricks-ai-bridge)**
src/databricks_ai_bridge/lakebase.py
Added AsyncLakebasePool class that wraps [psycopg_pool.AsyncConnectionPool](https://www.psycopg.org/psycopg3/docs/api/pool.html#psycopg_pool.AsyncConnectionPool)

**LangChain (databricks-langchain)**
**integrations/langchain/src/databricks_langchain/checkpoint.py**
Added AsyncCheckpointSaver class extending [langgraph.checkpoint.postgres.aio.AsyncPostgresSaver](https://pypi.org/project/langgraph-checkpoint-postgres/)

**integrations/langchain/src/databricks_langchain/store.py**
Added AsyncDatabricksStore class extending [AsyncBatchedBaseStore](https://reference.langchain.com/python/langgraph/store/#langgraph.store.postgres.AsyncPostgresStore)
Uses AsyncLakebasePool for connection pooling
Creates short-lived AsyncPostgresStore instances for each operation

Added unit tests for new classes 

Used in agent on apps langgraph examples:
short-term: https://github.com/bbqiu/agent-on-app-proto/pull/39
long-term: https://github.com/bbqiu/agent-on-app-proto/pull/42

### DIFF
--- a/src/databricks_ai_bridge/lakebase.py
+++ b/src/databricks_ai_bridge/lakebase.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import logging
 import time
 import uuid
@@ -73,26 +74,14 @@ class _LakebasePoolBase:
         self.host = resolved_host
         self.username = self._infer_username()
 
-        # Token caching (minting is sync via WorkspaceClient)
-        self._cache_lock = Lock()
         self._cached_token: Optional[str] = None
         self._cache_ts: Optional[float] = None
 
-    def _get_token(self) -> str:
-        """Get cached token or mint a new one if expired."""
-        with self._cache_lock:
-            now = time.time()
-            if (
-                self._cached_token
-                and self._cache_ts
-                and (now - self._cache_ts) < self.token_cache_duration_seconds
-            ):
-                return self._cached_token
-
-            token = self._mint_token()
-            self._cached_token = token
-            self._cache_ts = now
-            return token
+    def _is_token_valid(self) -> bool:
+        """Check if the cached token is still valid."""
+        if not self._cached_token or not self._cache_ts:
+            return False
+        return (time.time() - self._cache_ts) < self.token_cache_duration_seconds
 
     def _mint_token(self) -> str:
         try:
@@ -145,6 +134,10 @@ class LakebasePool(_LakebasePoolBase):
             workspace_client=workspace_client,
             token_cache_duration_seconds=token_cache_duration_seconds,
         )
+
+        # Sync lock for thread-safe token caching
+        self._cache_lock = Lock()
+
         # Create connection pool that fetches a rotating M2M OAuth token
         # https://docs.databricks.com/aws/en/oltp/instances/query/notebook#psycopg3
         pool = self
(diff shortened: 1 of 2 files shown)


src/databricks_ai_bridge/lakebase.py
```diff
@@ -180,3 +180,148 @@ def connection(self):
...
+            )
+
+        self.workspace_client = workspace_client
+        self.instance_name = instance_name
+        self.host = resolved_host
+        self.username = _infer_username(workspace_client)
+        self.token_cache_duration_seconds = token_cache_duration_seconds
+
+        # Token caching (use asyncio lock for async context)
+        import asyncio
```
bbqiu: is there a reason we don't want to do a top level import?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -180,3 +180,148 @@ def connection(self):
...
+
+        self._cache_lock = asyncio.Lock()
```
bbqiu: do we mean to use this cache lock somewhere? in the class's current form, i think there's potential for the cached token to get overwritten, which isn't a big deal

src/databricks_ai_bridge/lakebase.py
```diff
@@ -180,3 +180,148 @@ def connection(self):
     def close(self) -> None:
         """Close the connection pool."""
         self._pool.close()
+
+
+class AsyncLakebasePool:
```
bbqiu: high level, can we share some more code with the lakebase pool class? maybe we can extract to a base class that both inherit from

src/databricks_ai_bridge/lakebase.py
```diff
@@ -32,25 +32,12 @@
...
-        )
-
-    user = w.current_user.me()
-    return user.user_name
-
-
-class LakebasePool:
-    """Wrapper around a psycopg connection pool with rotating Lakehouse credentials.
+class _LakebasePoolBase:
+    """
```
jennsun, line 37: created new base class we will use to subclass for LakebasePool (sync ver) and AsyncLakebasePool (async ver)

src/databricks_ai_bridge/lakebase.py
```diff
@@ -83,61 +70,14 @@ def __init__(
...
-        self.workspace_client = workspace_client
-        self.instance_name = instance_name
         self.host = resolved_host
-        self.username = _infer_username(workspace_client)
-        self.token_cache_duration_seconds = token_cache_duration_seconds
+        self.username = self._infer_username()
 
-        # Token caching
+        # Token caching (minting is sync via WorkspaceClient)
         self._cache_lock = Lock()
```
bbqiu: would it be possible to have diff locks and `get_token` methods for async / sync? i'm not 100% sure of the performance implications since we shouldn't consistently acquire new connections, but using a blocking lock each time for a get_token call still unnecessarily introduces sync code here, esp if it's a cache hit

PR #256, opened 2025-12-26
Genie tool uses MCP Adapters under the hood

nisha2003:
Migrates the Genie class to use MCP (Model Context Protocol) adapters, replacing direct REST API calls in `poll_for_result` and `ask_question`.

Updated unit tests.

Manual test here: https://e2-spog.staging.cloud.databricks.com/editor/notebooks/2260046326096166?o=6051921418418893

### DIFF
--- a/integrations/langchain/tests/unit_tests/test_genie.py
+++ b/integrations/langchain/tests/unit_tests/test_genie.py
@@ -105,9 +105,7 @@ def test_query_genie_as_agent(MockWorkspaceClient, MockMCPClient):
         assert not any(m.name == "suggested_questions" for m in result["messages"])
 
         # Test with include_suggested_questions=True
-        result = _query_genie_as_agent(
-            input_data, genie, "Genie", include_suggested_questions=True
-        )
+        result = _query_genie_as_agent(input_data, genie, "Genie", include_suggested_questions=True)
         assert any(m.name == "suggested_questions" for m in result["messages"])
         sq_msg = [m for m in result["messages"] if m.name == "suggested_questions"][0]
         assert sq_msg.content == "What about tomorrow?\n\nShow by city"
@@ -307,9 +305,7 @@ def test_create_genie_agent_with_include_suggested_questions(MockWorkspaceClient
         conversation_id="conv-789",
     )
 
-    agent = GenieAgent(
-        "space-id", "Genie", include_suggested_questions=True, client=mock_client
-    )
+    agent = GenieAgent("space-id", "Genie", include_suggested_questions=True, client=mock_client)
     assert agent.description == "description"
 
     input_data = {"messages": [{"role": "user", "content": "What is the weather?"}]}


src/databricks_ai_bridge/genie.py
```diff
@@ -159,12 +164,60 @@ def _end_current_span(client, parent_trace_id, current_span, final_state, error=
             span_id=current_span.span_id,
             attributes=attributes,
         )
-    except mlflow.exceptions.MlflowTracingException as e:
+    except Exception as e:
```
bbqiu: ooc why did we make this less generic? were you running into an error that wasn't an MLflow tracing exception

src/databricks_ai_bridge/genie.py
```diff
@@ -51,7 +54,10 @@ def _parse_query_result(
 
     for item in output["data_array"]:
         row = []
-        for column, value in zip(columns, item):
+        values = item["values"]
+        for column, value_obj in zip(columns, values):
+            value = value_obj.get("string_value") if isinstance(value_obj, dict) else value_obj
```
bbqiu, line 102: is this change because the mcp server currently returns just a string instead of a structured output? can we add a comment describing this behavior

src/databricks_ai_bridge/genie.py
```diff
@@ -159,12 +164,60 @@ def _end_current_span(client, parent_trace_id, current_span, final_state, error=
...
             attributes=attributes,
         )
-    except mlflow.exceptions.MlflowTracingException as e:
+    except Exception as e:
         logging.warning(f"Failed to end span for {final_state}: {e}")
 
     return None
 
 
+def _parse_genie_mcp_response(
```
bbqiu, line 216: could we add some documentation here to:
1. lay out the different formats that the genie MCP server can return (and when they use each one)
2. what our alg is to parse each format (ordering of tries, fallback definitions etc)

src/databricks_ai_bridge/genie.py
```diff
@@ -177,6 +230,20 @@ def __init__(
...
         self.description = self.genie.get_space(space_id).description
+
+        server_url = f"{workspace_client.config.host}/api/2.0/mcp/genie/{space_id}"
+        self._mcp_client = DatabricksMCPClient(server_url, workspace_client)
+
+        tools = self._mcp_client.list_tools()
+        if not tools:
+            raise ValueError(f"No tools found in Genie MCP server for space {space_id}")
+
+        query_tools = [tool for tool in tools if "query" in tool.name.lower()]
```
bbqiu: can we do something like this instead since we know it's only 1 tool?
```
# Use next() with a generator:
  self._query_tool_name = next(
      (tool.name for tool in tools if tool.name.startswith("query_space_")),
      None
  )
  self._poll_tool_name = next(
      (tool.name for tool in tools if tool.name.startswith("poll_response_")),
      None
  )
```

src/databricks_ai_bridge/genie.py
```diff
@@ -1,4 +1,5 @@
 import bisect
```
bbqiu, line 1: can we delete unused methods like start_conversation and create_message?

pyproject.toml
```diff
@@ -103,3 +104,6 @@ filterwarnings = [
     "default::Warning:databricks_ai_bridge",
     "default::Warning:tests",
 ]
+
+[tool.uv.sources]
+databricks-mcp = { path = "databricks_mcp/", editable = true }
```
bbqiu, line 109: what is this for? i think for dev, we can manually do local installs of databricks-mcp instead of this

~~does this always try to install from local instead of pypi?~~ looks like it's only for local work, and on pypi, this won't apply -- if that's right, we should do this for all of our subpackage definitions to rely on the local installs when possible for dev

src/databricks_ai_bridge/genie.py
nisha2003: I put a deprecation message since I'm thinking there's maybe some usage of these methods right now in custom agents?

pyproject.toml
nisha2003: added for the other subpackages!

src/databricks_ai_bridge/genie.py
```diff
@@ -165,6 +175,104 @@ def _end_current_span(client, parent_trace_id, current_span, final_state, error=
...
+    msg_id = genie_response.get("messageId", message_id)
+    query_str = ""
+    description = ""
+
+    try:
+        query_attachments = content.get("queryAttachments", [])
+        text_attachments = content.get("textAttachments", [])
+
+        if query_attachments:
+            first_query = query_attachments[0]
```
aravind-segu: just curious here, do we always return only one query attachment?

src/databricks_ai_bridge/genie.py
```diff
@@ -165,6 +175,104 @@ def _end_current_span(client, parent_trace_id, current_span, final_state, error=
...
+            query_str = first_query.get("query", "")
```
aravind-segu, line 283: we are adding the query_str here even if the statement response failed. Is that ok? Is that what we do currently?

src/databricks_ai_bridge/genie.py
```diff
@@ -206,154 +349,148 @@ def create_message(self, conversation_id, content):
 
     @mlflow.trace()
     def poll_for_result(self, conversation_id, message_id):
```
aravind-segu: If we are keeping start_conversation/create_message and just adding a deprecated, should we keep original poll_for_result, mark as deprecated and then create a new poll for ask_question

src/databricks_ai_bridge/genie.py
```diff
@@ -165,6 +175,104 @@ def _end_current_span(client, parent_trace_id, current_span, final_state, error=
...
+            first_query = query_attachments[0]
+            query_str = first_query.get("query", "")
+            description = first_query.get("description", "")
+            statement_response = first_query.get("statement_response")
+
+            if (
+                statement_response
+                and statement_response.get("status", {}).get("state") == "SUCCEEDED"
+            ):
+                result = _parse_query_result(statement_response, truncate_results, return_pandas)
```
aravind-segu, line 291: Image

There are some cases where we return both query_attachments and text_attachments. Previously we were returning all attachments to the user. Now we will only be returnign query_attachements. Is this a conscious decision?

If not maybe its better to also return text_attachments here if query_attachments exist. Text attachments are usually what genie summarizes and basically is an answer to the user.

src/databricks_ai_bridge/genie.py
nisha2003: I think the old poll_for_result is used by ask_question. I guess it's internally used but the pattern of using start_conversation --> create_message --> poll_for_result is deprecated. Since on the user side it shouldn't be used I can mark poll_for_result to deprecated as well.

I feel like we shouldn't need a new poll for ask_question right? Let me know if I'm missing smth.

src/databricks_ai_bridge/genie.py
nisha2003: Yep, I think the old code did this (snippet below of the old poll_query_results). I imagine this is so that even if a SQL query fails there's some use in sending it back for debugging?

```
if state == "SUCCEEDED":
    result = _parse_query_result(resp, self.truncate_results, self.return_pandas)
    return GenieResponse(result, query_str, description, returned_conversation_id)
elif state in ["RUNNING", "PENDING"]:
    logging.debug("Waiting for query result...")
    time.sleep(ITERATION_FREQUENCY)
else:
     return GenieResponse(
        f"No query result: {resp['state']}",
        query_str,
        description,
        returned_conversation_id,
     )
```

src/databricks_ai_bridge/genie.py
nisha2003: Hm okay -- I did this because the old code had the following logic. It's also I took only one query attachment. If we want to change this to send all the attachments we probably need to update GenieResponse. Is this okay? I imagine it is since we'll just feed it to the agent + will have more information but let me know what you think.

`if current_status == "COMPLETED":
                            attachment = next(
                                (r for r in resp["attachments"] if "query" in r), None
                            )
                            if attachment:
                                query_obj = attachment["query"]
                                description = query_obj.get("description", "")
                                query_str = query_obj.get("query", "")
                                attachment_id = attachment["attachment_id"]
                                return poll_query_results(
                                    attachment_id,
                                    query_str,
                                    description,
                                    returned_conversation_id,
                                )
                            if current_status == "COMPLETED":
                                text_content = next(r for r in resp["attachments"] if "text" in r)[
                                    "text"
                                ]["content"]
                                return GenieResponse(
                                    result=text_content,
                                    conversation_id=returned_conversation_id,
                                )`

src/databricks_ai_bridge/genie.py
nisha2003: Addressed in comment below.

src/databricks_ai_bridge/genie.py
nisha2003: Added an internal poller that is called by `ask_question` and `poll_for_result`.

PR #259, opened 2026-01-02
OpenAI VS tool uses MCP Adapters under the hood

nisha2003:
Migrate OpenAI Vector Search Toolkit to use MCP adapters. We still preserve the direct API path for self-managed embeddings (for which there is no MCP support).

Other minor changes:
- McpServerToolkit sends meta parameters to the backend
- Filter parameters accepts both List[FilterItem] and Dict format

Updated unit tests to parametrize MCP and Direct API path.

Manual tests in this notebook: https://eng-ml-inference.staging.cloud.databricks.com/editor/notebooks/1465545330011655?o=1653573648247579

### DIFF
--- a/integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
+++ b/integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
@@ -232,7 +232,7 @@ class VectorSearchRetrieverTool(VectorSearchRetrieverToolMixin):
         else:
             return self._execute_direct_api_path(query, filters, openai_client, **kwargs)
 
-    def _get_mcp_toolkit(self) -> Callable:
+    def _create_or_get_mcp_toolkit(self) -> Callable:
         """
         If it does not exist, create the MCP tool execution function for this index.
         Otherwise, return the execution function.
@@ -289,7 +289,7 @@ class VectorSearchRetrieverTool(VectorSearchRetrieverToolMixin):
             return {}
         if isinstance(filters, dict):
             return filters
-        return {dict(item)["key"]: dict(item)["value"] for item in filters}
+        return {item.model_dump()["key"]: item.model_dump()["value"] for item in filters}
 
     def _build_mcp_meta(
         self, filters: Optional[Union[Dict[str, Any], List[FilterItem]]] = None, **kwargs: Any
@@ -311,7 +311,10 @@ class VectorSearchRetrieverTool(VectorSearchRetrieverToolMixin):
 
         combined_filters = {**self._normalize_filters(filters), **(self.filters or {})}
         if combined_filters:
-            meta["filters"] = json.dumps(combined_filters)
+            try:
+                meta["filters"] = json.dumps(combined_filters)
+            except (TypeError, ValueError) as e:
+                raise ValueError(f"Filters must be JSON serializable: {e}") from e
 
         if "score_threshold" in kwargs:
             meta["score_threshold"] = float(kwargs.pop("score_threshold"))
@@ -443,7 +446,7 @@ class VectorSearchRetrieverTool(VectorSearchRetrieverToolMixin):
         **kwargs: Any,
     ) -> List[Dict]:
         try:
-            mcp_execute = self._get_mcp_toolkit()
+            mcp_execute = self._create_or_get_mcp_toolkit()
             meta = self._build_mcp_meta(filters, **kwargs)
             mcp_response = mcp_execute(query=query, _meta=meta)
             documents = self._parse_mcp_response(mcp_response)


databricks_mcp/src/databricks_mcp/mcp.py
```diff
@@ -212,18 +213,32 @@ def list_tools(self) -> List[Tool]:
...
             arguments (dict[str, Any], optional): The arguments to pass to the tool. Defaults to None.
+            meta (dict[str, Any], optional): Optional metadata parameters for additional tool configurations.
+                For Vector Search tools, this can include:
+                - num_results: Number of results to return
+                - filters: JSON string of filter conditions
+                - query_type: Type of query (e.g., "ANN", "HYBRID")
+                - columns: Comma-separated list of columns to return
+                - score_threshold: Minimum similarity score threshold
+                - include_score: Whether to include similarity scores in results
+                - columns_to_rerank: Columns to use for reranking
```
sunishsheth2009: Since this is generic, I don't think we should include this doc here.

This is what the Python docs in MCP have:
https://github.com/modelcontextprotocol/python-sdk/blob/6b69f6354ad01e08abc8e1b81390c3a45132f594/src/mcp/client/experimental/tasks.py#L70

https://github.com/modelcontextprotocol/python-sdk/blob/6b69f6354ad01e08abc8e1b81390c3a45132f594/src/mcp/client/session.py#L375

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -195,52 +200,185 @@ def _validate_tool_inputs(self):
     def execute(
         self,
         query: str,
-        filters: Optional[List[FilterItem]] = None,
+        filters: Optional[Union[Dict[str, Any], List[FilterItem]]] = None,
```
sunishsheth2009, line 205: Why do the filters need to take in both now? Can it continue working with `FilterItem` as before?

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -195,52 +200,185 @@ def _validate_tool_inputs(self):
...
         openai_client: OpenAI = None,
         **kwargs: Any,
     ) -> List[Dict]:
         """
         Execute the VectorSearchIndex tool calls from the ChatCompletions response that correspond to the
```
sunishsheth2009, line 210: can we add example usage?

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
nisha2003: Oh, I added Dict support since that's what the MCP meta filter field expects - makes it easier to pass filters directly without wrapping them in FilterItems. Both formats get normalized to dict internally.

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
nisha2003: Added meta parameters + fixed the class docstring.

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -67,15 +73,14 @@ class VectorSearchRetrieverTool(VectorSearchRetrieverToolMixin):
...
-    :class:`~databricks_ai_bridge.vector_search_retriever_tool.VectorSearchRetrieverToolMixin` for additional supported constructor
-    arguments not listed below, including `query_type` and `num_results`.
+    **Note**: Any additional keyword arguments passed to the constructor will be passed along when executing the tool.
+    The ``execute()`` method supports meta parameters such as ``num_results``, ``score_threshold``, ``query_type``,
+    ``filters``, ``columns``, and ``columns_to_rerank``. See
+    :class:`~databricks_ai_bridge.vector_search_retriever_tool.VectorSearchRetrieverToolMixin` for additional supported
+    constructor arguments.
 
-    WorkspaceClient instances with auth types PAT, OAuth-M2M (client ID and client secret), or model serving credential strategy will be used to instantiate the underlying VectorSearchClient.
+    WorkspaceClient instances with auth types PAT, OAuth-M2M (client ID and client secret), or model serving credential
```
bbqiu, line 82: will this also work for personal OAuth tokens? i believe so, but i'm not 100% sure if we have added the code to pass it along properly

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
bbqiu: sorry if i'm missing something, but would it be possible to add an actual example? ex. when creating the vs tool, including in the docstring an example where we initialize a vs retriever tool with dict filters

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -195,52 +202,185 @@ def _validate_tool_inputs(self):
...
-            oai_client = openai_client or OpenAI()
-            if not oai_client.api_key:
-                raise ValueError(
-                    "OpenAI API key is required to generate embeddings for retrieval queries."
-                )
+            return self._execute_mcp_path(query, filters, **kwargs)
+        else:
+            return self._execute_direct_api_path(query, filters, openai_client, **kwargs)
+
+    def _get_mcp_toolkit(self) -> Callable:
```
bbqiu: nit: can we add a docstring describing the get/create behavior here

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -195,52 +202,185 @@ def _validate_tool_inputs(self):
...
+        filters: Optional[Union[Dict[str, Any], List[FilterItem]]] = None,
         openai_client: OpenAI = None,
         **kwargs: Any,
     ) -> List[Dict]:
         """
         Execute the VectorSearchIndex tool calls from the ChatCompletions response that correspond to the
         self.tool VectorSearchRetrieverToolInput and attach the retrieved documents into tool call messages.
 
+        Automatically routes to MCP path (Databricks-managed embeddings) or
+        Direct API path (self-managed embeddings) based on index configuration.
```
bbqiu: suggestion from claude:
```
Execute vector search with automatic routing:
  - MCP path: Used for Databricks-managed embeddings (no embedding model configuration needed)
  - Direct API path: Used for self-managed embeddings (requires openai_client)
```

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -195,52 +202,185 @@ def _validate_tool_inputs(self):
...
+            )
+        except Exception as e:
+            raise RuntimeError(
+                f"Failed to initialize MCP toolkit for index {self.index_name}. "
+                f"Ensure the index exists and is configured for Databricks-managed embeddings. "
+                f"Error: {e}"
+            ) from e
+
+        tools = self._mcp_toolkit.get_tools()
+        if len(tools) != 1:
```
annzhang-db: Curious if we need this check, although currently the Vector Search Managed MCP does only provide one tool per index and I think it's unlikely we change that, it does mean that potentially, we update the managed server and break everyone using older versions of databricks-ai-bridge. Maybe we can modify this to be if `len(tools) < 1` to be more safe.

annzhang-db: On second thought this is likely fine, we should probably consider it a breaking change if we modify the vector search managed MCP to that extent.

bbqiu: i like <1 tool, it's basically a check whether or not the MCP server worked

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
bbqiu: sorry, should've said this last time, but can we also rename this method to be get or create for clarity?

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -195,52 +202,193 @@ def _validate_tool_inputs(self):
...
+            filters: Either a dict or List[FilterItem]
+
+        Returns:
+            Dict of filter key-value pairs
+        """
+        if filters is None:
+            return {}
+        if isinstance(filters, dict):
+            return filters
+        return {dict(item)["key"]: dict(item)["value"] for item in filters}
```
bbqiu: would model_dump work here? to use a native pydantic method instead

we can try / catch for non-FilterItem objects that make it to this function if that's possible

integrations/openai/src/databricks_openai/vector_search_retriever_tool.py
```diff
@@ -195,52 +202,193 @@ def _validate_tool_inputs(self):
...
+            query_type = kwargs.pop("query_type", self.query_type)
+            if query_type:
+                meta["query_type"] = query_type
+
+        if self.columns:
+            meta["columns"] = ",".join(self.columns)
+
+        combined_filters = {**self._normalize_filters(filters), **(self.filters or {})}
+        if combined_filters:
+            meta["filters"] = json.dumps(combined_filters)
```
bbqiu: are combined filters guaranteed to be json dumpable? can we try/catch and rethrow a nice error?

PR #263, opened 2026-01-06
Databricks MCP: throw a better error for incompatible workspace clients

bbqiu:
- for PAT ws clients when connecting to db apps
- for workspace clients with SPs that don't have API access
  - original report: https://databricks.slack.com/archives/C065NC65Q9F/p1767708726230909

- tested with unit tests + manually

### DIFF
--- a/databricks_mcp/tests/unit_tests/test_mcp.py
+++ b/databricks_mcp/tests/unit_tests/test_mcp.py
@@ -469,81 +469,35 @@ class TestIsDatabricksAppsUrl:
 
 
 class TestIsOauthAuth:
-    """Test cases for _is_oauth_auth helper function."""
-
-    def test_is_oauth_auth_returns_true_when_oauth_token_succeeds(self):
+    @pytest.mark.parametrize(
+        "side_effect,expected",
+        [
+            (None, True),  # oauth_token succeeds
+            (ValueError("not available"), False),  # oauth_token raises
+        ],
+    )
+    def test_is_oauth_auth(self, side_effect, expected):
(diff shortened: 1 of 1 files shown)


databricks_mcp/tests/integration_tests/test_mcp_client.py
```diff
@@ -0,0 +1,37 @@
+from databricks.sdk import WorkspaceClient
+
+from databricks_mcp import DatabricksMCPClient
+
+# os.environ["DATABRICKS_CONFIG_PROFILE"] = "dogfood"
```
aravind-segu: nit: remove comment

databricks_mcp/src/databricks_mcp/mcp.py
```diff
@@ -23,6 +23,18 @@
...
+
+def _is_databricks_apps_url(url: str) -> bool:
+    """Check if the URL is hosted on Databricks Apps."""
+    parsed = urlparse(url)
+    return parsed.netloc.endswith(".databricksapps.com")
+
+
+def _is_pat_auth(workspace_client: WorkspaceClient) -> bool:
+    """Check if the workspace client is using PAT authentication."""
+    return workspace_client.config.auth_type == "pat"
```
aravind-segu: Not really a resilient way to check for pat token type. For Example in notebooks, the auth_type would be `runtime`. 

I think we should either call the oauthToken method copy this implementation here to check if its not an OAuth token instead: 

https://github.com/databricks/databricks-sdk-py/blob/fd2932e669dda96a07bda1f554825d9aea6e8187/databricks/sdk/config.py#L282

PR #269, opened 2026-01-11
databricks-openai: update openai clients to strip the "strict" field for non GPT models

bbqiu:
reported here: https://github.com/databricks-eng/universe/pull/1470397#issuecomment-3720603871

for non-gpt models, in openai-agents-sdk > 0.6.3, they introduced adding the `strict` field when converting to chat completions format, resulting in: 
```
Error code: 400 - {'error_code': 'BAD_REQUEST', 'message': '{"message":"tools.0.custom.strict: Extra inputs are not permitted"}'}
```

PR to strip the `strict` field in our openai clients

tested manually + unit tests

with this PR: 
screenshot 2026-01-12-16-45-46-Arc

without this PR:
screenshot 2026-01-12-16-46-28-Arc


aravind-segu (approved):
LGTM

PR #273, opened 2026-01-13
CI: Add GitHub Action for linting, formatting and typechecking

fanzeyi:
[Click Here to Review](https://ghstack.zeyi.fan/databricks/databricks-ai-bridge/pull/273)

As demonstrated in this PR:

image

https://github.com/databricks/databricks-ai-bridge/pull/273/changes#diff-f2bb640113bd724e62965bf48aaf16e342e05ee0d840eef176dcd883529a9258R117

### DIFF
--- a/.github/workflows/ty.yml
+++ b/.github/workflows/ty.yml
@@ -1,0 +1,89 @@
+name: Linting
+
+on:
+  pull_request:
+    branches:
+      - main
+
+env:
+  UV_VERSION: "0.9.11"
+  PYTHON_VERSION: "3.12"
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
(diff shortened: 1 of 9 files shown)


.github/workflows/ty.yml
```diff
@@ -0,0 +1,89 @@
...
+
+  lint:
+    name: ruff check for ${{ matrix.project.dir }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        project:
+          - dir: "."
+            ty_flags: "--extra memory"
```
bbqiu, line 38: is there a way for us to add these ty flags automatically? ex. if we add a memory dependency for openai (which is happening soon), could we automatically add here?

if too much effort, it's fine, just include a note at the top about how to change this file

bbqiu (approved):
lgtm

.github/workflows/ty.yml
fanzeyi: @bbqiu  good point. uv has `--all-extras` flag and we can use that.

PR #274, opened 2026-01-15
OpenAI SDK Add MemorySession that follows the Session protocol

jennsun:
MemorySession aligning with OpenAI Agents SDK target framework: https://docs.google.com/document/d/1gX1XEUdCUvumv-Buks-OcL69DqH3I0P34SVPwgc34Cg/edit?tab=t.0#heading=h.71k2cv1krxyr

Custom Session Implementation for short-term / long term memory on Databricks with connection pooling/token caching and rotation handled with LakebasePool capabilities

You can implement your own session memory by creating a class that follows the [Session](https://openai.github.io/openai-agents-python/ref/memory/session/#agents.memory.session.Session) protocol:
https://openai.github.io/openai-agents-python/sessions/#custom-session-implementations

More on Session protocol:
https://openai.github.io/openai-agents-python/ref/memory/session/#agents.memory.session.Session

Before:
```
from agents.extensions.memory import SQLAlchemySession
from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine

def get_engine() -> AsyncEngine:
    """
    Get or create the async database engine.
    
    For Lakebase, this will check if the token is still valid and
    recreate the engine if the token has expired.
    """
    global _engine, _cache_ts

    # Check if we need to recreate the engine due to token expiration
    if _engine is not None and _instance_name is not None:
        with _cache_lock:
            now = time.time()
            if _cache_ts and (now - _cache_ts) >= DEFAULT_TOKEN_CACHE_DURATION_SECONDS:
                # Token expired, dispose old engine
                logger.info("Lakebase token expired, recreating engine")
                _engine.sync_engine.dispose()
                _engine = None

    if _engine is None:
        _engine = create_async_engine(
            get_database_url(),
            echo=False,  # Set to True for SQL debugging
            pool_pre_ping=True,
        )
    return _engine

    engine = get_engine()
    session = SQLAlchemySession(
        session_id=resolved_thread_id,
        engine=engine,
        create_tables=create_tables,
    )
   result = await Runner.run(agent, user_message, session=session)
```

By using LakebasePool in our custom Session class implementation, we do not need to create a [SQLAlchemy engine from the OpenAI Agents SDK example](https://openai.github.io/openai-agents-python/sessions/#sqlalchemy-sessions) and can leverage the token caching/connection pooling logic Lakebase Pool offers already. Our Session class will have a cached pool associated with each Session instance so there aren't new pools created per session/user does not have to manage the pool

```
from databricks_openai.agents.session import LakebaseSession

session = MemorySession(
        session_id=thread_id,
        instance_name=LAKEBASE_INSTANCE_NAME,
    )

result = await Runner.run(agent, user_message, session=session)
```

local testing + app testing

Notebook with smoke tests, CRUD operations tests, and integration tests:
https://eng-ml-agent-platform.staging.cloud.databricks.com/editor/notebooks/1816672511960993?o=2850744067564480#command/8078768795088187

App: https://eng-ml-agent-platform.staging.cloud.databricks.com/apps/j-openai-stateful?o=2850744067564480

Ran server + tested locally

image

test app:
image

image

thread viewable in lakebase:
image

used in openai stateful agent example: https://github.com/databricks/app-templates/pull/90

Unit tests:
`python -m pytest integrations/openai/tests/unit_tests/test_session.py -v`


integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,647 @@
...
+from threading import Lock
+from typing import Any, Dict, Optional, Tuple, Union, cast
+from uuid import UUID
+
+from databricks.sdk import WorkspaceClient
+
+try:
+    from agents.items import TResponseInputItem
+    from agents.memory.session import SessionABC
+    from databricks_ai_bridge.lakebase import AsyncLakebasePool, LakebasePool
```
jennsun, line 16: note: dependency for ai-bridge needs to depend on next release in which asynclakebasepool is available

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -18,14 +18,15 @@ def auth_flow(self, request: Request) -> Generator[Request, Response, None]:
...
 
     Databricks model endpoints (except GPT) don't support the 'strict' field
     in tool schemas, but openai-agents SDK v0.6.4+ includes it.
     """
-    if tools is None:
-        return None
+    # Handle None or OpenAI's Omit sentinel (non-iterable placeholder)
+    if tools is None or not isinstance(tools, list):
+        return tools
     for tool in tools:
```
jennsun, line 30: When the OpenAI Agents SDK calls the Databricks client, it may pass an Omit sentinel object (from openai._types) instead of None or a list for optional parameters like tools, causing error 
```
--- First turn ---
TypeError: 'Omit' object is not iterable
File <command-8078768795088187>, line 78
     75     print(f"\n🎉 Agent integration test completed!")
     76     print(f"💡 Check Lakebase tables with session_id = '{test_session_id}'")
---> 78 await test_agent_with_session()
File /local_disk0/.ephemeral_nfs/envs/pythonEnv-58a54a3c-4fd4-4b6e-b9cd-3cc110aea59e/lib/python3.12/site-packages/databricks_openai/utils/clients.py:29, in _strip_strict_from_tools(tools)
     27 if tools is None:
     28     return None
---> 29 for tool in tools:
     30     if isinstance(tool, dict) and "function" in tool:
     31         tool.get("function", {}).pop("strict", None
```

here  I updated function to check if `tools` is a list before iterating through it

src/databricks_ai_bridge/lakebase.py
```diff
@@ -50,34 +50,88 @@ def __init__(
...
+            hostname: The database hostname (e.g., from Databricks Apps valueFrom: "database")
+
+        Returns:
+            Tuple of (instance_name, host)
+
+        Raises:
+            ValueError: If no matching instance is found
+        """
+        try:
+            instances = list(self.workspace_client.database.list_database_instances())
```
bbqiu: hm is this safe to do? is there no better way than to manually list and search through every single instance?

bbqiu: from gemini:
```
2. The "Heuristic" Way (Fastest / Zero API Calls)
In almost all Databricks Lakebase/PostgreSQL implementations, the Instance Name is the prefix of the Hostname.

Format: <instance-name>.<region>.db.databricks.com (or similar)

Example: If hostname is my-db-instance-xyz.us-east-1.db.databricks.com, the instance name is my-db-instance-xyz.

If you can confirm this pattern in your environment (print one hostname and instance_name to check), you can skip the API call entirely:

Python
def _resolve_from_hostname(self, hostname: str) -> tuple[str, str]:
    # Extract the first part of the domain
    potential_name = hostname.split('.')[0]
    
    # Optional: basic validation to ensure it looks like an ID
    if not potential_name:
         raise ValueError(f"Invalid hostname format: {hostname}")
         
    return potential_name, hostname
```

not sure if this is an established, could be helpful to ask lakebase folks or look around for other examples of this resolution

jennsun: unfortunately the host string does not contain the instance name :( let me look into better resolution method here

integrations/openai/tests/integration_tests/test_memory_session.py
```diff
@@ -0,0 +1,413 @@
+"""Integration tests for MemorySession and AsyncMemorySession.
```
jennsun, line 1: integration tests follow notebook examples here: 

Notebook with smoke tests, CRUD operations tests, and integration tests:
https://eng-ml-agent-platform.staging.cloud.databricks.com/editor/notebooks/1816672511960993?o=2850744067564480#command/8078768795088187

src/databricks_ai_bridge/lakebase.py
jennsun: unfortunately, it seems like that feature is not supported in the sdk yet - our best option is listing and searching through all instances (current approach) https://databricks.slack.com/archives/C05J5LYK9BP/p1769560056487269?thread_ts=1769556233.287279&cid=C05J5LYK9BP

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,647 @@
...
+        Args:
+            session_id: UUID identifier for this conversation session.
+            sessions_table: Name of the sessions table. Defaults to "agent_sessions".
+            messages_table: Name of the messages table. Defaults to "agent_messages".
+        """
+        self.session_id = session_id
+        self.sessions_table = sessions_table
+        self.messages_table = messages_table
+
+    # --- SQL Building Helpers ---
```
bbqiu, line 129: ooc, why did we have to rewrite these ourselves? presuming we are using the schema that the sessions API also uses, shouldn't we be able to reuse those queries?

src/databricks_ai_bridge/lakebase.py
bbqiu: ah okay -- can we make this a shared utility then? also if it's returning a generator, maybe we can just iterate through the generator instead of materializing it all in case there are many -- we can maybe even stick this within an  asyncio event loop to make it async?

jennsun: would we prefer to separate the hostname string resolution in lakebasepool to another pr? We can move it here and I can delete from this branch since it is a separate change that can be isolated: https://github.com/databricks/databricks-ai-bridge/pull/310

integrations/openai/src/databricks_openai/agents/session.py
bbqiu: talked offline: https://github.com/openai/openai-agents-python/blob/669a3555ce64e0bc70c26b9b733f2b435010b49d/src/agents/extensions/memory/sqlalchemy_session.py#L53

maybe can subclass this?

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -18,14 +18,15 @@ def auth_flow(self, request: Request) -> Generator[Request, Response, None]:
...
     """Remove 'strict' field from tool function definitions.
 
     Databricks model endpoints (except GPT) don't support the 'strict' field
     in tool schemas, but openai-agents SDK v0.6.4+ includes it.
     """
-    if tools is None:
-        return None
+    # Handle None or OpenAI's Omit sentinel (non-iterable placeholder)
+    if tools is None or not isinstance(tools, list):
+        return tools
```
fanzeyi, line 29:
```suggestion
    if not tools:
        return tools
```

PR #276, opened 2026-01-15
ai-sdk-provider: add GitHub Actions CI and fix dedupe logic

eLysgaard:
I noticed the dedupe logic wasn't always working with KA agents, took a close look at the code and saw the issue. I also added GH workflow to run the tests on every PR

----------------

## Summary

- Add GitHub Actions workflow for ai-sdk-provider with:
  - Tests on Node.js 18, 20, 22
  - Linting, typechecking, and format checking
  - Build verification and package export checks
  - Path filtering to only run on relevant changes

- Fix shouldDedupeOutputItemDone to only consider text-deltas after the last response.output_item.done event, not the entire stream history. This fixes duplicate text appearing when multiple messages are streamed in sequence.

- Add comprehensive unit tests for shouldDedupeOutputItemDone covering edge cases and the new behavior.

### DIFF
--- a/integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
+++ b/integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
@@ -264,7 +264,7 @@ export class DatabricksResponsesAgentLanguageModel implements LanguageModelV2 {
   }
 }
 
-function shouldDedupeOutputItemDone(
+export function shouldDedupeOutputItemDone(
   incomingParts: LanguageModelV2StreamPart[],
   previousParts: LanguageModelV2StreamPart[]
 ): boolean {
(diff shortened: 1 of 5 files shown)


integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -287,26 +287,48 @@ function shouldDedupeOutputItemDone(
...
-    } else {
-      // We've hit a non-text-delta part, stop here
+    if (
+      part.type === 'text-delta' &&
+      part.providerMetadata?.databricks?.itemType === 'response.output_item.done'
+    ) {
+      startIndex = i + 1
       break
     }
   }
```
simonfaltum: Could maybe do 

```
const startIndex = previousParts.findLastIndex(
    (p) =>
      p.type === 'text-delta' &&
      p.providerMetadata?.databricks?.itemType === 'response.output_item.done'
  ) + 1;
```

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -287,26 +287,48 @@ function shouldDedupeOutputItemDone(
...
+      currentText = ''
+    }
   }
+  reconstructuredTexts.push(currentText)
 
-  return doneTextDelta.delta.includes(reconstructedText)
+  // 2. check if the .done text is contained in the reconstructed text and that it follows the same order as the previous parts
+  if (reconstructuredTexts.length === 0) {
+    return false
+  }
```
simonfaltum: Maybe 

```
const reconstructedTexts = previousParts
    .slice(startIndex)
    .reduce<string[]>((acc, part) => {
      if (part.type === 'text-delta') {
        const currentIdx = acc.length - 1;
        acc[currentIdx] = (acc[currentIdx] || '') + part.delta;
      } else if (acc[acc.length - 1] !== '') {
        // Only push a new empty string if the last one was actually used
        acc.push('');
      }
      return acc;
    }, [''])
    .map(t => t.trim())
    .filter(t => t.length > 0);
```

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -287,26 +287,48 @@ function shouldDedupeOutputItemDone(
...
+  let lastIndex = 0
+  for (let i = 0; i < reconstructuredTexts.length; i++) {
+    const indexOfReconstructedText = doneTextDelta.delta.indexOf(reconstructuredTexts[i], lastIndex)
+    if (indexOfReconstructedText === -1) {
+      return false
+    }
+    lastIndex = indexOfReconstructedText + reconstructuredTexts[i].length
+  }
+  return true
```
simonfaltum:
```
let lastIndex = 0;
  return reconstructedTexts.length > 0 && reconstructedTexts.every((text) => {
    const foundIndex = doneTextDelta.delta.indexOf(text, lastIndex);
    if (foundIndex === -1) return false;
    
    lastIndex = foundIndex + text.length;
    return true;
  });
  ```

PR #285, opened 2026-01-20
 ai-sdk-provider: v0.2.0 - Tool calling improvements and API naming updates

eLysgaard:
Summary

  - Adds tool calling support for Responses endpoints
  - Introduces clearer provider method names (chatCompletions, responses)
  - Removes internal XML tag parsing for FMAPI tool calls in favor of standard OpenAI format

  Changes

  Added
  - chatCompletions() provider method as the new name for FM API endpoints
  - responses() provider method as the new name for Responses endpoints
  - Tool calling support for Responses and ChatCompletion endpoints
  - Test coverage for responses-prepare-tools.ts

  Changed
  - FMAPI now uses standard OpenAI tool_calls format instead of XML tag parsing
  - System messages preserve role: 'system' instead of being converted to role: 'user'
  - Tool messages use OpenAI tool_call_id format instead of XML serialization
  - Empty streaming chunks are now discarded to prevent spurious text-start/text-end cycles

  Deprecated
  Renamed the following to avoid confusion on naming:
  - fmapi() - use chatCompletions() instead
  - responsesAgent() - use responses() instead

  Test plan

  - All 287 tests pass
  - Typecheck passes
  - Lint passes
  - Format passes

Manually testing using https://github.com/databricks/databricks-ai-bridge/pull/239 and https://github.com/databricks/app-templates/tree/main/e2e-chatbot-app-next as test bed

### DIFF
--- a/integrations/ai-sdk-provider/src/databricks-provider.ts
+++ b/integrations/ai-sdk-provider/src/databricks-provider.ts
@@ -14,17 +14,9 @@ export type DatabricksLanguageModelConfig = {
 export interface DatabricksProvider extends ProviderV2 {
   /** Agents */
   chatAgent(modelId: string): LanguageModelV2 // agent/v2/chat
-
-  /**
-   * @deprecated Use `responses` instead.
-   */
-  responsesAgent(modelId: string): LanguageModelV2 // agent/v1/responses
   responses(modelId: string): LanguageModelV2 // agent/v1/responses
 
-  /**
-   * @deprecated Use `chatCompletions` instead.
-   */
-  fmapi(modelId: string): LanguageModelV2 // llm/v1/chat
+  /** Foundation Models */
   chatCompletions(modelId: string): LanguageModelV2 // llm/v1/chat
 }
 
@@ -96,9 +88,5 @@ export const createDatabricksProvider = (
     imageModel: notImplemented('ImageModel'),
     textEmbeddingModel: notImplemented('TextEmbeddingModel'),
     languageModel: notImplemented('LanguageModel'),
-    // @deprecated
-    responsesAgent: createResponsesAgent,
-    // @deprecated
-    fmapi: createFmapi,
   }
 }
--- a/integrations/ai-sdk-provider/CHANGELOG.md
+++ b/integrations/ai-sdk-provider/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - FMAPI: System messages now preserve `role: 'system'` instead of being converted to `role: 'user'`
 - FMAPI: Tool messages now use OpenAI `tool_call_id` format instead of XML serialization
 
-### Deprecated
+### Removed
 
 - `fmapi()` provider method - use `chatCompletions()` instead
 - `responsesAgent()` provider method - use `responses()` instead
--- a/integrations/ai-sdk-provider/tests/databricks-provider.test.ts
+++ b/integrations/ai-sdk-provider/tests/databricks-provider.test.ts
@@ -18,8 +18,8 @@ describe('createDatabricksProvider', () => {
 
       expect(provider).toBeDefined()
       expect(typeof provider.chatAgent).toBe('function')
-      expect(typeof provider.responsesAgent).toBe('function')
-      expect(typeof provider.fmapi).toBe('function')
+      expect(typeof provider.responses).toBe('function')
+      expect(typeof provider.chatCompletions).toBe('function')
       expect(typeof provider.imageModel).toBe('function')
       expect(typeof provider.textEmbeddingModel).toBe('function')
       expect(typeof provider.languageModel).toBe('function')
@@ -83,16 +83,16 @@ describe('createDatabricksProvider', () => {
       expect(model.specificationVersion).toBe('v2')
     })
 
-    it('responsesAgent method exists and returns DatabricksResponsesAgentLanguageModel', () => {
-      const model = provider.responsesAgent('test-model')
+    it('responses method exists and returns DatabricksResponsesAgentLanguageModel', () => {
+      const model = provider.responses('test-model')
 
       expect(model).toBeInstanceOf(DatabricksResponsesAgentLanguageModel)
       expect(model.modelId).toBe('test-model')
       expect(model.specificationVersion).toBe('v2')
     })
 
-    it('fmapi method exists and returns DatabricksFmapiLanguageModel', () => {
-      const model = provider.fmapi('test-model')
+    it('chatCompletions method exists and returns DatabricksFmapiLanguageModel', () => {
+      const model = provider.chatCompletions('test-model')
 
       expect(model).toBeInstanceOf(DatabricksFmapiLanguageModel)
       expect(model.modelId).toBe('test-model')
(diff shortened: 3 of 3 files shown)


integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-message-parts.ts
```diff
@@ -97,54 +81,14 @@ const convertToolCallToContent = (toolCall: FmapiToolCall): LanguageModelV2Conte
   return {
     type: 'tool-call',
     toolCallId: toolCall.id,
-    toolName: toolCall.function.name,
+    toolName: DATABRICKS_TOOL_CALL_ID,
```
simonfaltum, line 84: Is it intended that the toolName is now hardcoded to a TOOL_CALL_ID ?

integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-input.ts
```diff
@@ -1,127 +1,128 @@
...
-          type: 'text',
-          text: serializeToolResult({
-            id: part.toolCallId,
-            content: convertToolResultOutputToContentValue(part.output),
-          }),
+            arguments: typeof part.input === 'string' ? part.input : JSON.stringify(part.input),
+          },
         })
         break
     }
```
simonfaltum, line 109: this case is entirely gone?

integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-input.ts
```diff
@@ -1,127 +1,128 @@
...
   }
 
-  return items
+  return {
+    role: 'assistant',
+    // Use null instead of empty string when there's no text content
+    // (required by Claude API when tool_calls are present)
```
simonfaltum: Claude API ? Isn't it FMAPI?

integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-input.ts
```diff
@@ -1,127 +1,128 @@
...
     }
   }
 
-  return items
+  return { role: 'user', content }
 }
 
-const convertAssistantContent = (message: LanguageModelV2AssistantMessage): FmapiContentItem[] => {
-  const items: FmapiContentItem[] = []
+const convertAssistantMessage = (message: LanguageModelV2AssistantMessage): FmapiInputMessage => {
```
simonfaltum: Not sure if I am reading these two functions right but it looks like convertUserMessage returns { role: 'user', content } where content can be [] but convertAssistantMessage returns null for empty content
Maybe we should align these for consistency?

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-prepare-tools.ts
```diff
@@ -0,0 +1,83 @@
...
+    case 'none':
+      return 'none'
+    case 'required':
+      return 'required'
+    case 'tool':
+      return { type: 'function', name: toolChoice.toolName }
+    default:
+      return 'auto'
+  }
+}
```
simonfaltum, line 83: if there is no toolChoice we return undefined, but if its an unknown toolChoice type this silently converts it to auto? is that intended?

integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-message-parts.ts
simonfaltum, line 84: The pattern of using DATABRICKS_TOOL_CALL_ID as the toolName and putting the actual tool name in providerMetadata.databricks.toolName is a bit confusing and not very clear why this is happening / why this needs to happen? idk if we should add a comment or some documentation explaining it

simonfaltum (approved):
Approving with a few comments and minor suggestions. Please take a look before merging

integrations/ai-sdk-provider/src/databricks-provider.ts
```diff
@@ -82,11 +90,15 @@ export const createDatabricksProvider = (
...
+    responses: createResponsesAgent,
+    chatCompletions: createFmapi,
     chatAgent: createChatAgent,
-    responsesAgent: createResponsesAgent,
-    fmapi: createFmapi,
     imageModel: notImplemented('ImageModel'),
     textEmbeddingModel: notImplemented('TextEmbeddingModel'),
     languageModel: notImplemented('LanguageModel'),
+    // @deprecated
+    responsesAgent: createResponsesAgent,
```
bbqiu: can we remove these in 1 minor version? i think it's best to drop these while usage is still low

integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-input.ts
bbqiu: would itbe possible to add some "integration" tests that use your local credentials to call databricks-hosted LLMs? would be good to get some of these API behaviors that you've noticed codified / tested for each release

we don't have to add to CI rn, we can add it when we get a runner w/ databricks creds

integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-message-parts.ts
bbqiu: if possible, could we avoid this? i think this will also make debugging apps that use the ai-sdk a bit more unintuitive

bbqiu (approved):
overall LGTM with a few comments

we can do a followup PR for this, but can we please add "integration" testing for
- claude-3-7 with tool calling
- gpt-5-2-codex (responses only) with tool calling
  - if we query with chat completions, we should throw a good error here
- gemini 2 (gemini 3 won't work bc of [thought signature requirement](https://databricks.atlassian.net/browse/ML-61153))
- GPT-OSS
- calling an MAS / KA / custom ResponseAgent
- ResponsesAgent hosted on apps

i put integration testing since it won't run as a part of CI, but we can add it once we get databricks creds loaded into a github runner

integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-input.ts
eLysgaard: Yeah makes sense

integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-input.ts
eLysgaard: Yeah this was a slip up. But there is actually small differences in how e.g. Claude and ChatGPT behaves when calling them through FMAPI, and this was an edge case caught when testing with a claude model

integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-input.ts
eLysgaard: Yes, because tool results should be passed in via `role: 'tool'` instead of as `role: 'assistant'`

integrations/ai-sdk-provider/src/fmapi-language-model/fmapi-convert-to-message-parts.ts
eLysgaard: I've added a section in the README which describes why this approach is required. TLDR when using the AI SDK you have to declare which tools are available when invoking the model. For agents authored elsewhere like ResponsesAgent and Agents on Apps we don't know which tools they have access to. The Chat UI template (https://github.com/databricks/app-templates/tree/main/e2e-chatbot-app-next) also uses this [here](https://github.com/databricks/app-templates/blob/0cd1cf1b586dcfb071e6cf8315b353001cc3856c/e2e-chatbot-app-next/server/src/routes/chat.ts#L209-L218).

https://github.com/databricks/databricks-ai-bridge/blob/a02e7b09c0640bff40b3a9423b00627b8f60f4ba/integrations/ai-sdk-provider/README.md?plain=1#L96-L141

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-prepare-tools.ts
eLysgaard: Good catch, this should also be undefined

integrations/ai-sdk-provider/src/databricks-provider.ts
eLysgaard: Yes, I'll remove it!

Issue #284, opened 2026-01-19
Eager imports of mlflow cause cascading failures on Windows


ClayGendron:
### Problem

`databricks_langchain` imports `mlflow` eagerly at module level in `genie.py:3`, which triggers mlflow's entire import chain including pyspark. This causes failures on Windows due to pyspark 4.x incompatibility.

### Import chain
```
from databricks_langchain import ChatDatabricks
  → databricks_langchain.genie imports mlflow
  → mlflow.types.schema imports pyspark
  → pyspark crashes on Windows (UnixStreamServer)
```

### Environment
- Windows 11
- Python 3.13.3
- databricks-langchain==0.11.0
- mlflow==2.22.4
- pyspark==4.1.1

### Suggested fix
Lazy-load mlflow only when GenieAgent is actually used, not at module import time.

PR #286, opened 2026-01-20
Lazy imports mlflow in genie to avoid windows pyspark failures

sunishsheth2009:
Lazy imports mlflow in genie to avoid windows pyspark failures

Resolving this issue: https://github.com/databricks/databricks-ai-bridge/issues/284

### DIFF
--- a/integrations/langchain/src/databricks_langchain/genie.py
+++ b/integrations/langchain/src/databricks_langchain/genie.py
@@ -1,25 +1,25 @@
 from typing import Callable, Optional
 
-import mlflow
 import pandas as pd
 from databricks.sdk import WorkspaceClient
 from databricks_ai_bridge.genie import Genie
 
 
-@mlflow.trace()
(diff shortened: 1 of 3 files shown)


sunishsheth2009:
> LGTM.
> 
> aside: I think at some point we should introduce a `_compat.py` to deal with these optional imports. For example in this case we could:
> 
> ```python
> @functool.decorator
> def mlflow_trace(f, *args, **kwargs):
>     try:
>         import mlflow
>         return mlflow.trace(f, *args, **kwargs)
>     except ImportError:
>         # warning
>         return f
> ```
> 
> Then use it as
> 
> ```python
> from ._compat import mlflow_trace
> 
> @mlflow_trace
> def _concat_message_array(messages):
>     ...
> ```

Yes that is a good idea which i thought about. But luckily there was mlflow.start_span that I could use instead here but if there are no options, we should be able to do this

PR #298, opened 2026-01-27
[ML-60694] Dont convert numbers to scientific notation for Genie

dhruv0811:
## Summary
- Prevent scientific notation display for FLOAT/DOUBLE/DECIMAL columns in Genie query results
- Keep float values as formatted strings for markdown output while preserving actual float types for `return_pandas=True`
- Use [`disable_numparse` ](https://pypi.org/project/tabulate/#:~:text=the%20longtable%20package.-,Column%20alignment,None%20(to%20disable%20alignment).&text=To%20disable%20this%20feature%20use%20disable_numparse=True%20.)on specific float column indices to preserve INT column alignment

## Test plan
- [x] Added `test_parse_query_result_preserves_float_precision` - verifies no scientific notation in markdown output
- [x] Added `test_parse_query_result_returns_floats_for_pandas` - verifies proper float64 types for DataFrame output
- [x] Added `test_parse_query_result_preserves_float_precision_with_truncate ` - verifies no scientific notation in truncation path
- [x] All existing tests pass (116 passed)

### DIFF
--- a/src/databricks_ai_bridge/genie.py
+++ b/src/databricks_ai_bridge/genie.py
@@ -50,6 +50,12 @@ def _parse_query_result(
         header: list[str] = [str(col["name"]) for col in columns]
         rows: list[list] = []
 
+        # Track float column indices (+1 offset for index column added by to_markdown)
+        float_column_indices: list[int] = []
+        for idx, col in enumerate(columns):
+            if col["type_name"] in ["FLOAT", "DOUBLE", "DECIMAL"]:
+                float_column_indices.append(idx + 1)
+
         for item in output["data_array"]:
             row: list = []
             for column, value in zip(columns, item):
@@ -61,7 +67,11 @@ def _parse_query_result(
                 if type_name in ["INT", "LONG", "SHORT", "BYTE"]:
                     row.append(int(value))
                 elif type_name in ["FLOAT", "DOUBLE", "DECIMAL"]:
-                    row.append(float(value))
+                    if return_pandas:
+                        row.append(float(value))
+                    else:
+                        # Keep as string to avoid scientific notation in markdown
+                        row.append(value)
                 elif type_name == "BOOLEAN":
                     row.append(value.lower() == "true")
                 elif type_name == "DATE":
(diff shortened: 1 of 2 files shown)


src/databricks_ai_bridge/genie.py
```diff
@@ -28,6 +28,15 @@ def _count_tokens(text):
     return len(encoding.encode(text))
 
 
+def _format_numeric_string(value_str: str) -> str:
+    """
+    Format a numeric string to remove unnecessary trailing zeros.
+    """
+    if "." in value_str:
+        value_str = value_str.rstrip("0").rstrip(".")
```
annzhang-db: Curious why we want to remove trailing zeros here, since our goal is to preserve the response from Genie as close to the original as possible. Does having trailing zeros mess up some of the subsequent logic? If not, can we just use a string representation of the float?

tests/databricks_ai_bridge/test_genie.py
annzhang-db, line 1: Can we also add a test for `_truncate_result` or where `truncate_result=True`?

src/databricks_ai_bridge/genie.py
```diff
@@ -50,6 +59,12 @@ def _parse_query_result(
         header: list[str] = [str(col["name"]) for col in columns]
         rows: list[list] = []
 
+        # Track float column indices (+1 offset for index column added by to_markdown)
+        float_column_indices: list[int] = []
```
annzhang-db, line 54: I believe we are using this as a "boolean" (if it has indices or not), so we don't actually need to keep track of the indices. In fact, what do you think about using `disable_numparse` always? Since integers will already be converted to ints, disabling numparse should only affect our floats which are getting converted to strings, then we don't need to keep track of this at all.

dhruv0811: The documentation on tabulate is bad, but this arg also accepts a list of column indices to disable number parsing only on those columns. If I set disable_numparse to true everywhere, it makes all columns left-aligned, which fails 13 tests. I can also set disable_numparse to true in the tests, so that they also expect left aligned columns. This felt like too large/widespread of a change, so I decided to limit the columns which we are disabling on. I can also follow copilots suggestion to pin the tabulate version so that we ensure this functionality is always present!

src/databricks_ai_bridge/genie.py
```diff
@@ -106,24 +125,30 @@ def _parse_query_result(
...
 
+        # Disable numparse for float columns to prevent scientific notation while preserving INT alignment
+        disable_numparse = float_column_indices if float_column_indices else False
+
         if truncate_results:
-            query_result = _truncate_result(dataframe)
+            query_result = _truncate_result(dataframe, disable_numparse)
         else:
-            query_result = dataframe.to_markdown()
+            query_result = dataframe.to_markdown(disable_numparse=disable_numparse)
```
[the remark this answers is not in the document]
dhruv0811: We're pinned to tabulate >=0.9.0 which supports the list-of-indices feature. So this should not be issue.

src/databricks_ai_bridge/genie.py
dhruv0811: Good point! I'll keep it as close to the original and leave any trailing zeros in the string representation. There is no subsequent logic that this messes up either.

src/databricks_ai_bridge/genie.py
annzhang-db: Ah okay, that makes sense, we can keep it this way! Let's make sure that there is a test for data without any float columns, but I think that should be covered by the existing tests. Sounds good to me :)

PR #299, opened 2026-01-28
Add async support to ChatDatabricks using databricks-openai

smurching:
Implements native async/await support for ChatDatabricks by integrating the AsyncDatabricksOpenAI client from the databricks-openai package.

Changes:
- Add databricks-openai>=0.3.0 dependency to pyproject.toml
- Create get_async_openai_client() utility function that returns AsyncDatabricksOpenAI client
- Add async_client cached property to ChatDatabricks class
- Implement _agenerate() method for async non-streaming invocations ([see LangChain docs](https://reference.langchain.com/python/langchain_core/language_models/?utm_source=chatgpt.com#langchain_core.language_models.BaseChatModel))
- Implement _astream() method for async streaming responses
- Support both Chat Completions API and Responses API in async mode

The async implementation mirrors the sync implementation's logic and supports all the same features including streaming, usage metadata, custom inputs, and both API types.

Tested with integration tests:
- test_chat_databricks_ainvoke: PASSED (2/2)
- test_chat_databricks_astream: PASSED (2/2)
- test_chat_databricks_abatch: PASSED (2/2)

### DIFF
--- a/integrations/langchain/src/databricks_langchain/utils.py
+++ b/integrations/langchain/src/databricks_langchain/utils.py
@@ -2,6 +2,7 @@ from typing import Any, List, Union
 from urllib.parse import urlparse
 
 import numpy as np
+from databricks_openai import AsyncDatabricksOpenAI
 from openai import AsyncOpenAI, OpenAI
 
 
@@ -59,7 +60,6 @@ def get_async_openai_client(workspace_client: Any = None, **kwargs) -> AsyncOpen
             such as timeout and max_retries.
     """
     from databricks.sdk import WorkspaceClient
-    from databricks_openai import AsyncDatabricksOpenAI
 
     # If workspace_client is provided, use it directly
     if workspace_client is not None:
--- a/integrations/langchain/pyproject.toml
+++ b/integrations/langchain/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
     "unitycatalog-langchain[databricks]>=0.3.0",
     "databricks-sdk>=0.65.0",
     "openai>=1.99.9",
-    "databricks-openai>=0.3.0",
+    "databricks-openai>=0.9.0",
     "langchain-mcp-adapters>=0.1.13",
     "databricks_mcp>=0.5.1",
 ]


integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -322,6 +323,18 @@ def client(self) -> OpenAI:
...
 
+    @cached_property
+    def async_client(self) -> AsyncOpenAI:
+        # Async OpenAI client using AsyncDatabricksOpenAI from databricks-openai
+        # Prepare kwargs for the SDK call
+        openai_kwargs = {}
+        if self.timeout is not None:
+            openai_kwargs["timeout"] = self.timeout
+        if self.max_retries is not None:
+            openai_kwargs["max_retries"] = self.max_retries
```
smurching: Can we share logic with the sync path for this

integrations/langchain/src/databricks_langchain/utils.py
```diff
@@ -49,6 +49,35 @@ def get_openai_client(workspace_client: Any = None, **kwargs) -> OpenAI:
...
+            # Otherwise, create default workspace client and use it
+            workspace_client = WorkspaceClient()
+            return AsyncDatabricksOpenAI(workspace_client=workspace_client, **kwargs)
+
+    except ImportError as e:
+        raise ImportError(
+            "Failed to create the async OpenAI client. "
+            "Please run `pip install databricks-openai databricks-sdk` to install "
+            "required dependencies."
+        ) from e
```
fanzeyi: We already declare `databricks-sdk>=0.65.0` as a required dependency as well as `databricks-openai` with this change. This seems redundant?

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -710,27 +762,122 @@ def _stream(
...
+        )
+
+        usage_chunk_emitted = False
+        final_usage: dict[str, int] | None = None
+
+        if self.use_responses_api:
+            prev_chunk = None
+            stream: AsyncStream[ResponseStreamEvent] = await self.async_client.responses.create(
+                **data
+            )
```
fanzeyi:
```suggestion
            stream = cast(AsyncStream[ResponseStreamEvent], await self.async_client.responses.create(
                **data
            ))
```

ty unfortunately can't work out through openai's typing hints that this will always be `AsyncStream[ResponseStreamEvent]`.

cast will enable typing to proceed from here.

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -710,27 +762,122 @@ def _stream(
...
+                    final_usage = usage
+            # Emit special usage chunk at end of stream
+            if stream_usage and final_usage and not usage_chunk_emitted:
+                yield self._build_usage_chunk(final_usage)
+                usage_chunk_emitted = True
+        else:
+            first_chunk_role = None
+            stream: AsyncStream[
+                ChatCompletionChunk
+            ] = await self.async_client.chat.completions.create(**data)
```
fanzeyi:
```suggestion
            stream = cast(AsyncStream[
                ChatCompletionChunk
            ], await self.async_client.chat.completions.create(**data))
```

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -710,27 +762,122 @@ def _stream(
...
-                )
+                    generation_chunk = ChatGenerationChunk(message=chunk_message)
+                    if run_manager:
+                        await run_manager.on_llm_new_token(
+                            generation_chunk.text,
+                            chunk=generation_chunk,
+                        )
+                    yield generation_chunk
+                elif chunk.choices:
+                    choice = chunk.choices[0]
```
fanzeyi, line 840: do we need to consider more than one `choices`?

> A list of chat completion choices. Can contain more than one elements if n is greater than 1.

https://platform.openai.com/docs/api-reference/chat-streaming/streaming

it seems we do support setting `n > 1`

PR #302, opened 2026-01-28
bugfix: dropped .done events and error on missing tool results

eLysgaard:
This PR resolves two bugs:
1. The dedupe logic matching on .done events was faulty and removed sequential .done events (e.g. when MAS emits tool call -> name -> tool result
2. Multi turn failed due to tool calls with no results. AI SDK expects this. Solution is to mark tool calls with no results as provider executed.

I discovered this while working to confirm upgrading to the latest AI SDK dependencies, but this bugs are not related to dependency versions and I figure we should fix them on their own.

With fix
image

Without fix
(Notice how some parts are missing in the response, and multi turn doesn't work)
image

# Generated summary

Fixes MissingToolResultsError for agent-orchestrated tools by marking tool calls without results as
  providerExecuted.

  When Databricks agents orchestrate tool calls server-side, the API executes the tool and returns the result
   inline - no separate function_call_output event is sent back to the client. The AI SDK expects every
  tool-call to have a matching tool-result, causing MissingToolResultsError.

  Solution: In the stream's flush callback, after all chunks have been processed, we identify tool calls
  without matching results and re-emit them with providerExecuted: true. This signals to the AI SDK that the
  tool was executed by the provider, so no client-side result is expected.

  - Skips MCP approval requests since those intentionally pause for user approval
  - Fixed two shouldDedupeOutputItemDone test expectations (should return false when there's no previous text
   to compare against)

  Test plan

  - Added test: emits correction tool-call with providerExecuted for tool calls without results
  - Added test: does not emit correction tool-call when result already exists
  - All 290 tests pass
  - TypeScript compiles without errors
  - Manual testing with chat app confirms agent tool calls work without errors

### DIFF
--- a/integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
+++ b/integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
@@ -28,6 +28,7 @@ import { convertToResponsesInput } from './responses-convert-to-input'
 import { getDatabricksLanguageModelTransformStream } from '../stream-transformers/databricks-stream-transformer'
 import { prepareResponsesTools } from './responses-prepare-tools'
 import { callOptionsToResponsesArgs } from './call-options-to-responses-args'
+import { MCP_APPROVAL_REQUEST_TYPE } from '../mcp'
 
 function mapResponsesFinishReason({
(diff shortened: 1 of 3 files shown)


fanzeyi (approved):
lgtm

PR #310, opened 2026-01-30
db-ai-bridge: Add hostname resolution support for Lakebase Pool

jennsun:
Allow LakebasePool to accept either instance name or hostname. When a hostname is provided (detected by .database. pattern), it resolves to the instance name by listing database instances.

This would support cases like when Databricks Apps resolves to hostname using resource reference (`valueFrom` in app.yaml)

conversation thread from prev PR: https://github.com/databricks/databricks-ai-bridge/pull/274#discussion_r2730754993

### DIFF
--- a/src/databricks_ai_bridge/lakebase.py
+++ b/src/databricks_ai_bridge/lakebase.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import re
 import time
 import uuid
 from enum import Enum
@@ -101,9 +102,11 @@ class _LakebasePoolBase:
 
         # If input is hostname (e.g., from Databricks Apps valueFrom resolution)
         # resolve to lakebase name
-        if self._is_hostname(instance_name):
+        if _is_hostname(instance_name):
             # Input is a hostname - resolve to instance name
-            self.instance_name, self.host = self._resolve_from_hostname(instance_name)
+            self.instance_name, self.host = _resolve_instance_name_from_hostname(
+                self.workspace_client, instance_name
+            )
         else:
             # Input is an instance name
             self.instance_name = instance_name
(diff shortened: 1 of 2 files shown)


src/databricks_ai_bridge/lakebase.py
```diff
@@ -97,34 +97,88 @@ def __init__(
...
-        except Exception as exc:
-            raise ValueError(
-                f"Unable to resolve Lakebase instance '{instance_name}'. "
-                "Ensure the instance name is correct."
-            ) from exc
-
-        resolved_host = getattr(instance, "read_write_dns", None) or getattr(
-            instance, "read_only_dns", None
-        )
+        # If input is hostname (e.g., from Databricks Apps valueFrom resolution)
```
bbqiu, line 103: does it make sense to have the helper inside of this method?

i think if the method here is accepting a param "instance name", then this might be a helpful standalone helper? something that will resolve to a validated instance name, whether given a hostname or an instance name

tests/databricks_ai_bridge/test_lakebase.py
```diff
@@ -1056,3 +1056,104 @@ def test_execute_grant_handles_insufficient_privilege(self):
         error_msg = str(exc_info.value)
         assert "Insufficient privileges" in error_msg
         assert "CAN MANAGE" in error_msg
```
bbqiu, line 1058: can we add an integration test where we use real values from dogfood to verify this logic?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -97,34 +97,88 @@ def __init__(
...
         self._cache_ts: float | None = None
 
+    @staticmethod
+    def _is_hostname(value: str) -> bool:
+        """Check if the value looks like a Lakebase hostname rather than an instance name."""
+        # Hostname pattern: instance-{uuid}.database.{env}.cloud.databricks.com
+        # or similar patterns containing ".database." and ending with a domain
+        return ".database." in value and (value.endswith(".com") or value.endswith(".net"))
+
+    def _resolve_from_hostname(self, hostname: str) -> tuple[str, str]:
```
bbqiu: can we include some examples of what a hostname should look like?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -97,34 +97,88 @@ def __init__(
...
+                if not instance_name:
+                    raise ValueError(
+                        f"Found matching instance for hostname '{hostname}' "
+                        "but instance name is not available."
+                    )
+                return instance_name, hostname
+
+        raise ValueError(
+            f"Unable to find database instance matching hostname '{hostname}'. "
+            "Ensure the hostname is correct and the instance exists."
```
bbqiu: could also be a perms issue

src/databricks_ai_bridge/lakebase.py
```diff
@@ -97,34 +97,88 @@ def __init__(
...
+            hostname: The database hostname (e.g., from Databricks Apps valueFrom: "database")
+
+        Returns:
+            Tuple of (instance_name, host)
+
+        Raises:
+            ValueError: If no matching instance is found
+        """
+        try:
+            instances = list(self.workspace_client.database.list_database_instances())
```
bbqiu: nit: can we include a comment here referencing our discussion about the performance implication of this

PR #313, opened 2026-02-02
OpenAI Client: Support querying apps

bbqiu:
support querying apps that support `/responses`

ran all integration tests

### DIFF
--- a/integrations/openai/src/databricks_openai/utils/clients.py
+++ b/integrations/openai/src/databricks_openai/utils/clients.py
@@ -99,15 +99,20 @@ def _wrap_app_error(e: Exception, app_name: str) -> ValueError:
             )
         elif status_code == 403:
             hint = f"Hint: Ensure you have CAN_USE permission on app '{app_name}'."
+        elif "DNS" in message or "resolution" in message.lower():
+            hint = (
+                f"Hint: App '{app_name}' may be stopped or unavailable. "
+                f"Check the app status in the Databricks workspace."
+            )
         elif status_code >= 500:
             hint = (
                 f"Hint: App '{app_name}' encountered an internal error. "
                 f"Check the app logs and status in the Databricks workspace."
             )
-        elif "DNS" in message or "resolution" in message.lower():
+        elif status_code != 200:
             hint = (
-                f"Hint: App '{app_name}' may be stopped or unavailable. "
-                f"Check the app status in the Databricks workspace."
+                f"Hint: App '{app_name}' returned a non-200 status code: {status_code}. "
+                f"Check the app logs and status in the Databricks workspace."
             )
         else:
             hint = None
--- a/integrations/openai/tests/integration_tests/test_apps_errors.py
+++ b/integrations/openai/tests/integration_tests/test_apps_errors.py
@@ -20,7 +20,7 @@ class TestDatabricksAppsErrors:
         "app_name",
         [
             "ai-slide-generator-dev",
-            "agent-builder-assistant",
+            "agent-lc-ts-dev",
             "at-observability-poc",
         ],
     )
@@ -39,10 +39,10 @@ class TestDatabricksAppsErrors:
             )
 
     def test_no_permission_to_app(self):
-        client_id = os.environ.get("TEST_SP_CLIENT_ID")
-        client_secret = os.environ.get("TEST_SP_CLIENT_SECRET")
+        client_id = os.environ.get("DATABRICKS_CLIENT_ID")
+        client_secret = os.environ.get("DATABRICKS_CLIENT_SECRET")
         if not client_id or not client_secret:
-            pytest.skip("TEST_SP_CLIENT_ID and TEST_SP_CLIENT_SECRET env vars required")
+            pytest.skip("DATABRICKS_CLIENT_ID and DATABRICKS_CLIENT_SECRET env vars required")
 
         ws = WorkspaceClient(
             host="https://e2-dogfood.staging.cloud.databricks.com",


integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -80,27 +184,53 @@ class DatabricksOpenAI(OpenAI):
...
+        workspace_client: WorkspaceClient | None = None,
+        base_url: str | None = None,
+    ):
         if workspace_client is None:
             workspace_client = WorkspaceClient()
 
-        current_host = workspace_client.config.host
+        self._workspace_client = workspace_client
+
+        if base_url is not None:
```
bbqiu, line 249: only do this check if the base url has `databricksapps` in it (meaning that the baseurl refers to an app)

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -135,9 +271,38 @@ class AsyncDatabricksChat(AsyncChat):
...
+                base_url=app_url,
+                api_key="no-token",
+                http_client=_get_authorized_async_http_client(self._workspace_client),
+            )
+        return self._app_clients_cache[app_name]
+
+    async def create(self, **kwargs):
+        model = kwargs.get("model", "")
+
+        if isinstance(model, str) and model.startswith("apps/"):
```
bbqiu: do not use this magic string, define it in a var and call len() on it instead of having 5:

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -156,27 +321,53 @@ class AsyncDatabricksOpenAI(AsyncOpenAI):
...
+        >>> client = AsyncDatabricksOpenAI(
+        ...     base_url="https://my-app.aws.databricksapps.com",
+        ...     workspace_client=WorkspaceClient(),  # Must use OAuth
+        ... )
+        >>> response = await client.responses.create(
+        ...     input=[{"role": "user", "content": "Hello"}],
+        ... )
+
+    Example - Query a Databricks App by name (apps/ prefix):
+        >>> client = AsyncDatabricksOpenAI()
```
bbqiu, line 377: also explain here that we require a workspace client intialized with an oauth token

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -156,27 +321,53 @@ class AsyncDatabricksOpenAI(AsyncOpenAI):
         workspace_client: Databricks WorkspaceClient to use for authentication. Pass a custom
             WorkspaceClient to set up your own authentication method. If not provided, a default
             WorkspaceClient will be created using standard Databricks authentication resolution.
+        base_url: Optional base URL for direct App access. When provided, OAuth authentication
```
bbqiu: clarify that not all base urls require an oauth token, only the ones that we need when we point towards a databricks app

integrations/openai/tests/integration_tests/test_apps_errors.py
```diff
@@ -0,0 +1,56 @@
+import os
+
+import pytest
+from databricks.sdk import WorkspaceClient
+
+from databricks_openai import DatabricksOpenAI
+
+DATABRICKS_CONFIG_PROFILE = "dogfood"
```
bbqiu: read this from env vars isntead

integrations/openai/tests/integration_tests/test_apps_errors.py
```diff
@@ -0,0 +1,56 @@
+import os
```
bbqiu, line 1: add another integration test file for querying an app that actually does return for /resopnses.

can use dogfood and app name agent-customer-support to test

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -64,6 +139,40 @@ class DatabricksChat(Chat):
...
+        self._app_clients_cache: dict[str, OpenAI] = {}
+
+    def _get_app_client(self, app_name: str) -> OpenAI:
+        """Get or create a client for a specific app."""
+        if app_name not in self._app_clients_cache:
+            _validate_oauth_for_apps(self._workspace_client)
+            app_url = _get_app_url(self._workspace_client, app_name)
+            self._app_clients_cache[app_name] = OpenAI(
+                base_url=app_url,
+                api_key="no-token",
```
jennsun, line 172: can you leave a comment here specifying why we aren't passing in a token here since we are doing our auth via the http client to inject databricks oauth tokens in each request instead? Also wondering if it's cleaner to have potentially empty "" here instead since `no-token` seems like a magic string here

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -43,11 +49,80 @@ def _should_strip_strict(model: str | None) -> bool:
...
+
+
+def _validate_oauth_for_apps(workspace_client: WorkspaceClient) -> None:
+    """Validate that workspace_client uses OAuth (required for Apps)."""
+    try:
+        workspace_client.config.oauth_token()
+    except Exception as e:
+        raise ValueError(
+            "Querying Databricks Apps requires OAuth authentication. "
+            "See https://docs.databricks.com/aws/en/dev-tools/auth/oauth-u2m.html"
```
jennsun: just wondering if there any value in linking to https://docs.databricks.com/aws/en/dev-tools/auth/oauth-m2m as well? referencing the exception we throw in mlflow to validate oauth authentication, perhaps we should keep consistent here:
https://github.com/mlflow/mlflow/blob/master/mlflow/utils/databricks_utils.py#L1568

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -2,11 +2,17 @@
...
 from httpx import AsyncClient, Auth, Client, Request, Response
-from openai import AsyncOpenAI, OpenAI
+from openai import APIConnectionError, APIStatusError, AsyncOpenAI, OpenAI
 from openai.resources.chat import AsyncChat, Chat
 from openai.resources.chat.completions import AsyncCompletions, Completions
+from openai.resources.responses import AsyncResponses, Responses
 from typing_extensions import override
 
+# Prefix for routing requests to Databricks Apps
+_APPS_ENDPOINT_PREFIX = "apps/"
```
jennsun, line 12: Do we feel strongly about having the `apps/` prefix here or do we want to be consistent with the mlflow client and use `apps:/`?

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -43,11 +49,80 @@ def _should_strip_strict(model: str | None) -> bool:
...
+
+
+def _wrap_app_error(e: Exception, app_name: str) -> ValueError:
+    """Wrap OpenAI API errors with helpful hints for Databricks Apps."""
+    if isinstance(e, APIStatusError):
+        status_code = e.status_code
+        message = e.message
+        if status_code == 404 or status_code == 405:
+            hint = (
+                f"Hint: App '{app_name}' may not support the Responses API. "
```
jennsun:
```suggestion
                f"Hint: App '{app_name}' may not support the OpenAI Responses API. "
```
nit

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -43,11 +49,80 @@ def _should_strip_strict(model: str | None) -> bool:
...
+                f"Ensure the app implements the /responses endpoint."
+            )
+        elif status_code == 403:
+            hint = f"Hint: Ensure you have CAN_QUERY permission on app '{app_name}'."
```
jennsun: should the permission be `CAN_USE` here for apps? 
apps: https://docs.databricks.com/aws/en/dev-tools/databricks-apps/permissions#permission-levels

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -43,11 +49,80 @@ def _should_strip_strict(model: str | None) -> bool:
...
+        return ValueError(error_msg)
+    elif isinstance(e, APIConnectionError):
+        message = str(e)
+        if "DNS" in message or "resolution" in message.lower():
+            hint = (
+                f"Hint: App '{app_name}' may be stopped or unavailable. "
+                f"Check the app status in the Databricks workspace."
+            )
+        else:
+            hint = f"Hint: App '{app_name}' may be starting up or unavailable."
```
jennsun:
```suggestion
            hint = f"Hint: App '{app_name}' may be starting up or unavailable."
                       f"Check the app status in the Databricks workspace."
```

integrations/openai/src/databricks_openai/utils/clients.py
jennsun: let's cover status code 401 for this too - I recall getting 401s from not having CAN_USE permission on app

Image

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -43,11 +49,80 @@ def _should_strip_strict(model: str | None) -> bool:
...
+        elif status_code == 403:
+            hint = f"Hint: Ensure you have CAN_QUERY permission on app '{app_name}'."
+        elif "DNS" in message or "resolution" in message.lower():
+            hint = (
+                f"Hint: App '{app_name}' may be stopped or unavailable. "
+                f"Check the app status in the Databricks workspace."
+            )
+        else:
+            hint = None
+
```
jennsun, line 119: aside from 4xxs could we cover 5xxs as well? 5xxs/internal errors: could say message like:
```
"Hint: App '{app_name}' encountered an internal error. "
"Check the app logs and status in the Databricks workspace."
```

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -80,27 +189,59 @@ class DatabricksOpenAI(OpenAI):
         workspace_client: Databricks WorkspaceClient to use for authentication. Pass a custom
             WorkspaceClient to set up your own authentication method. If not provided, a default
             WorkspaceClient will be created using standard Databricks authentication resolution.
+        base_url: Optional base URL to override the default serving endpoints URL. When the URL
```
jennsun, line 207: do we anticipate this base_url being used for anything other than apps? wonder if `app_url` would be clearer for users to pass in

jennsun (approved):
Overall LGTM! Had some specifics related to apps / think we could add more robust error detection for the different cases querying apps throws :)

integrations/openai/src/databricks_openai/utils/clients.py
bbqiu: are we sure that this isn't a rewrapping from the MCP client? from my integration testing, apps that you do not have access to return a 403

integrations/openai/src/databricks_openai/utils/clients.py
bbqiu: i don't think that consistency across the two is necessary - the colon feels a bit unnecessary?

integrations/openai/src/databricks_openai/utils/clients.py
bbqiu: yup - customers use this base_url for other stuff as well!

integrations/openai/src/databricks_openai/utils/clients.py
jennsun: discussed offline - this 401 is a specific mcp initialization error, not from lacking can_use on apps https://modelcontextprotocol.io/docs/tutorials/security/authorization#the-authorization-flow-step-by-step

integrations/openai/src/databricks_openai/utils/clients.py
bbqiu: generically added a catch for all non-200 error codes

PR #316, opened 2026-02-03
db-openai: Add AsyncDatabricksSession class to support Session Protocol for Stateful Conversation Management via SQLAlchemy engine

jennsun:
Adds AsyncDatabricksSession, a session storage implementation for the OpenAI Agents SDK that persists conversation history to Databricks Lakebase. 

This class subclasses [OpenAI's SQLAlchemySession](https://openai.github.io/openai-agents-python/sessions/sqlalchemy_session/) [original code](https://github.com/openai/openai-agents-python/blob/main/src/agents/extensions/memory/sqlalchemy_session.py) to inherit all SQL logic while adding Lakebase-specific features:
  - Automatic OAuth token rotation via SQLAlchemy's do_connect event
  - Instance name resolution and username inference from _LakebasePoolBase

More on Session protocol:
https://openai.github.io/openai-agents-python/ref/memory/session/#agents.memory.session.Session

Using SQLAlchemy's default QueuePool (pool_size=5, max_overflow=10) for connection pooling

Usage: 
```
from databricks_openai.agents import AsyncDatabricksSession

  session = AsyncDatabricksSession(
  session_id=session_id,
        instance_name=LAKEBASE_INSTANCE_NAME,
    )
   result = Runner.run_streamed(agent, input=messages, session=session)

```

Example queries:
```
curl -X POST http://localhost:8000/invocations \
    -d '{"input": [{"role": "user", "content": "Hello I live in SF!"}]}'
```
returns responses with session id:
```
{"object":"response","output":[{"type":"message","id":"__fake_id__","content":[{"annotations":[],"text":"Hi! What part of San Francisco are you in, and what are you looking for—recommendations (food/coffee, parks, things to do), help planning a day, or something else?","type":"output_text","logprobs":[]}],"role":"assistant","status":"completed","provider_data":{"model":"databricks-gpt-5-2","response_id":"chatcmpl-D5Ki6f7TKBNrVVfuxDL0Lu16YCQjz"}}],"custom_outputs":{"session_id":"fd57ff2c-1d66-4da3-ba28-3216d4e6d86e"}}
```
follow-up stateful question:
```
curl -X POST http://localhost:8000/invocations \
    -d '{
        "input": [{"role": "user", "content": "What city did I say I live in?"}],
        "custom_inputs": {"session_id": "fd57ff2c-1d66-4da3-ba28-3216d4e6d86e"}
    }'

```
gives us:
```
{"object":"response","output":[{"type":"message","id":"__fake_id__","content":[{"annotations":[],"text":"You said you live in SF (San Francisco).","type":"output_text","logprobs":[]}],"role":"assistant","status":"completed","provider_data":{"model":"databricks-gpt-5-2","response_id":"chatcmpl-D5KiakiUvMW2weIeLXUuPesxyNpwv"}}],"custom_outputs":{"session_id":"fd57ff2c-1d66-4da3-ba28-3216d4e6d86e"}}
```

testing:
unit + integration tests

sample agent: [OpenAI MemorySession Stateful Agent Example](https://github.com/databricks/app-templates/pull/109)
sample app: https://eng-ml-agent-platform.staging.cloud.databricks.com/apps/j-openai-stateful?o=2850744067564480
image

### DIFF
--- a/integrations/openai/src/databricks_openai/agents/session.py
+++ b/integrations/openai/src/databricks_openai/agents/session.py
@@ -178,9 +178,7 @@ class AsyncDatabricksSession(SQLAlchemySession):
         """Get cached AsyncLakebaseSQLAlchemy or create a new one.
         The cache key uses both instance_name and engine_kwargs
         """
-        cache_key = cls._build_cache_key(
-            instance_name, pool_recycle=pool_recycle, **engine_kwargs
-        )
+        cache_key = cls._build_cache_key(instance_name, pool_recycle=pool_recycle, **engine_kwargs)
 
         if use_cached_engine:
             with cls._lakebase_sql_alchemy_cache_lock:


integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,233 @@
...
+
+        # Create engine with NullPool to avoid connection caching issues
+        # Each operation gets a fresh connection with a fresh token check
+        engine = create_async_engine(
+            url,
+            poolclass=NullPool,
+            connect_args={"sslmode": DEFAULT_SSLMODE},
+            **engine_kwargs,
+        )
+
```
jennsun, line 196: the engine should already handle connection pooling, I found this in the sqlalchemy docs: https://docs.sqlalchemy.org/en/14/core/pooling.html#connection-pool-configuration

where the QueuePool params are passed in when we create engine with default size of 5 here: https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/pool/impl.py#L74

and we create the engine here: https://github.com/databricks/databricks-ai-bridge/pull/316/changes#diff-bb5bcca7a0a0826e79bda9e1ddfaa096a923a2f2081f2aeafd73a25c495ca05eR189

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,234 @@
...
+            port=DEFAULT_PORT,
+            database=DEFAULT_DATABASE,
+        )
+
+        # Use default QueuePool with connection recycling.
+        # Connections are recycled before token cache expires (50 min),
+        # ensuring fresh tokens are injected via do_connect event.
+        engine = create_async_engine(
+            url,
+            pool_recycle=DEFAULT_POOL_RECYCLE_SECONDS,
```
jennsun: Connection pooling happens here via SQLAlchemy's default QueuePool (SQLAlchemy defaults to QueuePool) and pool_recycle=2700 setting ensures connections are recycled every 45 minutes (before the 50-min token cache expires), at which point the do_connect event injects a fresh token.

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,234 @@
+"""
```
bbqiu, line 1: do we want this to be importable from databricks_openai.agents?

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,234 @@
...
+
+# Constants for Lakebase connection
+DEFAULT_TOKEN_CACHE_DURATION_SECONDS = 50 * 60  # 50 minutes
+DEFAULT_POOL_RECYCLE_SECONDS = 45 * 60  # 45 minutes (before token cache expires)
+DEFAULT_SSLMODE = "require"
+DEFAULT_PORT = 5432
+DEFAULT_DATABASE = "databricks_postgres"
+
+
+class _LakebaseCredentials(_LakebasePoolBase):
```
bbqiu: does it make sense to rename `_LakebasePoolBase`?

also could you remind me why we didn't have the cache lock as an attribute of the _LakebasePoolBase instance?

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,234 @@
...
+        with self._cache_lock:
+            if cached := self._get_cached_token():
+                return cached
+            token = self._mint_token()
+            self._cached_token = token
+            self._cache_ts = time.time()
+            return token
+
+
+class MemorySession(SQLAlchemySession):
```
bbqiu: pls correct me if i'm wrong, but i think this is async only?

can we leave some clarification about this in the docstrings / throw a helpful error if someone tries to run this synchronously

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,234 @@
...
+        # ensuring fresh tokens are injected via do_connect event.
+        engine = create_async_engine(
+            url,
+            pool_recycle=DEFAULT_POOL_RECYCLE_SECONDS,
+            connect_args={"sslmode": DEFAULT_SSLMODE},
+            **engine_kwargs,
+        )
+
+        # Attach event to inject Lakebase token before each connection
+        # Note: do_connect fires on sync_engine even for async operations
```
bbqiu: we are creating an async engine right? is this an old comment

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,234 @@
...
+        """The resolved Lakebase host."""
+        return self._credentials.host
+
+    @property
+    def username(self) -> str:
+        """The database username."""
+        return self._credentials.username
+
+    @property
+    def connection_url(self) -> str:
```
bbqiu: nit: do we wanna reuse this in _create_engine func above?

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,234 @@
...
+                SQLAlchemy's create_async_engine().
+        """
+        # Create credential provider
+        self._credentials = _LakebaseCredentials(
+            instance_name=instance_name,
+            workspace_client=workspace_client,
+            token_cache_duration_seconds=token_cache_duration_seconds,
+        )
+
+        engine = self._create_engine(**engine_kwargs)
```
bbqiu: iiuc, this can potentially be run on every new conversation right?

is there any way to reuse an engine across sessions? if not, should we try to make these operations async via event loops

integrations/openai/src/databricks_openai/agents/session.py
jennsun: sure, I'll make this importable so instead of:
```
from databricks_openai.agents.session import MemorySession
```
import path will look like:
```
from databricks_openai.agents import MemorySession
```

integrations/openai/src/databricks_openai/agents/session.py
jennsun: we had separate locks (sync vs async) for the sync vs async LakebasePools we implemented, so each of the subclasses adds their own cache lock

jennsun: I'll rename `_LakebasePoolBase` to be `_LakebaseBase` as it's a more generic class for resolving lakebase host/username/token caching - the actual pooling logic is actually implemented in the subclasses LakebasePool/AsyncLakebasePool

integrations/openai/src/databricks_openai/agents/session.py
jennsun: yes, but the async engine is a wrapper around a sync engine:

there is not yet an “async” version of a SQLAlchemy event handler

"Events can be registered at the instance level (e.g. a specific [AsyncEngine](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncEngine) instance) by associating the event with the sync attribute that refers to the proxied object. For example to register the [PoolEvents.connect()](https://docs.sqlalchemy.org/en/20/core/events.html#sqlalchemy.events.PoolEvents.connect) event against an [AsyncEngine](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncEngine) instance, use its [AsyncEngine.sync_engine](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncEngine.sync_engine) attribute as target."
link: https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#using-events-with-the-asyncio-extension

integrations/openai/src/databricks_openai/agents/session.py
jennsun: yes it's async only - taking a look at the source code all of these classes implement the async interface (since they follow the session protocol):
https://github.com/openai/openai-agents-python/blob/main/src/agents/memory/session.py 

i'll cover this in unit tests/rename to AsyncDatabricksSession to make it clearer

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,308 @@
...
+        with self._cache_lock:
+            if cached := self._get_cached_token():
+                return cached
+            token = self._mint_token()
+            self._cached_token = token
+            self._cache_ts = time.time()
+            return token
+
+
+class AsyncDatabricksSession(SQLAlchemySession):
```
bbqiu, line 55: from talking to the research team working on DBRA, they actually have a very similar snippet as us to manage a SQLAlchemy connection to lakebase: https://sourcegraph.prod.databricks-corp.com/databricks-eng/universe/-/blob/research/aroll/app/aroll_app/db/connection.py?L162-182

would it make sense for us to further abstract this by providing a similar AsyncLakebaseSQLAlchemy / LakebaseSQLAlchemy class?

jennsun: discussed offline but I'll refactor such that:

- in db-ai-bridge we add AsyncLakebaseSqlAlchemy support for creating SQLAlchemy engines (similar to https://sourcegraph.prod.databricks-corp.com/databricks-eng/universe/-/blob/research/aroll/app/aroll_app/db/connection.py?L185)

- in db-openai we subclass SQLAlchemySession in AsyncDatabricksSession (session protocol is async + specific to openai agents sdk) and pass in engine from AsyncLakebaseSqlAlchemy

this will create much cleaner separation of concerns for future frameworks to reuse any sqlalchemy engines etc!

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+        asyncio.run(main())
+        ```
+    """
+
+    # Class-level cache for AsyncLakebaseSQLAlchemy instances, keyed by instance_name.
+    # This allows multiple AsyncDatabricksSession instances to share a single engine/pool.
+    _lakebase_sql_alchemy_cache: dict[str, AsyncLakebaseSQLAlchemy] = {}
+    _lakebase_sql_alchemy_cache_lock = Lock()
```
jennsun, line 93: thoughts on the class-level cache for AsyncLakebaseSQLAlchemy engines keyed by instance_name?

this is so we reuse a single SQLAlchemy engine / pool per Lakebase instance, avoiding repeated pool creation, TCP handshakes, and auth setup.

sessions are still created per Runner.run(), but engines are shared

tests/databricks_ai_bridge/test_lakebase.py
```diff
@@ -255,7 +255,7 @@ async def test_async_lakebase_pool_logs_cache_seconds(monkeypatch, caplog):
         )
 
     assert any(
-        record.levelno == logging.INFO and re.search(r"cache=3000s$", record.getMessage())
+        record.levelno == logging.INFO and re.search(r"cache=900s$", record.getMessage())
```
bbqiu, line 259: nit: can we add a few tests for AsyncLakebaseSQLAlchemy

src/databricks_ai_bridge/lakebase.py
```diff
@@ -6,11 +6,15 @@
 import uuid
```
bbqiu, line 6: (ok for followup PR) we should probably think about separating this file into a few separate ones since it's getting quite long

integrations/openai/tests/unit_tests/test_clients.py
```diff
@@ -204,7 +206,7 @@ def test_chat_completions_strips_strict_for_claude(self):
                 client.chat.completions.create(
                     model="databricks-claude-3-7-sonnet",
                     messages=[{"role": "user", "content": "hi"}],
-                    tools=tools,
+                    tools=cast(Any, tools),
```
bbqiu, line 230: did we delete this change from the diff? i think we still need it cc @fanzeyi who ran into a bug that that was fixing earlier

integrations/openai/pyproject.toml
```diff
@@ -20,6 +20,12 @@ dependencies = [
     "openai-agents>=0.5.0"
 ]
 
+[project.optional-dependencies]
+memory = [
```
bbqiu, line 24: can we update the CI job for this memory extra too

integrations/openai/src/databricks_openai/agents/__init__.py
```diff
@@ -1,3 +1,4 @@
 from databricks_openai.agents.mcp_server import McpServer
+from databricks_openai.agents.session import AsyncDatabricksSession
 
-__all__ = ["McpServer"]
+__all__ = ["McpServer", "AsyncDatabricksSession"]
```
bbqiu: how are we able to do this without running into import issues? don't we need to do a similar thing as what we did last time with the langchain CheckpointSaver class

integrations/openai/tests/unit_tests/test_clients.py
jennsun: context: https://github.com/databricks/databricks-ai-bridge/pull/274#discussion_r2756134906

integrations/openai/pyproject.toml
jennsun: good catch!

integrations/openai/src/databricks_openai/agents/session.py
bbqiu: this approach looks good to me to minimize IO. two comments:
- we may want to include a param for a func for customers to customize the cache key. currently, diff engine kwargs for the same instance name will be ignored
- let's also call this out in the docstring and add a param to optionally disable this engine caching

bbqiu: the best case would be include engine kwargs + instance name in the cache key

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -0,0 +1,177 @@
+"""
+AsyncDatabricksSession - Async SQLAlchemy-based session storage for Databricks Lakebase.
+
+This module provides a AsyncDatabricksSession class that subclasses OpenAI's SQLAlchemySession
```
bbqiu: nit: grammar. there's another docstring w/ the same typo
```suggestion
This module provides an AsyncDatabricksSession class that subclasses OpenAI's SQLAlchemySession
```

bbqiu (approved):
lgtm, feel free to merge after addressing comments!

integrations/openai/src/databricks_openai/agents/session.py
jennsun: sounds good - going to create cache key that takes into consideration both instance name + engine kwards, as well as ability to not cache the engines (but defaults to caching)

integrations/openai/tests/unit_tests/test_clients.py
jennsun: added this back [here](https://github.com/databricks/databricks-ai-bridge/pull/316/changes#diff-f2bb640113bd724e62965bf48aaf16e342e05ee0d840eef176dcd883529a9258R29) and included [unit tests](https://github.com/databricks/databricks-ai-bridge/pull/316/changes#diff-adbecfa5c86bf2741b62cd58d88073c2ce8a965ca9407e6665d0d02a5285caf9R131) to make sure the non-list inputs are handled gracefully!

PR #215, opened 2025-11-11
Add support for passing WorkspaceClient to VectorSearchRetrieverTool for langchain and openai

bbqiu:
to fix https://github.com/bbqiu/agent-on-app-proto/issues/32

allow for passing of workspace client with PAT and OAuth-m2m auth types

manually ran integration tests with different env vars (left is SP client id/secret, right is pat):
screenshot 2025-11-11-11-13-17-iTerm2

screenshot 2025-11-11-11-15-28-iTerm2

PR #317, opened 2026-02-05
Added Vector Search Integration Tests

dhruv0811:
## Summary

Added comprehensive Vector Search integration tests covering the core bridge utilities, LangChain wrapper, and OpenAI wrapper. Tests validate our bridge code's behavior — kwargs pass-through, auth credential forwarding, filter handling, column selection, and response parsing — against a live Databricks workspace. Also restructured `tests/integration_tests/` into feature-based subdirectories to support future test suites.

**64 total integration tests** (32 core + 17 LangChain + 15 OpenAI), all passing against prod workspace.

### Bug fix: PAT auth missing `workspace_url`

Starting with `databricks-vectorsearch` 0.61 (released Oct 28, 2025), `VectorSearchClient` requires `workspace_url` when `personal_access_token` is provided. Our bridge code (introduced Nov 11, 2025 in PR #215) never forwarded `workspace_url` in the PAT auth path, causing `ValueError: Workspace URL is required` for any user authenticating with a Personal Access Token.

**Fix:** Added `client_args.setdefault("workspace_url", config.host)` in the PAT auth path of both:
- `integrations/langchain/src/databricks_langchain/vectorstores.py` (line 279)
- `integrations/openai/src/databricks_openai/vector_search_retriever_tool.py` (line 122)

Updated corresponding unit test assertions in both integrations to expect the corrected behavior.

### Test structure
```
tests/integration_tests/
├── conftest.py                              # Shared: workspace_client, markers
├── vector_search/
│   ├── conftest.py                          # VS-specific: client, indexes, RUN_INTEGRATION_TESTS gate
│   ├── test_vectorsearch_behavior.py        # Smoke tests: search + filter per index type, score validation
│   └── test_bridge_utilities.py             # IndexDetails, parse_vector_search_response, validators
├── lakebase/
│   └── test_lakebase_integration.py         # Moved here (own LAKEBASE_INTEGRATION_TESTS gate)

integrations/langchain/tests/integration_tests/
└── test_langchain_vectorsearch.py           # VectorSearchRetrieverTool + DatabricksVectorSearch
                                             #   init, execution, kwargs, auth paths, filtering

integrations/openai/tests/integration_tests/
└── test_openai_vectorsearch.py              # VectorSearchRetrieverTool
                                             #   init, execution, kwargs, auth paths, filtering
```

## Test plan
- [x] All 32 core vector search integration tests pass
- [x] All 17 LangChain vector search integration tests pass
- [x] All 15 OpenAI vector search integration tests pass
- [x] All unit tests pass (53 OpenAI + 41 LangChain vector search retriever tool)
- [x] CI passes via ai-oss runner: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/21924868689

### DIFF
--- a/integrations/langchain/tests/integration_tests/test_langchain_vectorsearch.py
+++ b/integrations/langchain/tests/integration_tests/test_langchain_vectorsearch.py
@@ -222,11 +222,15 @@ class TestLangChainKwargsPassThrough:
         )
 
     def test_score_threshold_kwarg_accepted(self, vectorstore):
+        # Use a gibberish query + high threshold to prove score_threshold is forwarded
         results = vectorstore.similarity_search_with_score(
-            "machine learning", k=3, score_threshold=0.99
+            "xyzzy_nonexistent_query_12345", k=3, score_threshold=0.99
         )
-        # High threshold may return fewer or no results, but should not error
+        # Garbage query with 0.99 threshold should return 0 results
         assert isinstance(results, list)
+        assert len(results) == 0, (
+            f"Expected 0 results with gibberish query + score_threshold=0.99, got {len(results)}"
+        )
 
     def test_invalid_kwargs_silently_dropped(self, vectorstore):
         docs = vectorstore.similarity_search(
--- a/integrations/openai/tests/integration_tests/test_openai_vectorsearch.py
+++ b/integrations/openai/tests/integration_tests/test_openai_vectorsearch.py
@@ -216,9 +216,13 @@ class TestOpenAIKwargsPassThrough:
             workspace_client=workspace_client,
             score_threshold=0.99,  # type: ignore[unknown-argument]
         )
-        result = tool.execute(query="machine learning")
-        # High threshold may return fewer or no results, but should not error
+        # Use a gibberish query + high threshold to prove score_threshold is forwarded
+        result = tool.execute(query="xyzzy_nonexistent_query_12345")
+        # Garbage query with 0.99 threshold should return 0 results
         assert isinstance(result, list)
+        assert len(result) == 0, (
+            f"Expected 0 results with gibberish query + score_threshold=0.99, got {len(result)}"
+        )
 
     def test_invalid_kwargs_filtered_out(self, workspace_client):
         from databricks_openai.vector_search_retriever_tool import (
--- a/tests/integration_tests/conftest.py
+++ b/tests/integration_tests/conftest.py
@@ -34,10 +34,20 @@ def workspace_client():
     The SDK auto-detects auth from env vars (e.g. DATABRICKS_HOST,
     DATABRICKS_CLIENT_ID, DATABRICKS_CLIENT_SECRET for OAuth M2M).
     For local testing, set DATABRICKS_CONFIG_PROFILE instead.
+
+    Converts non-standard auth types (e.g. databricks-cli) to PAT so
+    credentials are forwarded correctly to downstream clients like
+    VectorSearchClient and Genie.
     """
     from databricks.sdk import WorkspaceClient
 
-    return WorkspaceClient()
+    wc = WorkspaceClient()
+    if wc.config.auth_type not in ("pat", "oauth-m2m", "model_serving_user_credentials"):
+        headers = wc.config.authenticate()
+        token = headers.get("Authorization", "").replace("Bearer ", "")
+        if token:
+            return WorkspaceClient(host=wc.config.host, token=token, auth_type="pat")
+    return wc
 
 
 # =============================================================================


integrations/openai/tests/unit_tests/test_clients.py
dhruv0811, line 1: Note: these minor changes are related to some typechecking errors from before

tests/integration_tests/vector_search/test_bridge_utilities.py
```diff
@@ -0,0 +1,291 @@
...
+        results = parse_vector_search_response(
+            search_resp=search_resp,
+            retriever_schema=retriever_schema,
+            document_class=dict,
+            include_score=True,
+        )
+
+        assert len(results) > 0
+        doc, _ = results[0]
+        assert "score" in doc["metadata"]
```
annzhang-db, line 201: Can we check that the score is a (reasonable) number? I believe we had a previous issue with parsing the score and it was populated as an empty string e.x. `"score": ""`.

tests/integration_tests/vector_search/test_vectorsearch_behavior.py
```diff
@@ -0,0 +1,213 @@
...
+# =============================================================================
+# Delta-Sync Search Behavior Tests
+# =============================================================================
+
+
+@pytest.mark.behavior
+class TestDeltaSyncSearchBehavior:
+    """Verify similarity search behavior on delta-sync managed embeddings index."""
+
+    def test_basic_text_search_returns_results(self, delta_sync_index):
```
annzhang-db: A lot of the tests here are similar (with different `num_results`), I think we could test with just `num_results=3` and get similar coverage to make sure that it is limited to just 3 but we can still get multiple. Generally, we want to limit how many calls we make to vector search if possible.

tests/integration_tests/vector_search/test_vectorsearch_behavior.py
```diff
@@ -0,0 +1,213 @@
+"""
+Behavior validation tests for Vector Search operations.
```
annzhang-db: I think we can simplify this test file a lot, we generally trust that vector search behavior is semantically correct and we don't need to test that for them. We can have very cursory/rough tests here, maybe just a call or two for each type of index. We can focus more on making sure that our integration is passing through the right things/detecting API signature incompatibilities rather than behavior changes.

tests/integration_tests/vector_search/test_vectorsearch_contract.py
```diff
@@ -0,0 +1,206 @@
...
+)
+
+
+# =============================================================================
+# Import Contract Tests
+# =============================================================================
+
+
+@pytest.mark.contract
+class TestImportContract:
```
annzhang-db: We shouldn't need `None` assertions here, I believe it would fail at import time if the import is gone. These imports should also be covered by other behavior that uses the imports as they would fail if the import fails. It should be okay to remove these.

tests/integration_tests/vector_search/test_vectorsearch_contract.py
```diff
@@ -0,0 +1,206 @@
...
+Contract validation tests for databricks-vectorsearch dependency.
+
+These tests verify the API contract hasn't changed in ways that would break
+databricks-ai-bridge. They check:
+- Import paths exist
+- Method signatures accept expected parameters
+- Response structures contain expected keys
+
+Run daily to catch breaking changes from upstream dependencies early.
+"""
```
annzhang-db: Generally, issues here should also be caught by our more end-to-end integration tests. We can combine some cursory checks here with cursory checks from the behavior test file, and focus more on coverage in e2e integration tests.

integrations/langchain/tests/integration_tests/test_langchain_vectorsearch.py
```diff
@@ -0,0 +1,202 @@
...
+        )
+
+        tool = VectorSearchRetrieverTool(
+            index_name=DELTA_SYNC_INDEX,
+            workspace_client=workspace_client,
+            num_results=2,
+        )
+        result = tool.invoke("technology")
+        assert isinstance(result, str)
+        assert len(result) > 0
```
annzhang-db, line 147: let's assert on the number of results if we know there are documents here, also there isn't a `score_threshold` (defaults to 0.0), so we should always get 2 results.

integrations/langchain/tests/integration_tests/test_langchain_vectorsearch.py
annzhang-db, line 1: Let's add tests for potential `kwargs` getting passed through to vector search

integrations/openai/tests/integration_tests/test_openai_vectorsearch.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+        # strict and additionalProperties should be removed for filter compatibility
+        assert tool.tool is not None
+        assert "strict" not in tool.tool.get("function", {})
+        assert "additionalProperties" not in tool.tool.get("function", {}).get("parameters", {})
+
+
+# =============================================================================
+# VectorSearchRetrieverTool Execution Tests
+# =============================================================================
```
annzhang-db, line 100: Can we add tests with parity to langchain tests for things like filtering?

integrations/openai/tests/integration_tests/test_openai_vectorsearch.py
annzhang-db, line 1: Let's also test kwargs forwarding here, and make sure that kwargs not accepted by vector search client are filtered out.

tests/integration_tests/vector_search/test_vectorsearch_contract.py
dhruv0811: I removed these as a whole

tests/integration_tests/vector_search/test_vectorsearch_contract.py
dhruv0811: Removed this file as a whole

integrations/langchain/src/databricks_langchain/vectorstores.py
```diff
@@ -276,6 +276,7 @@ def __init__(
                     "credential_strategy", CredentialStrategy.MODEL_SERVING_USER_CREDENTIALS
                 )
             elif config.auth_type == "pat":
+                client_args.setdefault("workspace_url", config.host)
```
annzhang-db, line 279: thanks for fixing this!

integrations/langchain/tests/integration_tests/test_langchain_vectorsearch.py
```diff
@@ -0,0 +1,309 @@
...
+            index_name=DELTA_SYNC_INDEX,
+            workspace_client=workspace_client,
+        )
+
+    def test_score_threshold_kwarg_accepted(self, vectorstore):
+        results = vectorstore.similarity_search_with_score(
+            "machine learning", k=3, score_threshold=0.99
+        )
+        # High threshold may return fewer or no results, but should not error
+        assert isinstance(results, list)
```
annzhang-db, line 230: can you validate that this behaves as expected (so that we know it's getting passed through)? you can have the query be something totally random that you know is not in your index and should not return results

tests/integration_tests/conftest.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Create a WorkspaceClient using environment variables.
+
+    The SDK auto-detects auth from env vars (e.g. DATABRICKS_HOST,
+    DATABRICKS_CLIENT_ID, DATABRICKS_CLIENT_SECRET for OAuth M2M).
+    For local testing, set DATABRICKS_CONFIG_PROFILE instead.
+    """
+    from databricks.sdk import WorkspaceClient
+
+    return WorkspaceClient()
```
annzhang-db: Do we want the logic here to be consistent with in the integration tests?
```
 def workspace_client():
      wc = WorkspaceClient()
      if wc.config.auth_type not in ("pat", "oauth-m2m", "model_serving_user_credentials"):
          headers = wc.config.authenticate()
          token = headers.get("Authorization", "").replace("Bearer ", "")
          if token:
              return WorkspaceClient(host=wc.config.host, token=token, auth_type="pat")
      return wc
```

annzhang-db (approved):
approved with a couple small comments! thanks for the work here

dhruv0811:
Addressed comments, final workflow run: 
https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/21969163093

PR #321, opened 2026-02-10
databricks-openai 0.11.0 importing `McpServer` crashes with `NameError` when memory extras aren't installed fix

jennsun:
From traceback:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File ".../databricks_openai/agents/__init__.py", line 2, in <module>
    from databricks_openai.agents.session import AsyncDatabricksSession
  File ".../databricks_openai/agents/session.py", line 55, in <module>
    class AsyncDatabricksSession(SQLAlchemySession):
  File ".../databricks_openai/agents/session.py", line 101, in AsyncDatabricksSession
    token_cache_duration_seconds: int = DEFAULT_TOKEN_CACHE_DURATION_SECONDS,
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
NameError: name 'DEFAULT_TOKEN_CACHE_DURATION_SECONDS' is not defined

due to missing stubs for the constants we defined

```
in session.py (lines 48-52)
-  Added fallback stubs for DEFAULT_TOKEN_CACHE_DURATION_SECONDS and DEFAULT_POOL_RECYCLE_SECONDS in the except ImportError block. These are needed because Python evaluates default parameter values at class definition time which happens unconditionally on import before the runtime _session_imports_available guard in __init__ can protect anything.

in  lakebase.py
- Moved psycopg.rows import DictRow under try catch so it is covered by [memory] extras message. It was an unconditional top-level import that crashed before lakebase.py's own  try/except could produce its helpful error message. Since DictRow is only used in type annotations and the file has from __future__ import  annotations, it's safe under TYPE_CHECKING. wherever sqlalchemy is actually used at runtime, it's imported lazily inside method bodies:

manual testing:
w/o memory extra
```
python3.11 -m venv /tmp/test-no-memory
source /tmp/test-no-memory/bin/activate

pip install -e integrations/openai

python -c "from databricks_openai.agents import McpServer; print('PASS: McpServer imported')" # able to import successfully

python - <<'EOF'
from databricks_openai.agents import AsyncDatabricksSession
try:
    AsyncDatabricksSession(session_id="test", instance_name="test")
except ImportError as e:
    print(f"PASS: got helpful ImportError: {e}") # PASS: got helpful ImportError: AsyncDatabricksSession requires databricks-openai[memory]. Please install with: pip install databricks-openai[memory]
EOF
```
w/ memory extra

```
python3.11 -m venv /tmp/test-with-memory
source /tmp/test-with-memory/bin/activate

pip install -e "integrations/openai[memory]"

python -c "from databricks_openai.agents import McpServer; print('PASS: McpServer imported')"
python -c "from databricks_openai.agents import AsyncDatabricksSession; print('PASS: AsyncDatabricksSession imported')"
```

on master branch without these changes:
```
python -c "from databricks_openai.agents import McpServer; print('PASS: McpServer imported')" # able to import successfully
```
fails with the traceback server error we were seeing:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/jenny.sun/databricks-ai-bridge/integrations/openai/src/databricks_openai/agents/__init__.py", line 2, in <module>
    from databricks_openai.agents.session import AsyncDatabricksSession
  File "/Users/jenny.sun/databricks-ai-bridge/integrations/openai/src/databricks_openai/agents/session.py", line 55, in <module>
    class AsyncDatabricksSession(SQLAlchemySession):
  File "/Users/jenny.sun/databricks-ai-bridge/integrations/openai/src/databricks_openai/agents/session.py", line 101, in AsyncDatabricksSession
    token_cache_duration_seconds: int = DEFAULT_TOKEN_CACHE_DURATION_SECONDS,
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
NameError: name 'DEFAULT_TOKEN_CACHE_DURATION_SECONDS' is not defined
```

### DIFF
--- a/src/databricks_ai_bridge/lakebase.py
+++ b/src/databricks_ai_bridge/lakebase.py
@@ -11,14 +11,13 @@ from typing import TYPE_CHECKING, Any, List, Literal, Optional, Sequence
 from databricks.sdk import WorkspaceClient
 
 if TYPE_CHECKING:
-    from psycopg.rows import DictRow
     from sqlalchemy import URL
     from sqlalchemy.ext.asyncio import AsyncEngine
 
 try:
     import psycopg
     from psycopg import sql
-    from psycopg.rows import dict_row
+    from psycopg.rows import DictRow, dict_row
     from psycopg_pool import AsyncConnectionPool, ConnectionPool
 except ImportError as e:
     raise ImportError(


src/databricks_ai_bridge/lakebase.py
```diff
@@ -18,7 +17,7 @@
 try:
     import psycopg
     from psycopg import sql
-    from psycopg.rows import dict_row
+    from psycopg.rows import DictRow, dict_row
     from psycopg_pool import AsyncConnectionPool, ConnectionPool
```
jennsun, line 21: This import was actually already a top-level import but should move it in here to be safe (was caught by session.py's except ImportError block)

PR #324, opened 2026-02-11
CI Coverage for Base Package and Extra Dependencies to mirror Installation Flow

jennsun:


.github/workflows/main.yml
```diff
@@ -61,9 +61,9 @@ jobs:
           uv run --resolution ${{ matrix.uv-resolution }} \
             --exact --group tests pytest --cov --cov-report=term tests/unit_tests
 
-  lakebase_memory_test:
+  extra_dependency_test:
     runs-on: ubuntu-latest
-    name: "lakebase_memory_test (python: ${{ matrix.python-version }}, uv: ${{ matrix.uv-resolution }})"
+    name: "extra_dependency_test (python: ${{ matrix.python-version }}, uv: ${{ matrix.uv-resolution }})"
     strategy:
```
jennsun, line 89: renaming so all future extra optional dependencies can be added under this test

integrations/openai/tests/unit_tests/test_session.py
```diff
@@ -4,15 +4,13 @@
...
-try:
-    from databricks_ai_bridge.lakebase import _LakebaseBase  # noqa: F401
-    from psycopg.rows import DictRow  # noqa: F401
-    from sqlalchemy.ext.asyncio import AsyncEngine  # noqa: F401
-except ImportError as e:
-    raise ImportError(
-        "AsyncDatabricksSession tests require databricks-openai[memory]. "
-        "Please install with: pip install databricks-openai[memory]"
-    ) from e
+pytest.importorskip("psycopg")
```
bbqiu: this feels a little brittle since the underlying implementation / requirements of the lakebase file may continuously change. what if we just try to import the exact stuff we need in this file?

ex.
```
_LakebaseBase = pytest.importorskip(databricks_ai_bridge.lakebase_LakebaseBase)
...
```

(from AI) assigning it to the imported module apparently can make linters mad, so importing it again may be necessary

integrations/openai/tests/unit_tests/test_session.py
```diff
@@ -4,15 +4,13 @@
...
+pytest.importorskip("psycopg_pool")
+pytest.importorskip("sqlalchemy")
+
+from databricks_ai_bridge.lakebase import _LakebaseBase  # noqa: E402, F401
```
bbqiu: do we still need these linter suppression comments

jennsun: added due to these ruff check failures: https://github.com/databricks/databricks-ai-bridge/actions/runs/21920027776/job/63297416711?pr=324

but I think we can actually remove them since our pytest.importorskip line is the only guard needed so no longer need these :)

integrations/openai/tests/unit_tests/test_session.py
jennsun: updated to skip:
```pytest.importorskip("databricks_ai_bridge.lakebase")```

File: integrations/langchain/tests/unit_tests/test_checkpoint.py [hunk 4]
```diff
@@ -4,8 +4,7 @@
 
 import pytest
 
-pytest.importorskip("psycopg")
-pytest.importorskip("psycopg_pool")
+pytest.importorskip("databricks_ai_bridge.lakebase")
```
bbqiu, line 7: can we do the same for all the stuff we're importing that has extra dependencies? ex. importing async checkpoint saver / checkpoint saver

bbqiu: ditto for every other file -- then the test file stands independnetly, and doesn't have to change dependent on if the lakebase file itself stops using psycopg under the hood etc.

jennsun: the issue is that if we do pyskip for `from databricks_langchain import AsyncCheckpointSaver, CheckpointSaver` it will succeed because we swallow the import error and stub out the classes so the import will always be successful

PR #327, opened 2026-02-13
Generic Python Lakebase Checkpointing

coltonpeltier-db:
# What
Added classes and functions to handle generic agent checkpointing with lakebase instances.

# Why
Existing checkpoint implementations depend on langchain / langgraph

# What Changed
Added `lakebase_checkpointer.py` file which contains two key classes:
- `GenericCheckpoint` - class that defines the checkpoints schema. Can be subclassed to change checkpoint schema.
- `Lakebase_Checkpointer` - accepts a `LakebaseClient` and the checkpoint class as input. Implements helper functions for interacting with `LakebaseClient` with the checkpoint class (saves, inserts, updates, get).

### DIFF
--- a/src/databricks_ai_bridge/lakebase_checkpointer.py
+++ b/src/databricks_ai_bridge/lakebase_checkpointer.py
@@ -1,4 +1,4 @@
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 from datetime import datetime
 from .lakebase import LakebaseClient
 from typing import Tuple
@@ -8,12 +8,26 @@ from typing import Any, Union, Tuple
 from uuid import uuid4
 
 class GenericCheckpoint(BaseModel):
-    # Define the fields for your checkpoint
-    id : str = str(uuid4())
+    ####################
+    # CHECKPOINT COLUMNS
+    ####################
+    id : str = Field(default_factory = lambda: str(uuid4())) # `id` column must exist for locating the checkpoint, even in subclasses.
     state : dict = {}
-    creation_timestamp : datetime = datetime.now()
-    update_timestamp : datetime = datetime.now()
-    # Implement abstract methods
+    creation_timestamp : datetime = Field(default_factory=datetime.now)
+    update_timestamp : datetime = Field(default_factory=datetime.now)
+    ####################
+    # UPDATE ATTRIBUTES
+    ####################
+    def update(self, **kwargs):
+        for k,v in kwargs.items():
+            # Check attribute exists
+            assert hasattr(self, k), f"Attribute {k} does not exist in {self.__class__.__name__}"
+            setattr(self, k, v)
+        self.update_timestamp = datetime.now()
+    #################################
+    # SQL GENERATION IMPLEMENTATIONS
+    #################################
+    # If subclassing the `GenericCheckpoint` class, implement new methods to handle any changes in attributes.
     def generate_insert_sql(self, table_name : str) -> Tuple[str, tuple]:
         sql = f"""
             INSERT INTO {table_name}
@@ -44,12 +58,18 @@ class GenericCheckpoint(BaseModel):
         return sql, None
     
     def generate_retrieve_checkpoint_sql(self, table_name : str) -> Tuple[str, tuple]:
+        sql_all, params = self.generate_retrieval_all_checkpoints_sql(table_name = table_name)
+        sql = f"""
+            {sql_all} LIMIT 1
+        """
+        return sql, params
+    
+    def generate_retrieval_all_checkpoints_sql(self, table_name : str) -> Tuple[str, tuple]:
         sql = f"""
             SELECT id, state, creation_timestamp, update_timestamp
             FROM {table_name}
             WHERE id = %s
             ORDER BY update_timestamp DESC
-            LIMIT 1
         """
         return sql, (self.id,)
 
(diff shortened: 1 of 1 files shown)


src/databricks_ai_bridge/lakebase_checkpointer.py
```diff
@@ -0,0 +1,141 @@
...
+        sql = f"""
+            INSERT INTO {table_name}
+            (id, state, creation_timestamp, update_timestamp)
+            VALUES (%s, %s, %s, %s)
+        """
+        return sql, (self.id, Json(self.state), self.creation_timestamp, self.update_timestamp)
+    
+    def generate_update_sql(self, table_name : str) -> Tuple[str, tuple]:
+        sql = f"""
+            UPDATE {table_name}
```
jennsun, line 41: every sql method is using direct string interpolation, which puts us at risk of sql injection: can you take a look at the lakebaseclient implementation: https://github.com/databricks/databricks-ai-bridge/blob/main/src/databricks_ai_bridge/lakebase.py

which uses sql identifiers: https://www.psycopg.org/psycopg3/docs/api/sql.html

src/databricks_ai_bridge/lakebase_checkpointer.py
```diff
@@ -0,0 +1,141 @@
...
+        )
+        sql, params = _checkpoint.generate_insert_sql(table_name = self.table_name)
+        response = self.client.execute(sql=sql, params=params)
+        return
+
+    def save_checkpoint(self, id : str, overwrite : bool = False, **checkpoint_kwargs) -> None:
+        if overwrite:
+            checkpoint_exists = self.checkpoint_exists(id = id)
+            if checkpoint_exists:
+                self.update_checkpoint(id = id, **checkpoint_kwargs)
```
jennsun, line 136: update_checkpoint is a nonexistent function in this class -- is this a typo (should it be self.update_most_recent_checkpoint)?

src/databricks_ai_bridge/lakebase_checkpointer.py
```diff
@@ -0,0 +1,141 @@
...
+    def generate_retrieval_all_checkpoints_sql(self, table_name : str) -> Tuple[str, tuple]:
+        sql = f"""
+            SELECT id, state, creation_timestamp, update_timestamp
+            FROM {table_name}
+            WHERE id = %s
+            ORDER BY update_timestamp DESC
+        """
+        return sql, (self.id,)
+
+class LakebaseCheckpointer:
```
jennsun, line 76: Can we implement context manager support to mirror our existing checkpointsaver class? https://github.com/databricks/databricks-ai-bridge/blob/main/integrations/langchain/src/databricks_langchain/checkpoint.py#L48

src/databricks_ai_bridge/lakebase_checkpointer.py
```diff
@@ -0,0 +1,141 @@
...
+    def __init__(self, 
+                 lakebase_client : LakebaseClient, 
+                 sessions_table_name : str, 
+                 checkpoint_class : GenericCheckpoint = GenericCheckpoint):
+        self.checkpoint_class = checkpoint_class
+        self.client = lakebase_client
+        self.table_name = sessions_table_name
+        self.init_schema()
```
jennsun, line 84: `init_schema` is being called automatically which then attempts to [create a table each time](https://github.com/databricks/databricks-ai-bridge/pull/327/changes#diff-4d2f44fd406f48ff13422612b8a6ca347f438217a6afa4fa86c26da11fbc694bR49) 

can you add a create_tables bool (defaulting to true) to mirror our openai session class where we create tables on init: https://github.com/databricks/databricks-ai-bridge/blob/main/integrations/openai/src/databricks_openai/agents/session.py#L120

src/databricks_ai_bridge/lakebase_checkpointer.py
```diff
@@ -0,0 +1,141 @@
...
+            WHERE id = %s
+        """
+        response = self.client.execute(sql=sql, params=(id,))
+        count = response[0]["count"]
+        return True if count > 0 else False
+    
+    def update_most_recent_checkpoint(self, id : str, **checkpoint_kwargs) -> None:
+        # Get the most recent checkpoint
+        _checkpoint = self.get_most_recent_checkpoint(id = id)
+        _checkpoint.update(**checkpoint_kwargs)
```
jennsun, line 119: get_most_recent_checkpoint returns None when no recent checkpoint exists -- can we handle this case?

src/databricks_ai_bridge/lakebase_checkpointer.py
```diff
@@ -0,0 +1,141 @@
...
+    state : dict = {}
+    creation_timestamp : datetime = Field(default_factory=datetime.now)
+    update_timestamp : datetime = Field(default_factory=datetime.now)
+    ####################
+    # UPDATE ATTRIBUTES
+    ####################
+    def update(self, **kwargs):
+        for k,v in kwargs.items():
+            # Check attribute exists
+            assert hasattr(self, k), f"Attribute {k} does not exist in {self.__class__.__name__}"
```
jennsun, line 24: instead of assert, can we raise a ValueError or AttributeError instead? assert statements are stripped when python runs with optimization requested (command line option [-O](https://docs.python.org/3/using/cmdline.html#cmdoption-O)).

https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

PR #328, opened 2026-02-14
Add MCP Integration Tests

dhruv0811:
## Summary

Adds MCP integration tests covering the core `DatabricksMCPClient`, OpenAI wrappers (`McpServerToolkit` + `McpServer`), and LangChain wrappers (`DatabricksMCPServer` + `DatabricksMultiServerMCPClient`). Tests validate against a live Databricks MCP server across **UC Functions**, **Vector Search**, and **Genie**.

**64 integration tests** (21 core + 25 OpenAI + 18 LangChain), gated behind `RUN_MCP_INTEGRATION_TESTS=1`.

### What's tested

- Tool listing and execution for UC Functions, Vector Search, and Genie
- Schema-level listing (UC + VS)
- Multi-server client with UC + VS servers loaded together
- PAT and OAuth-M2M auth paths
- `get_databricks_resources()` normalization and type correctness
- `handle_tool_error`, `timeout` kwargs pass-through (LangChain)
- MLflow tracing, timeout defaults/override (OpenAI Agents)
- Auth provider freshness per request (LangChain factory)
- Error paths: nonexistent function, nonexistent tool name, wrong arguments — across all 3 layers

## Test plan
- [x] All 64 MCP integration tests pass (21 core + 25 OpenAI + 18 LangChain)
- [x] Existing unit tests unaffected
- [x] CI Run: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/22639223391

### DIFF
--- a/databricks_mcp/tests/integration_tests/conftest.py
+++ b/databricks_mcp/tests/integration_tests/conftest.py
@@ -5,7 +5,7 @@ These tests require a live Databricks workspace with a pre-created UC function.
 They are NOT run by default — set RUN_MCP_INTEGRATION_TESTS=1 to enable.
 
 Prerequisites:
-    Run setup_workspace.py once to create the test UC function.
+    The test UC function must exist in the workspace.
 
 Environment Variables:
 ======================
@@ -21,20 +21,35 @@ from __future__ import annotations
 import os
 
 import pytest
+from mcp.shared.exceptions import McpError
 
 from databricks_mcp import DatabricksMCPClient
 
-# =============================================================================
-# Env var gate — skip entire module if not enabled
-# =============================================================================
 
-pytestmark = pytest.mark.skipif(
-    os.environ.get("RUN_MCP_INTEGRATION_TESTS") != "1",
-    reason="MCP integration tests disabled. Set RUN_MCP_INTEGRATION_TESTS=1 to enable.",
-)
+def _find_mcp_error(exc_group: ExceptionGroup) -> McpError | None:  # ty: ignore[unresolved-reference]
+    """Recursively unwrap nested ExceptionGroups to find a McpError."""
+    for exc in exc_group.exceptions:
+        if isinstance(exc, McpError):
+            return exc
+        if isinstance(exc, ExceptionGroup):  # ty: ignore[unresolved-reference]
+            found = _find_mcp_error(exc)
+            if found:
+                return found
+    return None
+
+
+def _skip_if_not_found(exc_group: ExceptionGroup, context: str) -> None:  # ty: ignore[unresolved-reference]
+    """Skip the test if the McpError indicates a missing resource, otherwise re-raise."""
+    mcp_error = _find_mcp_error(exc_group)
+    if mcp_error:
+        msg = str(mcp_error)
+        if "NOT_FOUND" in msg or "not found" in msg.lower():
+            pytest.skip(f"{context}: {mcp_error}")
+    raise exc_group
+
 
 # =============================================================================
-# Constants (must match setup_workspace.py)
+# Constants
 # =============================================================================
 
 CATALOG = "integration_testing"
@@ -97,11 +112,9 @@ def cached_tools_list(mcp_client):
     """
     try:
         tools = mcp_client.list_tools()
-    except Exception as e:
-        pytest.skip(
-            f"Could not list tools from MCP server — is the test function set up? "
-            f"Run setup_workspace.py first. Error: {e}"
-        )
+    except ExceptionGroup as e:  # ty: ignore[unresolved-reference]
+        _skip_if_not_found(e, "UC function not found in workspace")
+    assert tools, "list_tools() returned no tools — is the test function set up?"
     return tools
 
 
@@ -151,9 +164,11 @@ def vs_schema_mcp_client(vs_schema_mcp_url, workspace_client):
 def cached_vs_tools_list(vs_mcp_client):
     """Cache the VS list_tools() result; skip if VS MCP endpoint unavailable."""
     try:
-        return vs_mcp_client.list_tools()
-    except Exception as e:
-        pytest.skip(f"VS MCP endpoint not available: {e}")
+        tools = vs_mcp_client.list_tools()
+    except ExceptionGroup as e:  # ty: ignore[unresolved-reference]
+        _skip_if_not_found(e, "VS MCP endpoint not available in workspace")
+    assert tools, "VS list_tools() returned no tools — is the VS index set up?"
+    return tools
 
 
 @pytest.fixture(scope="session")
@@ -196,9 +211,11 @@ def genie_mcp_client(genie_mcp_url, workspace_client):
 def cached_genie_tools_list(genie_mcp_client):
     """Cache the Genie list_tools() result; skip if Genie MCP endpoint unavailable."""
     try:
-        return genie_mcp_client.list_tools()
-    except Exception as e:
-        pytest.skip(f"Genie MCP endpoint not available: {e}")
+        tools = genie_mcp_client.list_tools()
+    except ExceptionGroup as e:  # ty: ignore[unresolved-reference]
+        _skip_if_not_found(e, "Genie MCP endpoint not available in workspace")
+    assert tools, "Genie list_tools() returned no tools — is the Genie space set up?"
+    return tools
 
 
 @pytest.fixture(scope="session")
--- a/databricks_mcp/tests/integration_tests/test_mcp_core.py
+++ b/databricks_mcp/tests/integration_tests/test_mcp_core.py
@@ -10,6 +10,8 @@ from __future__ import annotations
 import os
 
 import pytest
+from conftest import _skip_if_not_found
+from mcp.shared.exceptions import McpError
 from mcp.types import CallToolResult
 
 from databricks_mcp import DatabricksMCPClient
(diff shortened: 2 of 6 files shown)


integrations/langchain/src/databricks_langchain/chat_models.py
annzhang-db, line 1: why are we removing `ty:ignore[unresolved-attribute]: astral-sh/ty#1479 should fix this` in favor of `ty:ignore` in this file? if there is a version issue, can we pin the version of `ty` instead of broadening the suppression?

src/databricks_ai_bridge/lakebase.py
```diff
@@ -223,7 +223,7 @@ def connect(cls, conninfo: str = "", **kwargs):
         max_size = pool_kwargs.pop("max_size", DEFAULT_MAX_SIZE)
         timeout = pool_kwargs.pop("timeout", DEFAULT_TIMEOUT)
 
-        self._pool: ConnectionPool[psycopg.Connection[DictRow]] = ConnectionPool(  # type: ignore[invalid-assignment]
+        self._pool: ConnectionPool[psycopg.Connection[DictRow]] = ConnectionPool(  # ty:ignore[invalid-assignment]
             conninfo=self._conninfo(),
             kwargs=default_kwargs,
             min_size=min_size,  # type: ignore[invalid-argument-type]
```
annzhang-db, line 229: we are using `type` here and `ty` above, not sure exactly the typing issues you ran into but can we be consistent across the file?

databricks_mcp/tests/integration_tests/conftest.py
```diff
@@ -0,0 +1,222 @@
...
+
+import pytest
+
+from databricks_mcp import DatabricksMCPClient
+
+# =============================================================================
+# Env var gate — skip entire module if not enabled
+# =============================================================================
+
+pytestmark = pytest.mark.skipif(
```
annzhang-db: from what i've looked at, this will only skip **tests** in this file, of which there are none, so it probably isn't guarding anything and we don't need it. [pytest reference](https://docs.pytest.org/en/stable/how-to/skipping.html#skip-all-test-functions-of-a-class-or-module) for reference

integrations/langchain/tests/integration_tests/test_langchain_mcp.py
```diff
@@ -0,0 +1,451 @@
+"""
+Integration tests for LangChain MCP wrappers.
+
+Tests DatabricksMCPServer, DatabricksMultiServerMCPClient against a live
```
annzhang-db, line 4: Can we try testing `DatabricksMultiServerMCPClient` with multiple servers instead of always passing through one server?

integrations/langchain/tests/integration_tests/test_langchain_mcp.py
```diff
@@ -0,0 +1,451 @@
...
+# DatabricksMcpHttpClientFactory Token Refresh Tests
+# =============================================================================
+
+
+@pytest.mark.integration
+class TestDatabricksMcpHttpClientFactoryTokenRefresh:
+    """Verify DatabricksMcpHttpClientFactory re-creates auth providers for token refresh."""
+
+    def test_sequential_invocations_succeed(self, workspace_client):
+        """Two sequential get_tools() calls succeed — proves token refresh works."""
```
annzhang-db: Would the two sequential `get_tools()` calls not use the same token? Curious how token refresh is triggered here

databricks_mcp/tests/integration_tests/conftest.py
```diff
@@ -0,0 +1,222 @@
...
+@pytest.fixture(scope="session")
+def cached_tools_list(mcp_client):
+    """
+    Cache the list_tools() result for the session to minimize API calls.
+
+    Skips all dependent tests if the function doesn't exist.
+    """
+    try:
+        tools = mcp_client.list_tools()
+    except Exception as e:
```
annzhang-db: I'm a little worried on how we will catch issues with listing tools using the mcp client if we are catching general errors here, seems to be casting a wider net than just if the function is not set up. We can check what the behavior is when we use a non-existent function and catch that specifically.

databricks_mcp/tests/integration_tests/conftest.py
dhruv0811: Youre right, this gate is redundant, removed it.

integrations/langchain/src/databricks_langchain/chat_models.py
dhruv0811: Pulled from main and seems the ty issues are gone now, undid these changes!

databricks_mcp/tests/integration_tests/conftest.py
dhruv0811: Did a small experiment, the error thrown is an `McpError` with a bad request saying the function was `not found`. Modified the conftest file to catch only this error, validate its a not found error based on the message, otherwise raise the error.

integrations/langchain/tests/integration_tests/test_langchain_mcp.py
```diff
@@ -394,6 +394,145 @@ async def _test():
...
+        )
+
+        async def _test():
+            client = DatabricksMultiServerMCPClient([server])
+            await client.get_tools()
+
+        with pytest.raises(ExceptionGroup) as exc_info:
+            asyncio.run(_test())
+
+        # Unwrap nested ExceptionGroups to find McpError
```
dhruv0811: Note: The pytest match isnt working here since this Error is nested it seems, so unwrapping it a bit.

integrations/langchain/tests/integration_tests/test_langchain_mcp.py
dhruv0811: I believe the second call gets a new token since the factory creates a new OAuth provider. But youre right that the test doesnt really prove this currently. Will think more about how we can improve the auth token tests. The second test here checks that behavior for new provider, that should be sufficient?

integrations/langchain/tests/integration_tests/test_langchain_mcp.py
```diff
@@ -414,35 +519,55 @@ def test_sequential_invocations_succeed(self, workspace_client):
...
+            assert len(tools) > 0
+            await tools[0].ainvoke({"wrong_arg": "test"})
 
-        asyncio.run(_test())
+        with pytest.raises(McpError, match="BAD_REQUEST"):
+            asyncio.run(_test())
+
+
+# =============================================================================
+# DatabricksMcpHttpClientFactory Auth Test
```
dhruv0811, line 511: This test doesn't replicate what a user would do (they would never directly import and use `DatabricksMcpHttpClientFactory`), but this test still verifies that each `get_tools()` call under the hood will create a new OAuth provider (factory behavior).

databricks_mcp/tests/integration_tests/conftest.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+@pytest.fixture(scope="session")
+def cached_call_result(mcp_client, cached_tools_list):
+    """
+    Cache a call_tool() result for the session.
+
+    Uses the first tool from cached_tools_list.
+    """
+    tool_name = cached_tools_list[0].name
```
annzhang-db, line 128: minor: should we check the length of `cached_tools_list` to make sure that `list_tools` returned a non-empty list? if the behavior is that `list_tools` always throws an exception instead of returning an empty list for invalid functions, you can ignore this :) would also be curious what `cached_tools_list` is if it throws an exception that is not caught, and we are returning `tools` (might be None)?

databricks_mcp/tests/integration_tests/test_mcp_core.py
```diff
@@ -0,0 +1,250 @@
...
+        tool_names = [t.name for t in cached_tools_list]
+        assert any(FUNCTION_NAME in name or "echo_message" in name for name in tool_names), (
+            f"Expected a tool name containing '{FUNCTION_NAME}', got: {tool_names}"
+        )
+
+    def test_schema_level_lists_multiple_tools(self, schema_mcp_client):
+        """Schema-level URL should list at least the echo_message function."""
+        try:
+            tools = schema_mcp_client.list_tools()
+        except Exception as e:
```
annzhang-db: curious why we use `Exception` here but `ExceptionGroup` for listing specific UC functions, let's try to catch as specific of exceptions as possible when we skip so that the test fails if there are some new failure modes for `list_tools`

integrations/openai/tests/integration_tests/test_openai_mcp.py
```diff
@@ -0,0 +1,613 @@
...
+# =============================================================================
+
+
+@pytest.mark.integration
+class TestMcpServerAgentsUCFunctions:
+    """Test McpServer (Agents SDK) with UC function MCP endpoints (init, exec, auth, tracing, timeout)."""
+
+    # -- Listing --
+
+    def test_server_lists_tools(self, workspace_client):
```
annzhang-db: can we use `@pytest.mark.asyncio` and `async def test_server_lists_tools` instead of `asyncio.run(_test())`? that way pytest-asyncio can handle the event loop for us. i believe we already have this pattern in other tests, so you can refer to those also! would be curious if there was a particular reason why we chose this pattern instead

databricks_mcp/tests/integration_tests/conftest.py
dhruv0811: Good catch! I added an assert to make sure both the None case and the empty tools case gets caught and we get a clear assertion failure.

databricks_mcp/tests/integration_tests/test_mcp_core.py
dhruv0811: Replaced all the `Exception` catches with `ExceptionGroup` + skip if the particular error is not found.

integrations/openai/tests/integration_tests/test_openai_mcp.py
dhruv0811: Refactored to use `@pytest.mark.asyncio` instead! Thanks for this catch, the tests look a lot cleaner with this.

dhruv0811:
Addressed comments, re-ran CI: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/22639223391

PR #331, opened 2026-02-17
Fixing spans to use mlflow lazy imports and provide correct args for tracing

sunishsheth2009:
Fixing spans to use mlflow lazy imports and provide correct args for tracing

### DIFF
--- a/src/databricks_ai_bridge/_compat.py
+++ b/src/databricks_ai_bridge/_compat.py
@@ -1,0 +1,48 @@
+"""Compatibility module for handling optional dependencies."""
+
+from typing import Callable, Optional, TypeVar
+
+F = TypeVar("F", bound=Callable)
+
+
+def mlflow_trace(
+    func: Optional[F] = None,
+    *,
+    name: Optional[str] = None,
+    span_type: Optional[str] = None,
+) -> F:
+    """Decorator that traces a function with mlflow if available.
+
+    If mlflow is not installed, the function executes normally without tracing.
+
+    Args:
+        func: Function to trace
+        name: Optional name for the span (defaults to function name)
+        span_type: Optional span type (e.g., "CHAIN", "PARSER", "AGENT")
+
+    Returns:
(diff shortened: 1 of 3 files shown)


src/databricks_ai_bridge/_compat.py
```diff
@@ -0,0 +1,30 @@
...
+        @mlflow_trace
+        def my_function():
+            # Function logic
+            pass
+    """
+    try:
+        import mlflow
+
+        return mlflow.trace(func)  # type: ignore[return-value]
+    except ImportError:
```
dhruv0811: nit: Is it worth logging a warning message that mlflow is not installed so traces will not be available?

src/databricks_ai_bridge/genie.py
```diff
@@ -38,95 +40,93 @@ class GenieResponse:
     text_attachment_content: Optional[str] = ""
 
 
+@mlflow_trace
```
dhruv0811: Are missing the span_type information here?
`@mlflow.trace(span_type="PARSER")`

Might also need to update the decorator to accept the span_type param if we decide to keep this information.

src/databricks_ai_bridge/_compat.py
sunishsheth2009: I removed the warning since every @mlflow_trace would throw that warning. So adding using genie integration would throw 5-6 warnings which would create noise. So removed that. In the previous solution there was no warning as well. So we can follow the same?

src/databricks_ai_bridge/genie.py
sunishsheth2009: Good catch, resolved it

aravind-segu (approved):
LGTM! Can we do a quick E2E test in a notebook to make sure its tracing as expected?

For future reference we are not using the mlflow trace decorator directly due to this issue: https://github.com/databricks/databricks-ai-bridge/issues/284

PR #332, opened 2026-02-19
feat(ai-sdk-provider): add trace_id capture and databricks_options forwarding

smurching:
Support passing additional parameters to get back a trace ID from an agent serving endpoint or Databricks app:
* For agent serving endpoints, we augment the request payload with `{"databricks_options": {"return_trace": true}}`
* For agents on Databricks apps that use MLflow agent server, we pass an `x-mlflow-return-trace-id` header. This doesn't require any changes to the AI SDK provider, since it already supports custom headers

We expose the resulting trace via the `response.body` for streaming and non-streaming queries

### DIFF
--- a/integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
+++ b/integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
@@ -54,7 +54,6 @@ function mapResponsesFinishReason({
   return { raw: finishReason ?? undefined, unified }
 }
 
-
 export class DatabricksResponsesAgentLanguageModel implements LanguageModelV3 {
   readonly specificationVersion = 'v3'
 
--- a/integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-schema.ts
+++ b/integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-schema.ts
@@ -80,31 +80,32 @@ const responsesAgentOutputItem = z.discriminatedUnion('type', [
   responsesAgentMcpApprovalResponseSchema,
 ])
 
-export const responsesAgentResponseSchema = z.object({
-  id: z.string().optional(),
-  created_at: z.number().optional(),
-  error: z
-    .object({
-      code: z.string(),
-      message: z.string(),
-    })
-    .nullish(),
-  model: z.string().optional(),
-  output: z.array(responsesAgentOutputItem),
-  incomplete_details: z
-    .object({
-      reason: z.string().nullish().optional(),
-    })
-    .nullish(),
-  usage: z
-    .object({
-      input_tokens: z.number(),
-      output_tokens: z.number(),
-      total_tokens: z.number(),
-    })
-    .optional(),
-})
-.passthrough()
+export const responsesAgentResponseSchema = z
+  .object({
+    id: z.string().optional(),
+    created_at: z.number().optional(),
+    error: z
+      .object({
+        code: z.string(),
+        message: z.string(),
+      })
+      .nullish(),
+    model: z.string().optional(),
+    output: z.array(responsesAgentOutputItem),
+    incomplete_details: z
+      .object({
+        reason: z.string().nullish().optional(),
+      })
+      .nullish(),
+    usage: z
+      .object({
+        input_tokens: z.number(),
+        output_tokens: z.number(),
+        total_tokens: z.number(),
+      })
+      .optional(),
+  })
+  .passthrough()
 
 /**
  * Chunk schema


integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -189,9 +223,43 @@ export class DatabricksResponsesAgentLanguageModel implements LanguageModelV3 {
                 })
                 usage.inputTokens.total = chunk.value.response.usage.input_tokens
                 usage.outputTokens.total = chunk.value.response.usage.output_tokens
+                // Capture trace_id and span_id in the responseBody object
+                if (chunk.value.response.trace_id !== undefined) {
+                  responseBody.trace_id = chunk.value.response.trace_id
+                }
+                if (chunk.value.response.span_id !== undefined) {
+                  responseBody.span_id = chunk.value.response.span_id
+                }
```
smurching: I don't think `span_id` is ever used

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -95,6 +96,32 @@ export class DatabricksResponsesAgentLanguageModel implements LanguageModelV3 {
     const content = convertResponsesAgentResponseToMessagePart(response)
     const hasToolCalls = content.some((p) => p.type === 'tool-call')
 
+    // Extract trace_id from both root level (legacy) and nested databricks_output structure
```
smurching: `trace_id` at the root level can happen for streaming chunks from MLflow agent server, per https://github.com/mlflow/mlflow/pull/20387. It's not legacy

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -95,6 +110,22 @@ export class DatabricksResponsesAgentLanguageModel implements LanguageModelV3 {
     const content = convertResponsesAgentResponseToMessagePart(response)
     const hasToolCalls = content.some((p) => p.type === 'tool-call')
 
+    // Extract trace_id from both root level (legacy) and nested databricks_output structure.
```
smurching: Don't think this is legacy, it can happen when streaming MLflow agent server chunks per https://github.com/mlflow/mlflow/pull/20387

integrations/ai-sdk-provider/src/stream-transformers/databricks-stream-transformer.ts
```diff
@@ -17,6 +17,13 @@ export const getDatabricksLanguageModelTransformStream = () => {
   )
   return new TransformStream<LanguageModelV3StreamPart, LanguageModelV3StreamPart>({
     transform(chunk, controller) {
+      // Raw chunks (from includeRawChunks) must be transparent to delta boundary tracking.
```
smurching, line 20: AI explanation:


 What are raw chunks?                                                                                                                                     
                                                                                                                                                           
  When you call streamText with includeRawChunks: true, the AI SDK wraps each raw SSE event from the upstream provider as an extra chunk of type 'raw' and 
  injects it into the stream alongside the semantic chunks it would produce normally. So for a single Responses API SSE event like                       
  response.output_item.done, you might see both:
  - { type: 'raw', rawValue: { type: 'response.output_item.done', databricks_output: ... } } — the raw event verbatim
  - { type: 'text-delta', delta: "It's just blue duh!" } — the semantic chunk the SDK derived from it

  They travel together through the stream. We use includeRawChunks: true specifically because raw chunks carry the databricks_output.trace.info.trace_id
  field that the SDK doesn't expose through its normal semantic API.

  Why pass them through the delta boundary transformer unchanged?

  The delta boundary transformer in databricks-stream-transformer.ts tracks lastChunk to determine when to inject synthetic text-start and text-end events.
   Its logic is: "if the last chunk was a text-delta and the next chunk is not, inject a text-end first."

  The problem is that raw chunks are interleaved between semantic chunks. Without the bypass, a raw chunk arriving between two text-delta chunks would look
   like a "non-delta" chunk to the transformer, causing it to prematurely inject a text-end after the raw event — and then a fresh text-start for the next
  delta. This breaks the stream: the deduplication logic in the transformer would then drop all subsequent text-delta chunks with the same id as
  already-ended, resulting in an empty message in the UI. That was the actual bug text-streaming.test.ts was written to catch.

  By skipping raw chunks entirely (passing them through and not updating lastChunk), the transformer stays blind to them — it only tracks the semantic
  delta sequence, which remains contiguous.

integrations/ai-sdk-provider/tsdown.config.ts
```diff
@@ -9,6 +9,9 @@ export default defineConfig({
   // Output formats
   format: ['esm', 'cjs'],
 
+  // Use explicit extensions (.mjs/.cjs) for compatibility
```
smurching: why is this necessary?

smurching: AI explanation:

Not a breaking change — it actually fixes a pre-existing mismatch.                                                                                       
                                                                                                                                                           
  package.json already declares:                                                                                                                           
                                                                                    
  "exports": {                                                                                                                                             
    ".": {                                                                                                                                                 
      "import": { "default": "./dist/index.mjs" },
      "require": { "default": "./dist/index.cjs" }
    }
  }

  Without fixedExtension: true, tsdown respects the package's "type": "module" and outputs index.js (not index.mjs) for the ESM format. That means the
  "import" export condition was pointing at a file that didn't exist — the ESM path was broken for any consumer using the exports map.

  Adding fixedExtension: true makes tsdown emit index.mjs + index.cjs, which is exactly what the exports map already promises. So it's correcting the build
   to match the already-declared public interface, not changing that interface.

integrations/ai-sdk-provider/src/responses-agent-language-model/call-options-to-responses-args.ts
```diff
@@ -30,6 +33,9 @@ export type DatabricksProviderOptions = {
   reasoning?: {
     effort?: 'low' | 'medium' | 'high'
   }
+  databricksOptions?: {
+    return_trace?: boolean
+  }
 }
```
fanzeyi, line 37: It looks like this whole type is already Databricks provider specific. Do we need to introduce another level of `databricksOptions`? Can we just directly add a `return_trace`?

Also bikeshedding a bit, would `include_trace` or even `trace` be a better name here?

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -54,6 +55,20 @@ function mapResponsesFinishReason({
...
+ */
+function extractTraceIdFromDatabricksOutput(databricksOutput: unknown): string | undefined {
+  if (!databricksOutput || typeof databricksOutput !== 'object') return undefined
+  const trace = (databricksOutput as { trace?: unknown }).trace
+  if (!trace || typeof trace !== 'object') return undefined
+  const info = (trace as { info?: unknown }).info
+  if (!info || typeof info !== 'object') return undefined
+  const traceId = (info as { trace_id?: unknown }).trace_id
+  return typeof traceId === 'string' ? traceId : undefined
+}
```
fanzeyi: Use zod here?


```suggestion
 */
const databricksTraceIdZodSchema = z.object({
  trace: z.object({
    info: z.object({
      trace_id: z.string().optional(),
    }).optional(),
  }).optional(),
})

function extractTraceIdFromDatabricksOutput(databricksOutput: unknown): string | undefined {
  const parsed = databricksTraceIdZodSchema.safeParse(databricksOutput)
  return parsed.success ? parsed.data.trace?.info?.trace_id : undefined
}```

File: integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-schema.ts [hunk 8]
```diff
@@ -101,6 +103,11 @@ export const responsesAgentResponseSchema = z.object({
       total_tokens: z.number(),
     })
     .optional(),
+  trace_id: z.string().optional(),
+  span_id: z.string().optional(),
+  // Databricks-specific trace output - use z.any() to accept any structure
+  // We'll extract trace_id in the code rather than validating the full structure
+  databricks_output: z.any().optional(),
```
fanzeyi: hmm why can't we define `trace_id` here and make it optional? I think this will remove the need to do type narrowing in my other comment.

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -111,6 +141,9 @@ export class DatabricksResponsesAgentLanguageModel implements LanguageModelV3 {
         outputTokens: { total: response.usage?.output_tokens ?? 0, text: 0, reasoning: 0 },
       },
       warnings,
+      response: {
+        body: responseBody,
+      } as { headers?: SharedV3Headers; body?: unknown },
```
smurching: `body` is defined by AI SDK output here: https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-text#response.language-model-response-metadata.body; it's the main way/escape hatch to pass through non-standard LLM output in the AI SDK

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -158,6 +191,11 @@ export class DatabricksResponsesAgentLanguageModel implements LanguageModelV3 {
     // Track tool call IDs to tool names for looking up tool names in function_call_output events
     const toolNamesByCallId = new Map<string, string>()
 
+    // Lazily populated as the stream is consumed — only valid after the stream is fully drained.
+    // trace_id may be populated from either the response.output_item.done or responses.completed event,
+    // whichever arrives first. span_id is only available in responses.completed.
+    const responseBody: Record<string, unknown> = {}
```
smurching: Why not just let the streaming path parse the trace ID out from the appropriate streaming event?

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -189,9 +227,29 @@ export class DatabricksResponsesAgentLanguageModel implements LanguageModelV3 {
                 })
                 usage.inputTokens.total = chunk.value.response.usage.input_tokens
                 usage.outputTokens.total = chunk.value.response.usage.output_tokens
+                // Capture trace_id in the responseBody object
```
smurching: When consuming streaming output, I think we can rely on callers to just look for trace_id in the appropriate streaming event, by calling streamText with includeRawChunks: true per https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#include-raw-chunks

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-schema.ts
```diff
@@ -12,17 +12,19 @@ const responsesAgentMessageSchema = z.object({
...
+            z.object({
+              type: z.literal('url_citation'),
+              start_index: z.number(),
+              end_index: z.number(),
+              url: z.string(),
+              title: z.string(),
+            }),
+          ])
+        )
+        .optional(),
```
smurching, line 27: Annotations should be optional in the response (not every agent will include an `annotations` field in assistant response message content)

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-schema.ts
```diff
@@ -102,6 +104,7 @@ export const responsesAgentResponseSchema = z.object({
     })
     .optional(),
 })
+.passthrough()
```
smurching: Needed to ensure that we don't filter out extra Databricks-specific fields in the output schema, like `databricks_output` (in its current & future forms, with any fields that are added to it; those should all just be passed through via `body` to callers of `provider.generateText`)

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -306,7 +311,9 @@ export class DatabricksResponsesAgentLanguageModel implements LanguageModelV3 {
         )
         .pipeThrough(getDatabricksLanguageModelTransformStream()),
       request: { body: networkArgs.body },
-      response: { headers: responseHeaders },
+      response: {
+        headers: responseHeaders,
+      } as { headers?: SharedV3Headers; body?: unknown },
```
fanzeyi:
```suggestion
      },
```

integrations/ai-sdk-provider/src/responses-agent-language-model/responses-agent-language-model.ts
```diff
@@ -111,6 +113,9 @@ export class DatabricksResponsesAgentLanguageModel implements LanguageModelV3 {
         outputTokens: { total: response.usage?.output_tokens ?? 0, text: 0, reasoning: 0 },
       },
       warnings,
+      response: {
+        body: response,
+      } as { headers?: SharedV3Headers; body?: unknown },
```
fanzeyi:
```suggestion
      },
```

integrations/ai-sdk-provider/src/responses-agent-language-model/call-options-to-responses-args.ts
smurching: +1 to both, I think just a top-level `includeTrace` is probably more in the vein of how AI SDK names params, looking at e.g. `includeRawChunks` in `streamText`: https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#include-raw-chunks

PR #337, opened 2026-02-20
Support External MCP in get_databricks_resources

aravind-segu:
Automatically extract UCConnection resource for external mcp servers

Added unit tests

image

### DIFF
--- a/databricks_mcp/src/databricks_mcp/mcp.py
+++ b/databricks_mcp/src/databricks_mcp/mcp.py
@@ -16,6 +16,7 @@ from mlflow.models.resources import (
     DatabricksFunction,
     DatabricksGenieSpace,
     DatabricksResource,
+    DatabricksUCConnection,
     DatabricksVectorSearchIndex,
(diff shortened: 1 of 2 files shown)


databricks_mcp/src/databricks_mcp/mcp.py
bbqiu, line 253: nit: can we update this helper function

aravind-segu: update it to? The helper function just looks at the patterns in the MCP_URL_PATTERNS dict and matches it, so since we added a new pattern above, it should match the external mcp automatically

aravind-segu:
> changes lgtm with one nit
> 
> can we add an integration test where we call an external MCP connection and verify everything works e2e

I dont think this requires an integration test. This is not something to do with a calling an external mcp connection. it just looks at the url and returns a DatabricksUCConnection with the connection name. I think a unit test should be sufficient here.

PR #346, opened 2026-02-24
Add LangChain Lakebase integration tests

dhruv0811:
## Summary

Added LangChain Lakebase integration tests covering `DatabricksStore` (sync + async) and `CheckpointSaver` (sync + async) wrappers. Tests validate the bridge code's behavior — pool creation, connection borrowing, `_with_store` handoff, and `async with` lifecycle management — against a live Databricks Lakebase instance.

## Test plan
- [x] All 12 tests collected and skip gracefully without `LAKEBASE_INSTANCE_NAME`
- [x] Existing unit tests (`test_store.py`, `test_checkpoint.py`) still pass (16/16)
- [x] Core Lakebase tests pass (54/54)
- [x] LangChain Lakebase integration tests pass against live Lakebase instance via ai-oss runner: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/22643496765

### DIFF
--- a/integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
+++ b/integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
@@ -23,13 +23,14 @@ import pytest
 pytest.importorskip("langgraph", reason="langgraph not installed (requires memory extra)")
 
 from databricks_ai_bridge.lakebase import LakebaseClient
+from langgraph.checkpoint.base import Checkpoint, CheckpointMetadata
+
 from databricks_langchain import (
     AsyncCheckpointSaver,
     AsyncDatabricksStore,
     CheckpointSaver,
     DatabricksStore,
 )
-from langgraph.checkpoint.base import Checkpoint, CheckpointMetadata
 
 # Skip all tests if LAKEBASE_INSTANCE_NAME is not set
 pytestmark = pytest.mark.skipif(


tests/integration_tests/lakebase/test_lakebase_integration.py
```diff
@@ -99,7 +99,7 @@ def create_workspace_client_with_token(token: str | None = None) -> WorkspaceCli
                 "DATABRICKS_HOST must be set when using a custom token. "
                 "Example: export DATABRICKS_HOST=https://your-workspace.databricks.com"
             )
-        return WorkspaceClient(host=host, token=token)
+        return WorkspaceClient(host=host, token=token, auth_type="pat")
```
jennsun: why are we using pat tokens here instead of oauth?

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,384 @@
...
+
+def _drop_tables(tables: list[str]) -> None:
+    """Drop the given tables from the Lakebase instance."""
+    from databricks_ai_bridge.lakebase import LakebasePool
+
+    pool = LakebasePool(instance_name=get_instance_name())
+    with pool.connection() as conn:
+        for table in tables:
+            conn.execute(f"DROP TABLE IF EXISTS {table} CASCADE")
+    pool.close()
```
jennsun: instead of manually grabbing a connection from the pool to execute SQL queries, can we use the LakebaseClient class? https://github.com/databricks/databricks-ai-bridge/blob/main/src/databricks_ai_bridge/lakebase.py#L395

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,384 @@
...
+    yield
+    _drop_tables(CHECKPOINT_TABLES)
+
+
+# =============================================================================
+# DatabricksStore (Sync) Tests
+# =============================================================================
+
+
+class TestDatabricksStore:
```
jennsun, line 126: Can we add integration tests for the DatabricksStore which supports embedding_endpoint / embeddings for vector-based search via PostgresIndexConfig? 
```
  store = DatabricksStore(
      instance_name=get_instance_name(),
      embedding_endpoint="databricks-bge-large-en",
      embedding_dims=1024,
  )
```
followed by a search() call would be valuable, since this is a key differentiator of the Lakebase store

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,384 @@
...
+class TestAsyncDatabricksStore:
+    """Test asynchronous AsyncDatabricksStore against a live Lakebase instance."""
+
+    @pytest.mark.asyncio
+    async def test_async_store_put_and_get(self, unique_namespace, cleanup_store_tables):
+        """Test async pool open/close/borrow cycle with put and get."""
+        from databricks_langchain import AsyncDatabricksStore
+
+        store = AsyncDatabricksStore(instance_name=get_instance_name())
+        await store._lakebase.open()
```
jennsun: we do not need to be manually calling store._lakebase.open() / store._lakebase.close() - this reaches into a private attribute. Following the other three async store tests (test_async_store_search, test_async_store_delete, test_async_store_context_manager) can we rewrite it to match the pattern of the other tests:
```
  @pytest.mark.asyncio
  async def test_async_store_put_and_get(self, unique_namespace, cleanup_store_tables):
      """Test async put and get operations through bridge."""
      from databricks_langchain import AsyncDatabricksStore

      async with AsyncDatabricksStore(instance_name=get_instance_name()) as store:
          await store.setup()

          ns = unique_namespace
          await store.aput(ns, "async_key", {"data": "async hello"})

          item = await store.aget(ns, "async_key")
          assert item is not None
          assert item.value == {"data": "async hello"}
          assert item.key == "async_key"
```

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,384 @@
...
+
+        thread_id = uuid.uuid4().hex
+
+        with CheckpointSaver(instance_name=get_instance_name()) as saver:
+            saver.setup()
+
+            config = {"configurable": {"thread_id": thread_id, "checkpoint_ns": ""}}
+            checkpoint = Checkpoint(
+                v=1,
+                id=uuid.uuid4().hex,
```
jennsun: can we clarify the difference between the 2 `uuid.uuid4().hex` where this is the checkpoint id and other refers to thread id (We can even add a comment here to specify) 

```suggestion
                id=uuid.uuid4().hex, # checkpoint id
```

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,384 @@
...
+                    ts=f"2025-01-01T00:0{i}:00+00:00",
+                    channel_values={},
+                    channel_versions={},
+                    versions_seen={},
+                    pending_sends=[],
+                )
+                saver.put(config, checkpoint, CheckpointMetadata(), {})
+
+            checkpoints = list(saver.list(config))
+            assert len(checkpoints) == 3
```
jennsun, line 303: since list() should return checkpoints in reverse chronological order, can we verify ordering in this test? something like:
```
timestamps = [c.checkpoint["ts"] for c in checkpoints]
assert timestamps[0] > timestamps[1] > timestamps[2]
```

tests/integration_tests/lakebase/test_lakebase_integration.py
jennsun: if we explicitly want to use pat can we update the surrounding documentation that specifies to use oauth tokens?

dhruv0811: I was running into a `ValueError: validate: more than one authorization method configured: oauth and pat.`, will change this to use OAuth over PAT, thanks!

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
dhruv0811: It seems `list()` doesn't order by the checkpoints timestamp `ts` field, PostgresSaver orders by internal checkpoint ID/sequence which is random so the ordering of this list in unpredictable I believe.

dhruv0811: Reference: 

https://github.com/langchain-ai/langgraph/blob/48167d7fec9c25228426c92ba83d8650b77de0f3/libs/checkpoint-postgres/langgraph/checkpoint/postgres/__init__.py#L115

jennsun: ok thanks for confirming this!

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,398 @@
...
+            await store.setup()
+
+            ns = unique_namespace
+            await store.aput(ns, "ctx_key", {"data": "context managed"})
+
+            item = await store.aget(ns, "ctx_key")
+            assert item is not None
+            assert item.value == {"data": "context managed"}
+
+
```
jennsun, line 260: can we add test for vector search in the async store? similar to test_store_vector_search

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,398 @@
...
+# =============================================================================
+# AsyncDatabricksStore Tests
+# =============================================================================
+
+
+class TestAsyncDatabricksStore:
+    """Test asynchronous AsyncDatabricksStore against a live Lakebase instance."""
+
+    @pytest.mark.asyncio
+    async def test_async_store_put_and_get(self, unique_namespace, cleanup_store_tables):
```
jennsun: nit but can we name this test_async_store_setup_put_and_get to be parallel with the sync test test_store_setup_put_and_get

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,398 @@
...
+
+        ns = unique_namespace
+        store.put(ns, "doc_python", {"text": "Python is a programming language"})
+        store.put(ns, "doc_coffee", {"text": "Coffee is a popular beverage"})
+
+        results = store.search(ns, query="programming languages")
+        assert len(results) > 0
+        # The programming-related doc should rank higher than the coffee doc
+        assert results[0].key == "doc_python"
+
```
jennsun, line 186: can we add a context manager test for the sync class to be aligned with async tests like test_async_store_context_manager

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,398 @@
...
+                    channel_versions={},
+                    versions_seen={},
+                    pending_sends=[],
+                )
+                saver.put(config, checkpoint, CheckpointMetadata(), {})
+
+            checkpoints = list(saver.list(config))
+            assert len(checkpoints) == 3
+
+
```
jennsun, line 305: can we add a context manager for sync ver of class similar to test_async_checkpoint_context_manager?

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
dhruv0811: `DatabricksStore` doesnt support a context manager currently!

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,447 @@
...
+                    channel_versions={},
+                    versions_seen={},
+                    pending_sends=[],
+                )
+                saver.put(config, checkpoint, CheckpointMetadata(), {})
+
+            checkpoints = list(saver.list(config))
+            assert len(checkpoints) == 3
+
+    def test_checkpoint_context_manager(self, cleanup_checkpoint_tables):
```
annzhang-db: It looks like this test is basically identical to test_checkpoint_write_and_read, curious what the intention of having two distinct tests here is (and the same for async)

dhruv0811: With the context manager test I wanted to see that the pool is closed after exiting `with`. Youre right that this is identical to the other test, will merge with that test instead and add an assert for the additional check.

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,384 @@
...
+            checkpoints = list(saver.list(config))
+            assert len(checkpoints) == 3
+
+
+# =============================================================================
+# AsyncCheckpointSaver Tests
+# =============================================================================
+
+
+class TestAsyncCheckpointSaver:
```
bbqiu, line 311: for these tests, can we extract checkpoint construction into a helper across sync/async

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,384 @@
...
+# DatabricksStore (Sync) Tests
+# =============================================================================
+
+
+class TestDatabricksStore:
+    """Test synchronous DatabricksStore against a live Lakebase instance."""
+
+    def test_store_setup_put_and_get(self, unique_namespace, cleanup_store_tables):
+        """Test core bridge path: pool creation -> _with_store -> setup + batch (put/get)."""
+        from databricks_langchain import DatabricksStore
```
bbqiu: nit: move all imports to top level, and let's make this is a test a skip if the memory extra isn't installed (ex. can't import it)

bbqiu: not 100% sure how CI is set up (are we always testing with the memory extra installed?)

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
```diff
@@ -0,0 +1,384 @@
...
+# =============================================================================
+
+
+@pytest.fixture
+def unique_namespace() -> tuple[str, str]:
+    """Generate a UUID-based namespace tuple for test isolation."""
+    return ("test", f"ns_{uuid.uuid4().hex[:8]}")
+
+
+@pytest.fixture(scope="module")
```
bbqiu, line 81: i don't think it's too much of a problem since the async / sync classes all share the same class, but maybe a comment noting that dropping tables is only happening 1 time per run, not per test etc.

integrations/langchain/tests/integration_tests/test_langchain_lakebase.py
dhruv0811: CI is setup to always install the memory extra. So yes, we can move the imports to top level and add a skip if it isnt installed!

bbqiu (approved):
LGTM!

PR #348, opened 2026-02-24
Add FMAPI tool calling contract tests for DatabricksOpenAI

dhruv0811:
## Summary

End-to-end FMAPI tool calling integration tests for `DatabricksOpenAI` (sync + Agents SDK) and `ChatDatabricks` (LangGraph). Dynamically discovers all FMAPI models via the capabilities API and runs single-turn, multi-turn, and streaming tool calling tests against each.

### New files

- `src/databricks_ai_bridge/test_utils/fmapi.py` — shared test config (model discovery, skip lists, retry helpers)
- `integrations/openai/tests/integration_tests/test_fmapi_tool_calling.py` — OpenAI tests
- `integrations/langchain/tests/integration_tests/test_fmapi_tool_calling.py` — LangChain tests

### Test matrix

**OpenAI** (`TestAgentToolCalling` — async, Agents SDK + McpServer):
- Single-turn, multi-turn, streaming via `Runner.run` / `Runner.run_streamed`

**OpenAI** (`TestSyncClientToolCalling` — sync, direct `chat.completions.create`):
- Single-turn, multi-turn, streaming with manual tool call handling

**LangChain** (`TestLangGraphSync` + `TestLangGraphAsync`):
- Single-turn, multi-turn, streaming via `create_react_agent` + `invoke`/`ainvoke`/`stream`/`astream`

### Model discovery

Uses `capabilities.function_calling` from the serving-endpoints API (not probe-based). Models are discovered at test collection time, parametrized via `@pytest.mark.parametrize`. Gated behind `RUN_FMAPI_TOOL_CALLING_TESTS=1`.

### Known model issues (skipped)

| Model | Issue |
|---|---|
| `gpt-5-nano` | Too small for reliable tool calling |
| `gpt-oss-20b`, `gpt-oss-120b`, `llama-4-maverick` | Hallucinates tool names |
| `gemini-3-flash`, `gemini-3-pro`, `gemini-3-1-pro` | Requires `thought_signature` (Gemini 3.x) |
| `gemini-2-5-pro`, `gemini-2-5-flash` | Returns list content that breaks Agents SDK parsing |
| `gpt-5-*-codex` | Responses API only, no Chat Completions support |
| `gemma-3-12b` (LangChain only) | Outputs raw tool call text instead of executing tools |

## Test plan
- [x] Full CI run via ai-oss runner - https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/22656317474

### DIFF
--- a/integrations/openai/src/databricks_openai/utils/clients.py
+++ b/integrations/openai/src/databricks_openai/utils/clients.py
@@ -1,3 +1,4 @@
+import os
 from typing import Any, Generator
 
 from databricks.sdk import WorkspaceClient
@@ -12,6 +13,15 @@ from typing_extensions import override
 _APPS_ENDPOINT_PREFIX = "apps/"
 # Domain pattern indicating a Databricks App URL
 _DATABRICKS_APPS_DOMAIN = "databricksapps"
+
+
+def _get_openai_api_key():
+    """Return OPENAI_API_KEY from env if set, otherwise 'no-token'.
+
+    Passed through to the OpenAI client so that the agents SDK tracing
+    client can authenticate with the OpenAI API.
+    """
+    return os.environ.get("OPENAI_API_KEY") or "no-token"
 
 
 class BearerAuth(Auth):
@@ -214,7 +224,7 @@ class DatabricksResponses(Responses):
             # Authentication is handled via http_client, not api_key
             self._app_clients_cache[app_name] = OpenAI(
                 base_url=app_url,
-                api_key="no-token",
+                api_key=_get_openai_api_key(),
                 http_client=_get_authorized_http_client(self._workspace_client),
             )
         return self._app_clients_cache[app_name]
@@ -303,7 +313,7 @@ class DatabricksOpenAI(OpenAI):
         # Authentication is handled via http_client, not api_key
         super().__init__(
             base_url=target_base_url,
-            api_key="no-token",
+            api_key=_get_openai_api_key(),
             http_client=_get_authorized_http_client(workspace_client),
         )
 
@@ -361,7 +371,7 @@ class AsyncDatabricksResponses(AsyncResponses):
             # Authentication is handled via http_client, not api_key
             self._app_clients_cache[app_name] = AsyncOpenAI(
                 base_url=app_url,
-                api_key="no-token",
+                api_key=_get_openai_api_key(),
                 http_client=_get_authorized_async_http_client(self._workspace_client),
             )
         return self._app_clients_cache[app_name]
@@ -450,7 +460,7 @@ class AsyncDatabricksOpenAI(AsyncOpenAI):
         # Authentication is handled via http_client, not api_key
         super().__init__(
             base_url=target_base_url,
-            api_key="no-token",
+            api_key=_get_openai_api_key(),
             http_client=_get_authorized_async_http_client(workspace_client),
         )
 
--- a/integrations/openai/tests/unit_tests/test_clients.py
+++ b/integrations/openai/tests/unit_tests/test_clients.py
@@ -1,3 +1,4 @@
+import os
 from typing import Any, cast
 from unittest.mock import AsyncMock, MagicMock, patch
 
@@ -15,6 +16,7 @@ from databricks_openai.utils.clients import (
     _get_app_url,
     _get_authorized_async_http_client,
     _get_authorized_http_client,
+    _get_openai_api_key,
     _should_strip_strict,
     _strip_strict_from_tools,
     _validate_oauth_for_apps,
@@ -66,7 +68,11 @@ class TestDatabricksOpenAI:
 
     def test_init_with_default_workspace_client(self):
         """Test initialization with default WorkspaceClient."""
-        with patch("databricks_openai.utils.clients.WorkspaceClient") as mock_ws_client_class:
+        env = {k: v for k, v in os.environ.items() if k != "OPENAI_API_KEY"}
+        with (
+            patch.dict("os.environ", env, clear=True),
+            patch("databricks_openai.utils.clients.WorkspaceClient") as mock_ws_client_class,
+        ):
             mock_client = MagicMock(spec=WorkspaceClient)
             mock_client.config.host = "https://default.databricks.com"
             mock_client.config.authenticate.return_value = {"Authorization": "Bearer default-token"}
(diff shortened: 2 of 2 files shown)


integrations/openai/tests/integration_tests/test_fmapi_tool_calling.py
```diff
@@ -0,0 +1,216 @@
...
+from databricks.sdk import WorkspaceClient
+
+from databricks_openai import AsyncDatabricksOpenAI
+
+pytestmark = pytest.mark.skipif(
+    os.environ.get("RUN_FMAPI_TOOL_CALLING_TESTS") != "1",
+    reason="FMAPI tool calling tests disabled. Set RUN_FMAPI_TOOL_CALLING_TESTS=1 to enable.",
+)
+
+_FOUNDATION_MODELS = [
```
bbqiu: can we have a dynamic list that gets the latest models?

idk if we can use it, but can we do something like the following api? if not, we can list endpoints and filter by `databricks-` prefix. to keep costs low, we should try to have very simple tool calling etc.

https://sourcegraph.prod.databricks-corp.com/databricks-eng/universe/-/blob/model-serving/model-serving-api/api/proto/endpoints.proto?L441

integrations/openai/tests/integration_tests/test_fmapi_tool_calling.py
```diff
@@ -0,0 +1,216 @@
...
+_MCP_FUNCTION = "echo_message"
+
+
+@pytest.fixture(scope="module")
+def workspace_client():
+    return WorkspaceClient()
+
+
+@pytest.fixture(scope="module")
+def async_client(workspace_client):
```
bbqiu, line 50: can we test both sync/async

integrations/langchain/tests/integration_tests/test_fmapi_tool_calling.py
```diff
@@ -0,0 +1,413 @@
...
+from databricks_openai import DatabricksOpenAI
+from langchain_core.messages import AIMessage, AIMessageChunk, ToolMessage
+from langchain_core.tools import tool
+from langgraph.checkpoint.memory import MemorySaver
+from langgraph.prebuilt import create_react_agent
+from openai.types.chat import ChatCompletionToolParam
+
+from databricks_langchain import ChatDatabricks
+
+pytestmark = pytest.mark.skipif(
```
bbqiu, line 27: overall nit: does it make sense for us to centralize 1 config file between the two integration test files? that way we only have to maintain a singular file of "config" for the integration tests?

bbqiu: we can also store shared helpers etc. in some place that makes sense

integrations/langchain/tests/integration_tests/test_fmapi_tool_calling.py
```diff
@@ -0,0 +1,413 @@
...
+        "parameters": {
+            "type": "object",
+            "properties": {"x": {"type": "string"}},
+            "required": ["x"],
+        },
+    },
+}
+
+
+def _supports_tool_calling(client: DatabricksOpenAI, model: str) -> bool:
```
bbqiu: were we not able to retrieve this info via the API? there should be some "capabilities" field that we rely on

you can check out the playground and how it determines if some models are tool calling for inspo

integrations/langchain/tests/integration_tests/test_fmapi_tool_calling.py
```diff
@@ -0,0 +1,413 @@
...
+    """Sync LangGraph agent tests mirroring app-templates/agent-langgraph.
+
+    Each test follows the pattern:
+      ChatDatabricks -> create_react_agent -> agent.invoke / agent.stream
+    """
+
+    def test_single_turn(self, model):
+        """Single-turn: agent calls tools and produces a final answer.
+
+        Mirrors the basic app-template @invoke() handler.
```
bbqiu: nit: can remove stuff like this about the app templates

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -61,6 +62,137 @@ def _should_strip_strict(model: str | None) -> bool:
     return "gpt" not in model.lower()
 
 
+def _is_gemini_model(model: str | None) -> bool:
```
bbqiu: these changes seem a little heavier than i'd expect -- can we see if AI can help us simplify the changes?

might also be good to provide context that this helps for gemini 2.5pro but not for gemini 3 models which require the thought signature as a part of the subsequent request

integrations/langchain/tests/integration_tests/test_fmapi_tool_calling.py
dhruv0811: The SDK doesn't have support for this, but the API does, so moving to use that. Thanks for reminding me on this!

integrations/langchain/tests/integration_tests/test_fmapi_tool_calling.py
dhruv0811: Moved to `test_utils` foolder in the core package.

integrations/openai/src/databricks_openai/utils/clients.py
dhruv0811: Cleaned it up a bit, not sure if its possible to clean up further.

src/databricks_ai_bridge/test_utils/fmapi.py
```diff
@@ -0,0 +1,130 @@
...
+    "databricks-claude-sonnet-4-6",
+    "databricks-claude-opus-4-6",
+    "databricks-meta-llama-3-3-70b-instruct",
+    "databricks-gpt-5-2",
+    "databricks-gpt-5-1",
+    "databricks-qwen3-next-80b-a3b-instruct",
+]
+
+
+def has_function_calling(w: WorkspaceClient, endpoint_name: str) -> bool:
```
bbqiu, line 65: this is in the list endpoints API response btw -- you're already calling it, so no need to make more API to get this info

Image

bbqiu: ah if it's not in the SDK, this is fine to keep this way

integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -61,6 +62,77 @@ def _should_strip_strict(model: str | None) -> bool:
     return "gpt" not in model.lower()
 
 
+# Gemini FMAPI compat: rejects list content in tool messages (request side)
```
bbqiu: discussed offline -- since we're now modifying output, it may feel a bit unintuitive to customers as to why the stream looks diff here vs via rest request or playground

let's just add 2.5 pro/other models that return a thought signature to the skip list for now

src/databricks_ai_bridge/test_utils/fmapi.py
dhruv0811: Yeah its not there :( I double-checked

PR #356, opened 2026-03-04
Fix Responses API tool calling for codex models

dhruv0811:
## Summary

Fixes Responses API tool calling for codex and GPT models in `DatabricksOpenAI` and `ChatDatabricks`.

**`clients.py`** — Truncate FMAPI response/input IDs to 64 chars (FMAPI returns 191-char IDs but rejects them on replay). Covers both non-streaming (`_truncate_response_ids`) and streaming (`_truncate_input_ids`) paths.

**`chat_models.py`** — Convert tools from Chat Completions to Responses API format in `_prepare_inputs`. Fix `msg_`/`fc_` ID prefixes required by FMAPI in `_convert_lc_messages_to_responses_api`. Truncate all IDs to 64 chars (the LangChain sync path uses the standard SDK client, not `DatabricksOpenAI`, so it needs its own truncation). Strip output-only fields (`status`, `namespace`) from passthrough blocks — `model_dump()` on Pydantic response items includes these, but FMAPI rejects them as unknown input parameters.

**`fmapi.py`** — Add `discover_responses_models()` and `SKIP_RESPONSES_API` for Responses API test discovery.

**Test files** — Add `TestAgentToolCallingResponsesAPI` (OpenAI) and `TestLangGraphResponsesAPI` (LangChain) with single-turn, multi-turn, and streaming tests.

## Test plan
- [x] Full FMAPI tool calling integ tests CI run: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/22980584642
- [x] Unit tests pass (290 LangChain + 150 OpenAI)
- [x] Local live multi-turn agent test against FMAPI


integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -1463,13 +1490,23 @@ def _convert_lc_messages_to_responses_api(messages: list[BaseMessage]) -> list[d
...
                             "mcp_approval_request",
                         ):
-                            input_items.append(block | {"id": lc_msg.id})
+                            # FMAPI rejects output-only fields on input items.
+                            block.pop("status", None)
+                            # Fix ids: FMAPI requires fc_ prefix and max 64 chars.
+                            if "id" not in block:
+                                call_id = block.get("call_id", "")
+                                block["id"] = _truncate(
+                                    f"fc_{call_id}" if call_id else (lc_msg.id or "")
```
bbqiu: nit: let's check if it's already prefixed

also does it make sense to ever have a blank lc_msg.id? should we generate on in this case

src/databricks_ai_bridge/test_utils/fmapi.py
```diff
@@ -32,17 +33,24 @@
     "databricks-gpt-5-1-codex-mini",  # Responses API only, no Chat Completions support
     "databricks-gpt-5-2-codex",  # Responses API only, no Chat Completions support
     "databricks-gpt-5-3-codex",  # Responses API only, no Chat Completions support
+    # Added in this PR:
```
bbqiu: nit: remove this

integrations/langchain/src/databricks_langchain/chat_models.py
dhruv0811: Makes sense to keep the code defensive in the case its blank, modifying to generate an id.

PR #412, opened 2026-04-06
Support updated Workspace URL Path for AI Gateway via DatabricksOpenAI

jennsun:
Support Heilmeier: AI Gateway Token Scope - Unique Paths and Workspace URL as they plan to deprecate old URLs at end of may:
https://docs.google.com/document/d/17_Hdi27Ecjl8Ys9ZCvoOeC3L3AEZEOtNLh3FuiRIGGE/edit?tab=t.0

STAGING Testing notebook with updated workspace-based ai-gateway url path: https://dbc-fb904e0e-bca7.staging.cloud.databricks.com/editor/notebooks/2573471477383128?o=4146361371977516#command/4780814485864147

PROD Testing notebook in workspace without the new URL supported will use the ai gateway host url: https://eng-ml-inference-team-eu-central-1.cloud.databricks.com/editor/notebooks/421682237852321?o=5763476629593792#command/6870101689879415

- Adds support for the new workspace-based AI Gateway URL pattern `({host}/ai-gateway/{api_path})` where AI Gateway endpoints are accessed via a path prefix on the workspace host, instead of requiring a separate
  *.ai-gateway.* hostname
  - tries the new workspace /ai-gateway/ prefix first, falls back to the legacy *.ai-gateway.* hostname discovery via
  /api/ai-gateway/v2/endpoints
  - use_ai_gateway=True routes to mlflow/v1 (chat completions API)
  - use_ai_gateway_native_api=True routes to openai/v1 (responses API)

route not enabled yet:
image

  How detection works

  The SDK sends a lightweight GET {host}/ai-gateway/mlflow/v1/chat/completions
  to check if the workspace has the /ai-gateway/ prefix wired up at the proxy
  layer:
  - JSON response (any status code) → request reached the AI Gateway backend →
  workspace prefix is available
  - Bare 404 with no content-type → proxy doesn't recognize the route → fall
  back to legacy hostname

### DIFF
--- a/integrations/openai/src/databricks_openai/utils/clients.py
+++ b/integrations/openai/src/databricks_openai/utils/clients.py
@@ -3,7 +3,7 @@ from typing import Any, Generator
 from urllib.parse import urlparse
 
 from databricks.sdk import WorkspaceClient
-from httpx import URL, AsyncClient, Auth, Client, Request, Response
+from httpx import AsyncClient, Auth, Client, Request, Response
 from openai import APIConnectionError, APIStatusError, AsyncOpenAI, OpenAI
 from openai.resources.chat import AsyncChat, Chat
 from openai.resources.chat.completions import AsyncCompletions, Completions
@@ -363,8 +363,7 @@ def _truncate_input_ids(input_items: Any) -> None:
 
 
 class DatabricksResponses(Responses):
-    """Responses resource that handles apps/ prefix routing, id truncation,
-    and base URL swapping for AI Gateway mlflow/v1 → openai/v1."""
+    """Responses resource that handles apps/ prefix routing and id truncation."""
 
     def __init__(self, client, workspace_client: WorkspaceClient):
         super().__init__(client)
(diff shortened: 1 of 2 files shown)


integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -136,17 +136,52 @@ def _discover_ai_gateway_host(
...
+        which means the route is live → return True.
+      - Proxy blocks the request: returns a bare 404 with no content-type,
+        meaning /ai-gateway/ is not configured on this workspace → return False.
+    """
+    try:
+        response = http_client.get(f"{host}/ai-gateway/mlflow/v1/chat/completions")
+        content_type = response.headers.get("content-type", "")
+        if "application/json" in content_type:
+            return True
+        return response.status_code != 404
```
jennsun: unfortunately this detection method for when the workspace url is available is not the most elegant but the heilmeir team doesn't have a health checkpoint we can use, and currently listing ai gateway endpoints uses the old ai-gateway hostname https://docs.google.com/document/d/17_Hdi27Ecjl8Ys9ZCvoOeC3L3AEZEOtNLh3FuiRIGGE/edit?tab=t.0

aravind-segu (approved):

PR #415, opened 2026-04-16
Agent Memory SDKs: allow for custom schemas

bbqiu:
## Problem

When using the `public` schema, SPs need to be granted postgres permissions explicitly. this means that when creating a template through the UI (without any manual permission grants in the UI / through a script), the SP will not have permission to write/create on it.

## Solution
Adds a `schema` param to CheckpointSaver, AsyncCheckpointSaver, DatabricksStore, AsyncDatabricksStore, and AsyncDatabricksSession. Since it'll be a new creation from the UI (given a fresh db), this will hopefully bypass permission issues.

The implementation uses two layers:
- Pool/engine level: SET search_path on each connection
- Class level: CREATE SCHEMA IF NOT EXISTS in setup()/ensure_tables()

## Rollout 
For current users of this SDK, nothing changes since this param defaults to None.

For new users of the SDK (like the updated templates) that provide a schema, this will create tables in the specified schema instead of `public`

## Testing
unit tests
manually tested with templates

### DIFF
--- a/integrations/langchain/src/databricks_langchain/checkpoint.py
+++ b/integrations/langchain/src/databricks_langchain/checkpoint.py
@@ -57,13 +57,7 @@ class CheckpointSaver(PostgresSaver):
 
     def setup(self) -> None:
         """Set up the checkpoint database, creating the schema if specified."""
-        if self._schema:
-            from psycopg import sql
-
-            with self._lakebase.connection() as conn:
-                conn.execute(
-                    sql.SQL("CREATE SCHEMA IF NOT EXISTS {}").format(sql.Identifier(self._schema))
-                )
+        self._lakebase.create_schema()
         super().setup()
 
     def __enter__(self):
@@ -118,13 +112,7 @@ class AsyncCheckpointSaver(AsyncPostgresSaver):
 
     async def setup(self) -> None:
         """Set up the checkpoint database asynchronously, creating the schema if specified."""
-        if self._schema:
-            from psycopg import sql
-
-            async with self._lakebase.connection() as conn:
-                await conn.execute(
-                    sql.SQL("CREATE SCHEMA IF NOT EXISTS {}").format(sql.Identifier(self._schema))
-                )
+        await self._lakebase.create_schema()
         await super().setup()
 
     async def __aenter__(self):
--- a/integrations/langchain/src/databricks_langchain/checkpoint.py
+++ b/integrations/langchain/src/databricks_langchain/checkpoint.py
@@ -33,6 +33,7 @@ class CheckpointSaver(PostgresSaver):
         project: str | None = None,
         branch: str | None = None,
         workspace_client: WorkspaceClient | None = None,
+        schema: str | None = None,
         **pool_kwargs: Any,
     ) -> None:
         # Lazy imports
@@ -48,9 +49,24 @@ class CheckpointSaver(PostgresSaver):
             project=project,
             branch=branch,
             workspace_client=workspace_client,
+            schema=schema,
             **dict(pool_kwargs),
         )
+        self._schema = schema
         super().__init__(self._lakebase.pool)
+
+    def setup(self) -> None:
+        """Set up the checkpoint database, creating the schema if specified."""
+        if self._schema:
+            from psycopg import sql
+
+            with self._lakebase.connection() as conn:
+                conn.execute(
+                    sql.SQL("CREATE SCHEMA IF NOT EXISTS {}").format(
+                        sql.Identifier(self._schema)
+                    )
+                )
+        super().setup()
 
     def __enter__(self):
         """Enter context manager."""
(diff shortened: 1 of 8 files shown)


integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -168,25 +173,40 @@ def __init__(
...
 
+    async def _ensure_tables(self) -> None:
+        """Ensure schema and tables are created before any database operations."""
+        if self._schema and self._create_tables:
+            import re
+
+            from sqlalchemy import text
+
+            if not re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", self._schema):
+                raise ValueError(f"Invalid schema name: {self._schema!r}")
```
dhruv0811, line 184: very nit: can we add a better desc of what the name should look like? if the error is thrown

src/databricks_ai_bridge/lakebase.py
```diff
@@ -441,6 +445,20 @@ def connect(cls, conninfo: str = "", **kwargs):
...
         timeout = pool_kwargs.pop("timeout", DEFAULT_TIMEOUT)
 
+        # Set search_path on each connection when a schema is specified
+        user_configure = pool_kwargs.pop("configure", None)
+        _configure_fn = None
+        if self.schema:
+            _schema = self.schema
+            _user_configure = user_configure
+
+            def _configure_fn(conn):  # type: ignore[misc]  # noqa: F811
```
dhruv0811, line 455: Same function duplicated like thrice across the file it seems, can we extract this to have only one copy.

src/databricks_ai_bridge/lakebase.py
```diff
@@ -441,6 +445,20 @@ def connect(cls, conninfo: str = "", **kwargs):
         max_size = pool_kwargs.pop("max_size", DEFAULT_MAX_SIZE)
         timeout = pool_kwargs.pop("timeout", DEFAULT_TIMEOUT)
 
+        # Set search_path on each connection when a schema is specified
+        user_configure = pool_kwargs.pop("configure", None)
```
dhruv0811, line 449: This pop is run unconditionally. User provided configure is silently dropped when no schema is set. I think this violates our "nothing changes for current users" rollout claim. Maybe also add a test covering `configure=fn, schema=None`.

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -168,25 +173,40 @@ def __init__(
...
         )
 
+    async def _ensure_tables(self) -> None:
+        """Ensure schema and tables are created before any database operations."""
+        if self._schema and self._create_tables:
+            import re
+
+            from sqlalchemy import text
+
+            if not re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", self._schema):
```
dhruv0811, line 183: Why aren't we using `sql.Identifier(...)` like elsewhere in this PR?

integrations/openai/src/databricks_openai/agents/session.py
```diff
@@ -168,25 +173,40 @@ def __init__(
             session_id,
         )
 
+    async def _ensure_tables(self) -> None:
+        """Ensure schema and tables are created before any database operations."""
+        if self._schema and self._create_tables:
+            import re
+
+            from sqlalchemy import text
```
dhruv0811, line 181: Can we move imports to be at the top?

dhruv0811 (commented):
Looks good overall! Left some comments 

Also, even with these changes, when I tried it out, I hit a SP-permissions wall in the Drizzle (TypeScript frontend) migration. So just these changes wont be sufficient to get rid of the grant script for our advanced templates right?

integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -95,10 +110,23 @@ def __init__(
...
+        """Set up the checkpoint database asynchronously, creating the schema if specified."""
+        if self._schema:
+            from psycopg import sql
+
+            async with self._lakebase.connection() as conn:
+                await conn.execute(
+                    sql.SQL("CREATE SCHEMA IF NOT EXISTS {}").format(sql.Identifier(self._schema))
+                )
+        await super().setup()
+
```
jennsun, line 117: Rather than grabbing connections and executing SQL queries manually like this, can we update the LakebaseClient in this PR to support `create_schema` (similar to `create_role`) and use it whenever we call this updated setup() function)? 

https://github.com/databricks/databricks-ai-bridge/blob/main/src/databricks_ai_bridge/lakebase.py#L788

bbqiu: great suggestion, added

src/databricks_ai_bridge/lakebase.py
bbqiu: abstracted a bit more of it away, but unfortunately need to keep separate copies bc of sync / async

integrations/openai/src/databricks_openai/agents/session.py
bbqiu: removed this regex and used sql identifier instead

dhruv0811 (approved):
lgtm!

PR #416, opened 2026-04-16
LongRunningAgentServer: durable resume via heartbeat + CAS claim

dhruv0811:
## Summary

Adds durable execution to `LongRunningAgentServer`: pods that die mid-stream are seamlessly recovered by another pod, with SDK-agnostic history repair so agent code stays untouched.

## How it works

**Heartbeat + CAS claim.** The owning pod writes `heartbeat_at` every `heartbeat_interval_seconds` (3s). When `GET /responses/{id}` lands on any pod and the row's heartbeat is older than `heartbeat_stale_threshold_seconds` (10s), the serving pod atomically claims the row via a conditional `UPDATE … RETURNING` and bumps `attempt_number`.

**Rotate + replay + sanitize.** On claim, the new pod re-invokes the registered handler with:
- a rotated `conversation_id` (suffixed `::attempt-N`) so the SDK opens a fresh thread/session for the replay
- the original request's `input` enriched with the prior attempt's completed tool pairs + narrative messages + a synthetic `[INTERRUPTED]` output paired with any tool call that didn't finish
- all items passed through `sanitize_tool_items` (shared walker in `long_running/repair.py`) so Anthropic's `tool_use` → `tool_result` pairing holds

**SDK read-time repair.** Each integration wraps its own history store to filter orphans on load, preventing a past crash from poisoning subsequent turns:
- `databricks-openai`: `AsyncDatabricksSession.get_items()` always runs `sanitize_tool_items` on reads.
- `databricks-langchain`: the checkpointer's `(a)get_tuple` is wrapped by `_repair_loaded_checkpoint_tuple`, which injects synthetic `ToolMessage`s for orphan `AIMessage.tool_calls` in the trailing turn.

**Stream resume.** `GET /responses/{id}?stream=true&starting_after=N` replays the message log past the cursor; a `response.resumed` sentinel marks attempt boundaries.

## Additions

- `long_running/repair.py` — shared orphan-pair sanitizer + canonical `DEFAULT_SYNTHETIC_INTERRUPTED_OUTPUT`
- `long_running/server.py` — heartbeat writer, CAS claim, `_try_claim_and_resume`, rotate conversation_id, prior-attempt event inheritance, `response.resumed` sentinel
- `/_debug/kill_task/{id}` endpoint gated by `LONG_RUNNING_ENABLE_DEBUG_KILL=1` for integration tests
- `[durable]` INFO lifecycle logs (heartbeat, claim, stream start/stop)
- Idempotent `ADD COLUMN IF NOT EXISTS` migrations with a WARN summary when skipped due to insufficient privilege

## Settings (new)

- `heartbeat_interval_seconds = 3.0`
- `heartbeat_stale_threshold_seconds = 10.0` (validator: must exceed interval)

## Test plan

- [x] 179 unit tests pass (120 `long_running` + 40 `databricks-openai` session + 18 `databricks-langchain` checkpoint)
- [x] Deployed crash-recovery matrix via `/_debug/kill_task/{id}` on `agent-openai-advanced` + `agent-langgraph-advanced` — see [app-templates PR #195](https://github.com/databricks/app-templates/pull/195)
- [x] Multi-tool UI validation on Claude through the deployed `agent-openai-advanced` app: kill-mid-tool, verify durable resume preserves completed tool pairs and injects synthetic `[INTERRUPTED]` for the interrupted call
- [x] Passing CI run: https://github.com/databricks-eng/ai-oss-integration-tests-runner/actions/runs/24864211860

## Known follow-ups (non-blocking)

- Concurrent-claim / heartbeat-cascade / sequence-collision integration tests
- `DELETE /responses/{id}` implementation (currently 501)
- Exponential backoff in client resume loop
- Metrics: `heartbeat_miss_count`, `claim_latency_seconds`, `attempts_per_response`

https://github.com/user-attachments/assets/a30b81d6-f207-4fcc-8d28-daf9edcd89e5

https://github.com/user-attachments/assets/07b14931-0c70-497c-a7f3-c37aac051808


integrations/langchain/src/databricks_langchain/checkpoint.py
```diff
@@ -16,6 +21,109 @@
...
+    from langchain_core.messages import AIMessage, ToolMessage
+
+    _message_imports_available = True
+except ImportError:
+    AIMessage = object  # type: ignore
+    ToolMessage = object  # type: ignore
+    _message_imports_available = False
+
+
+def _build_tool_resume_repair(messages: Sequence[Any]) -> list[Any]:
```
bbqiu, line 34: why do we need these synthetic events?

from our discussion in person, i think it should be sufficient for us to serialize the entire message history from the crashed conversations into a single user message with a prompt to the LLM about "the agent task runnig this crashed, contniue from here"

(maybe can say recovering from crash in the text also to explain why there is duplicatd content like two tool calls in a row)

pros/cons list from talking to claude:

Image

imo just prose recovery is much cleaner and doesn't rely on agent authors to know to use these specific APIs

bbqiu (commented):
re: overall approach, i still think we should go for a non-framework specific solution

for future reference, can you generate an AGENTS.md file that describes all of th edesired behaviors of this long running agent server?

some CUJs explaining what agent authoring requirements there are + the interface that we expect this to use today / how they'll change for taskflow would also be helpful

a mermaid diagram for human reviewers in the future woudl also help here!

PR #418, opened 2026-04-18
Update ChatDatabricks to use DatabricksOpenAI

aravind-segu:
Update ChatDatabricks to use DatabricksOpenAI package. Additionally adds the ability to use the use_ai_gateway=True parameter

## Tests

Unit tests and integration tests

Notebook: https://dogfood.staging.databricks.com/editor/notebooks/667884564555753?o=6051921418418893

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -299,7 +299,7 @@ class ChatDatabricks(BaseChatModel):
     max_retries: Optional[int] = None
     """Maximum number of retries for failed requests. If None, uses the default retry count."""
     use_ai_gateway: bool = False
-    """If True, route requests through AI Gateway using the MLflow API."""
+    """If True, route requests through AI Gateway using the Supervisor API."""
     use_ai_gateway_native_api: bool = False
     """If True, route requests through AI Gateway using the native OpenAI-compatible API."""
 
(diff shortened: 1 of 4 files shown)


integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -298,6 +298,10 @@ class GetPopulation(BaseModel):
     """Timeout in seconds for the HTTP request. If None, uses the default timeout."""
     max_retries: Optional[int] = None
     """Maximum number of retries for failed requests. If None, uses the default retry count."""
+    use_ai_gateway: bool = False
```
bbqiu, line 301: https://github.com/databricks/databricks-ai-bridge/blob/main/integrations/openai/src/databricks_openai/utils/clients.py#L340-L345

can we just reuse these docstrings? i think as it stands, things are a bit confusing (we are mixing supervisor api w/ ai gateway)

integrations/langchain/tests/unit_tests/test_utils.py
```diff
@@ -2,61 +2,27 @@
...
     mock_workspace_client = Mock()
-    mock_workspace_client.serving_endpoints.get_open_ai_client.return_value = mock_openai_client
-
-    client = get_openai_client(workspace_client=mock_workspace_client)
 
-    # Verify the OpenAI client was obtained without kwargs
-    mock_workspace_client.serving_endpoints.get_open_ai_client.assert_called_once_with()
+    with patch("databricks_langchain.utils.DatabricksOpenAI", return_value=mock_client) as mock_cls:
+        with patch("databricks.sdk.WorkspaceClient", return_value=mock_workspace_client):
+            wc = mock_workspace_client if with_workspace_client else None
```
bbqiu, line 24: can we also make sure the passthrough works for the async path?

integrations/langchain/src/databricks_langchain/chat_models.py
bbqiu: could we also make sure this raises on initalization if both are true so the user has early feedback abotu somethign being wrong

bbqiu (approved):
lgtm with two nits! thank you for doing this

PR #419, opened 2026-04-20
Use last query attachment in _parse_attachments

nisha2003:
## Summary

Fix `_parse_attachments` to handle Genie self-correction. When Genie self-corrects, the response contains multiple query attachments and paired text attachments — the old logic returned the broken first attempt.                             
                                                                                                                          
  ## Changes                                                                                                              
  - `query_attachment`: return the last one (corrected query)
  - `text_attachment`: return the first text following the last query (its paired explanation)
  - `suggested_questions_attachment`: unchanged                                                                           
  
  ## Test plan                                                                                                            
  - [x] Added test cases for self-correction (single-query and full query+text interleaving)                                          
  - [x] Verified in e2-dogfood against space `01f0c4c9431611b8843a80bfd9ebe916` — [notebook](https://dogfood.staging.databricks.com/editor/notebooks/1504339675866128?o=6051921418418893#command/5643578246396567)

### DIFF
--- a/src/databricks_ai_bridge/genie.py
+++ b/src/databricks_ai_bridge/genie.py
@@ -199,15 +199,21 @@ def _parse_attachments(resp: Dict[str, Any]) -> Dict[str, Any]:
     if not isinstance(attachments, list):
         return result
 
+    # Genie may self-correct, producing multiple query+text pairs. We want
+    # the final query and its paired text (the first text attachment following
+    # the final query).
+    want_new_text = True
     for a in attachments:
         if not isinstance(a, dict):
             continue
 
         if "query" in a:
             result["query_attachment"] = a
+            want_new_text = True
 
-        elif "text" in a and result["text_attachment"] is None:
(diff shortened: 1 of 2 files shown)


src/databricks_ai_bridge/genie.py
```diff
@@ -199,15 +199,21 @@ def _parse_attachments(resp: Dict[str, Any]) -> Dict[str, Any]:
...
+        if "query" in a:
             result["query_attachment"] = a
+            want_new_text = True
 
-        elif "text" in a and result["text_attachment"] is None:
+        elif "text" in a and want_new_text:
             result["text_attachment"] = a
+            want_new_text = False
 
         elif "suggested_questions" in a and result["suggested_questions_attachment"] is None:
```
bbqiu: do we want to also get the last suggested qeustions?

bbqiu (approved):
lgtm after addressing one last comment! thank you

PR #431, opened 2026-06-05
Add experimental Session Store SDK adapters

annzhang-db:
## Summary

This draft prototypes how third-party agent SDKs can use the proposed Databricks Session Store REST API.

- adds a shared async REST client with Databricks authentication, pagination, LiteSwap routing, and Session/SessionEvent operations
- adds `databricks_openai.agents.DatabricksSession`, implementing the OpenAI Agents SDK `Session` protocol
- adds `DatabricksClaudeSessionStore`, implementing Claude Agent SDK transcript mirroring, listing, deletion, and subagent transcript discovery
- supports a separate Conversations API base URL for the Databricks OpenAI client
- documents branch installation and runnable OpenAI/Claude integration shapes

## Why this is a draft

The REST resource names, endpoint prefix, and idempotency contract are still being finalized. The public classes are marked experimental so we can use this branch in Databricks Apps while continuing to make breaking prototype changes.

## API design findings

- OpenAI maps directly to list, append, pop, and clear operations.
- Claude storage identity is `(project_key, session_id, optional subpath)`; the adapter includes the project hash in the physical Databricks session ID to avoid cross-project collisions.
- Claude subagent `subpath` values map naturally to child Sessions.
- Claude retries UUID-bearing transcript entries. The prototype performs best-effort read-before-append deduplication, but atomic retry safety still requires a caller idempotency key per event or equivalent conditional append semantics.

## How do you know it works?

The new tests cover authenticated request construction, URL encoding, pagination, errors, all event mutations, OpenAI item mapping and protocol conformance, Claude UUID deduplication, project isolation, subagent child sessions, listing, and cascade deletion.

- Core suite: 104 passed, 54 skipped, 1 environment-bound credential test deselected.
- OpenAI unit suite: 179 passed, 1 skipped, 2 environment-bound credential tests deselected.
- Ruff formatting and lint pass for the core and OpenAI packages.
- Targeted type checks pass; full package checks report only pre-existing unused-ignore warnings.
- A live LiteSwap smoke test successfully created, appended, listed, popped, cleared, and deleted a session.
- Live adapter smoke tests round-tripped an OpenAI item and verified Claude UUID deduplication, project listing, and subagent discovery.

### DIFF
--- a/integrations/openai/src/databricks_openai/utils/clients.py
+++ b/integrations/openai/src/databricks_openai/utils/clients.py
@@ -6,6 +6,7 @@ from httpx import AsyncClient, Auth, Client, Request, Response
 from openai import APIConnectionError, APIStatusError, AsyncOpenAI, OpenAI
 from openai.resources.chat import AsyncChat, Chat
 from openai.resources.chat.completions import AsyncCompletions, Completions
+from openai.resources.conversations import AsyncConversations, Conversations
 from openai.resources.responses import AsyncResponses, Responses
 from typing_extensions import override
 
@@ -135,6 +136,15 @@ def _resolve_base_url(
 
     # Fallback to using serving endpoints
     return f"{host}/serving-endpoints"
+
+
+def _resolve_conversations_base_url(
+    workspace_client: WorkspaceClient,
+    conversations_base_url: str | None,
+) -> str:
+    if conversations_base_url is not None:
+        return conversations_base_url
+    return f"{workspace_client.config.host}/api/2.1/unity-catalog"
 
 
 def _get_authorized_http_client(workspace_client: WorkspaceClient) -> Client:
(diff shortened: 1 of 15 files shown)


integrations/openai/src/databricks_openai/utils/clients.py
```diff
@@ -425,6 +441,20 @@ def responses(self) -> Responses:
             self._databricks_responses = DatabricksResponses(self, self._workspace_client)
         return self._databricks_responses
 
+    @property
+    def conversations(self) -> Conversations:
+        if not hasattr(self, "_databricks_conversations_client"):
+            self._databricks_conversations_client = OpenAI(
```
aravind-segu, line 447: we should use DatabricksOpenAI here right so workspace client authorization flows through?

Issue #433, opened 2026-06-06
databricks-openai 0.15.0 import fails with databricks-vectorsearch 0.74: cannot import VectorSearchIndex


mdbecker:
### Summary

I also submitted this to [the community hub](https://community.databricks.com/t5/generative-ai/official-databricks-openai-package-fails-to-import-after/td-p/158563) since it's unclear if these github issues are actively monitored and triaged.

`from databricks_openai import DatabricksOpenAI` fails in a clean Python 3.12 environment after installing `databricks-openai==0.15.0` with normal dependency resolution.

On a fresh install, `databricks-openai==0.15.0` resolves `databricks-vectorsearch==0.74`. Importing `DatabricksOpenAI` then eagerly imports `VectorSearchRetrieverTool`, which imports `VectorSearchIndex` from `databricks.vector_search.client`. In `databricks-vectorsearch==0.74`, that symbol does not appear to be available; the documented/public compatibility import appears to be `VectorSearchClient`.

`pip check` reports no broken requirements.

This appears to be more than a local compatibility issue: the default installation path for an official Databricks package currently produces a broken import in a clean environment. That raises a broader concern about release hygiene, CI coverage, source-code traceability, issue-tracker ownership, and whether these packages are suitable for production use without strict downstream pinning and independent smoke testing.

### Environment

```bash
python --version
# Python 3.12.x

python -m pip show databricks-openai databricks-vectorsearch databricks-ai-search databricks-sdk openai
# databricks-openai==0.15.0
# databricks-vectorsearch==0.74
# databricks-ai-search==0.73
# databricks-sdk==0.94.0
# openai==2.41.0

python -m pip check
# No broken requirements found.
```

### Minimal reproduction

```bash
python -m venv /tmp/db-openai-repro
source /tmp/db-openai-repro/bin/activate
python -m pip install --upgrade pip
python -m pip install databricks-openai==0.15.0
python -m pip show databricks-openai databricks-vectorsearch databricks-ai-search
python -m pip check

python - <<'PY'
from databricks_openai import DatabricksOpenAI
print(DatabricksOpenAI)
PY
```

### Expected behavior

`from databricks_openai import DatabricksOpenAI` succeeds.

### Actual behavior

```text
ImportError: cannot import name 'VectorSearchIndex' from 'databricks.vector_search.client'
```

Traceback excerpt:

```text
from databricks_openai import DatabricksOpenAI
...
[...]

PR #440, opened 2026-06-10
Lazily re-export databricks_openai package root symbols

mdbecker:
## Summary

This PR preserves the existing `databricks_openai` package-root API while resolving exported symbols lazily.

The main motivation is to make narrow package-root imports like this avoid loading unrelated package-root re-exports:

```python
from databricks_openai import DatabricksOpenAI
```

`DatabricksOpenAI` is implemented in `databricks_openai.utils.clients`. It does not require the MCP toolkit, vector search retriever tool, or Unity Catalog toolkit modules to be imported. Before this change, importing `DatabricksOpenAI` from the package root also eagerly imported those unrelated package-root re-exports because they were imported at package import time.

After this change, each package-root export is imported only when that symbol is accessed. The public aliases in `__all__` are unchanged.

This PR also updates the existing package-root import smoke check so it is collected and reported as a normal pytest test item.

## Changes

* Replace eager package-root imports in `databricks_openai/__init__.py` with lazy re-exports using `_LAZY_EXPORTS` and module-level `__getattr__`.
* Preserve the existing public aliases in `__all__`.
* Cache resolved aliases in module globals after first access.
* Keep `__dir__` aligned with module globals and `__all__`.
* Update the docstring alias list to include aliases that were already exported before this PR.
* Wrap the existing package-root import smoke assertions in `test_imports.py` in a `test_` function so the smoke check is collected and reported as a normal pytest test item.
* Add focused lazy-import regression tests covering:

  * `import databricks_openai` does not eagerly import MCP/vector/UC integration modules.
  * Accessing `databricks_openai.DatabricksOpenAI` loads the client module without importing unrelated integrations.

## References

* PEP 562 defines module-level `__getattr__` and `__dir__`, which is the mechanism used here for lazy package-root attribute access: [https://peps.python.org/pep-0562/](https://peps.python.org/pep-0562/)
* Pytest standard discovery collects `test_` functions/methods from `test_*.py` files. The existing package-root smoke check is now wrapped in a `test_` function so it appears as a collected and reportable test item: [https://docs.pytest.org/en/stable/explanation/goodpractices.html#conventions-for-python-test-discovery](https://docs.pytest.org/en/stable/explanation/goodpractices.html#conventions-for-python-test-discovery)

## Testing

Validated with:

```bash
uv run --exact --group tests pytest --cov --cov-report=term tests/unit_tests
uv run --exact --group tests pytest tests/unit_tests/test_imports.py -q
uv run --exact --group tests pytest tests/unit_tests/test_lazy_imports.py
uv run ruff check src tests
```

### DIFF
--- a/integrations/openai/src/databricks_openai/__init__.py
+++ b/integrations/openai/src/databricks_openai/__init__.py
@@ -8,17 +8,17 @@ Available aliases:
 - :class:`databricks_openai.UCFunctionToolkit`
 - :class:`databricks_openai.DatabricksFunctionClient`
 - :func:`databricks_openai.set_uc_function_client`
+- :class:`databricks_openai.DatabricksOpenAI`
+- :class:`databricks_openai.AsyncDatabricksOpenAI`
+- :class:`databricks_openai.McpServerToolkit`
+- :class:`databricks_openai.ToolInfo`
+- :class:`databricks_openai.VectorSearchRetrieverTool`
 
 Refer to the Unity Catalog `documentation <https://docs.unitycatalog.io/ai/integrations/openai/#using-unity-catalog-ai-with-the-openai-sdk>`_ for more information.
 """
 
-from unitycatalog.ai.core.base import set_uc_function_client
-from unitycatalog.ai.core.databricks import DatabricksFunctionClient
-from unitycatalog.ai.openai.toolkit import UCFunctionToolkit
-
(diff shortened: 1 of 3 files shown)


mdbecker:
@jennsun  @annzhang-db @kunyuchen this might be helpful in preventing issues like #433 moving forward

PR #443, opened 2026-06-14
genie: return final summary text from _parse_attachments

nisha2003:
## Summary

`Genie.poll_for_result` was surfacing only the trailing follow-up/clarifying question instead of the full answer for some Genie responses. `_parse_attachments` (since #419) picked the *first* text attachment following the last query, but internally-created messages emit a **follow-up question text attachment before the final summary**. The follow-up has an `attachment_id`; the final summary text attachment has **no** `attachment_id`.

This change makes `_parse_attachments`:
- Drop text emitted **strictly between the first and last query** (superseded self-correction attempts), keeping leading and trailing text.
- Among the kept text attachments, return the **final summary** — the text attachment **without** an `attachment_id` — falling back to the first kept text attachment when no id-less attachment is present.

Identifying the summary by the absence of `attachment_id` matches the agreed short-term approach from the incident discussion (the FinalSummaryAttachment is the only public-API attachment without an id).

## Tests

- Added a parametrized `_parse_attachments` case: follow-up question (with `attachment_id`) emitted before the id-less summary → the summary wins.
- Full `tests/databricks_ai_bridge/test_genie.py` suite passes (75 passed).

Context: ES-1953657. Supersedes the concat-all approach in #432.

### DIFF
--- a/src/databricks_ai_bridge/genie.py
+++ b/src/databricks_ai_bridge/genie.py
@@ -195,24 +195,26 @@ def _parse_attachments(resp: Dict[str, Any]) -> Dict[str, Any]:
         "suggested_questions_attachment": None,
     }
 
-    attachments = [a for a in (resp.get("attachments") or []) if isinstance(a, dict)]
+    attachments = resp.get("attachments") or []
+    if not isinstance(attachments, list):
+        return result
 
     for a in attachments:
+        if not isinstance(a, dict):
+            continue
+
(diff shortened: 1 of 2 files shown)


jennsun:
@nisha2003 the openai_test failures are known to be breaking on master but could you look into fixing the failures for Linting / typechecking for integrations/langchain (pull_request)?

PR #445, opened 2026-06-23
Enable HTTP redirect following in DatabricksOpenAI clients

annzhang-db:
## Summary
- `DatabricksOpenAI` and `AsyncDatabricksOpenAI` pass a custom httpx `http_client` to the OpenAI SDK. The OpenAI SDK only sets `follow_redirects=True` on the client it builds itself; when a custom client is supplied, httpx's own default of `follow_redirects=False` takes over, silently disabling redirect following.
- This restores the OpenAI SDK's intended default by building the authorized httpx clients with `follow_redirects=True`, applied to both the sync and async helpers for consistent behavior.
- Exposed as a `follow_redirects: bool = True` parameter on both client classes so callers can opt out if needed.

## Test plan
- [ ] `uv run pytest tests/unit_tests/test_clients.py` (added tests assert the default is `True` and can be overridden to `False` for both sync and async helpers)

> Note: the test suite could not be run locally in this environment (no network access to sync dependencies). Verified that httpx's `Client`/`AsyncClient` default `follow_redirects` to `False` and that the modules compile.

### DIFF
--- a/integrations/openai/src/databricks_openai/utils/clients.py
+++ b/integrations/openai/src/databricks_openai/utils/clients.py
@@ -137,14 +137,18 @@ def _resolve_base_url(
     return f"{host}/serving-endpoints"
 
 
-def _get_authorized_http_client(workspace_client: WorkspaceClient) -> Client:
+def _get_authorized_http_client(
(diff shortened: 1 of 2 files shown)


jennsun (approved):
lgtm!

PR #450, opened 2026-06-30
Support URL elicitation on MCP tool calls

artjen:
## Summary

Adds **URL-mode elicitation** to `DatabricksMCPClient` in an embedding-safe way. The MCP SDK ships no default way to surface elicitation to users, so this makes the client's elicitation callback injectable.

Because `databricks-mcp` is frequently embedded in custom agents — often headless (Model Serving) or async — an interactive `input()`/`webbrowser` callback can't be the imposed default: it silently cancels where there's no TTY and blocks the event loop from `acall_tool`. So:

- **`DatabricksMCPClient(..., elicitation_callback=None)`** — new optional param. Default `None` means the `ClientSession` advertises no elicitation capability (safe for headless/async). Agents inject their own callback to surface the URL through their own channel (UI, response stream, approval card).
- **`interactive_url_elicitation_callback`** (exported) — a ready-made opt-in handler for local/CLI use: scheme-validates the server URL (http/https only), shows a security warning + confirmation prompt, and opens the approved URL. Its blocking prompt/browser runs off the event loop via `run_in_executor`, so it's safe from async code. URL-mode only; form-mode is rejected with `INVALID_REQUEST`.

Scope is the core client only; threading the callback through the OpenAI/LangChain wrappers is a follow-up.

The URL-elicitation handling draws on the MCP Python SDK's reference example: https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/clients/url_elicitation_client.py

## How do you know it works?

`TestUrlElicitation` covers: accept/decline/cancel, no-stdin (`EOFError`), disallowed-scheme rejection, form-mode rejection, elicitation off by default, an injected callback being stored, and both session paths forwarding either `None` (default) or the injected callback. Full unit suite passes under both `uv` resolutions (`lowest-direct` and `highest`) — **133 passed** — and `ruff check`/`format` are clean.

Also bumps the `mcp` floor to `>=1.23.0` (the release that introduced URL elicitation), so the `lowest-direct` CI resolution has the required symbols.

### DIFF
--- a/databricks_mcp/README.md
+++ b/databricks_mcp/README.md
@@ -18,6 +18,28 @@ pip install git+https://git@github.com/databricks/databricks-ai-bridge.git#subdi
 
 - **OAuth Provider**: Enables authentication across Databricks Notebooks, Model Serving, and local environments using the Databricks CLI.
 
+## URL elicitation
+
+An MCP server may ask the user to open a URL mid tool call (for example, to complete an auth flow). `DatabricksMCPClient` supports this, but **elicitation is off by default** — safe for agents embedded in headless or async runtimes (Model Serving, notebooks), where a stdin prompt or browser launch is meaningless and would block the event loop.
+
+To opt in, pass an `elicitation_callback`. Agents should supply their own callback that surfaces the URL through their own channel (UI, response stream, approval card):
+
+```python
+from databricks_mcp import DatabricksMCPClient
+
+client = DatabricksMCPClient(server_url, elicitation_callback=my_callback)
+```
+
+For local or CLI use, a ready-made interactive handler is provided. It prints a security warning, asks for confirmation on stdin, and opens the approved URL in a browser (running the blocking prompt off the event loop so it is safe from async code). Only URL-mode elicitation is supported; form-mode requests are rejected:
+
+```python
+from databricks_mcp import DatabricksMCPClient, interactive_url_elicitation_callback
+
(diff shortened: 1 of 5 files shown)


artjen:
### Pre-existing CI failures (unrelated to this PR)

This PR only touches `databricks_mcp/`. The four red checks below fail independently of this change — they reproduce on `main` at this branch's base commit `59f216a`:

- **`openai_test (3.10, v0.3.0 / v0.4.0 / v0.5.0)`** — `integrations/openai/tests/unit_tests/test_vector_search_retriever_tool.py` fails with `MlflowException: Reading Databricks credential configuration failed ... tracking URI 'None'`. This is an environment/credential issue in the OpenAI vector-search tests, in a package this PR does not modify. (The `openai_test` `lowest-direct` and `highest` jobs pass.)
- **`typechecking for integrations/langchain`** — `ty` reports `unresolved-attribute` on `timedelta.seconds` at `tests/unit_tests/test_multi_server_mcp_client.py:54`, a file this PR does not touch.

All checks that exercise this change are green: `mcp_test (lowest-direct)`, `mcp_test (highest)`, `ruff check for databricks_mcp`, and `typechecking for databricks_mcp`.

For reference, the one regression this PR did introduce — `mcp_test (lowest-direct)` failing because URL elicitation requires `mcp>=1.23.0` — was fixed by bumping the dependency floor.

PR #452, opened 2026-07-12
feat: Add support for serialization and tracing metadata params

jacoblee93:
Adds support for serialization of runs and populates useful metadata parameters for tracing

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -25,7 +25,7 @@ from typing import (
 from databricks import sdk
 from langchain_core.callbacks import AsyncCallbackManagerForLLMRun, CallbackManagerForLLMRun
 from langchain_core.language_models import BaseChatModel
-from langchain_core.language_models.base import LanguageModelInput
+from langchain_core.language_models.base import LangSmithParams, LanguageModelInput
 from langchain_core.messages import (
     AIMessage,
     AIMessageChunk,
@@ -66,7 +66,7 @@ from openai import AsyncOpenAI, AsyncStream, OpenAI, Stream
 from openai.types.chat import ChatCompletion, ChatCompletionChunk
 from openai.types.completion_usage import CompletionUsage
 from openai.types.responses import Response, ResponseStreamEvent, ResponseUsage
-from pydantic import BaseModel, ConfigDict, Field, model_validator
+from pydantic import AliasChoices, BaseModel, ConfigDict, Field, SecretStr, model_validator
 from typing_extensions import override
 
 from databricks_langchain.utils import get_async_openai_client, get_openai_client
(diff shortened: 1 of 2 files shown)


prasadkona:
Just documenting one behavioral effect of this PR for anyone reading later — no action needed:

This change causes the **Databricks workspace host to be emitted into trace/run metadata on every invocation**. The path:

- `_get_invocation_params` now adds a `databricks_host` key (via `_get_databricks_host()`).
- `BaseChatModel.generate()` passes that dict to `callback_manager.on_chat_model_start(..., invocation_params=params, ...)`.
- `on_chat_model_start` folds `invocation_params` into the run's `extra` metadata, which tracers persist — e.g. LangSmith records it under the run's `extra.invocation_params`, and MLflow's LangChain tracer captures it onto the span.

Two clarifications:

- It flows through the general **`invocation_params`** metadata channel, **not** `_get_ls_params` — the new `_get_ls_params` only sets `ls_provider="databricks"` and does not include the host.
- It applies to **existing `workspace_client` users too**, not just callers of the new `databricks_host`/`databricks_token` fields, because `_get_databricks_host()` falls back to `self.workspace_client.config.host`. So integrations that previously did not surface the workspace URL in traces will start doing so after this change.

The host is workspace-identifying but not a secret (the token is correctly kept out of this path), so this is just a heads-up on the new trace content rather than a concern.

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -272,10 +272,29 @@ class GetPopulation(BaseModel):
...
+    @property
+    def lc_secrets(self) -> Dict[str, str]:
+        """Map the Databricks token field to its environment variable."""
+        return {"databricks_token": "DATABRICKS_TOKEN"}
+
     model: str = Field(alias="endpoint")
     """Name of Databricks Model Serving endpoint to query."""
     target_uri: Optional[str] = None
     """The target MLflow deployment URI to use. Deprecated: use workspace_client instead."""
+    databricks_host: Optional[str] = None
```
aravind-segu, line 294: We ask users to pass in the workspace_client, is there a specific reason to explicitly pass in databricks_host and token ?

aravind-segu (commented):
Added a quick question. Did we do an E2E test here to see if everything works here?

integrations/langchain/src/databricks_langchain/chat_models.py
jacoblee93: Our service must recreate chat model instances from serialized chat model runs. We cannot easily represent complex objects like client instances, so it's convenient here to add an initialization path that uses only simple primitives like strings.

jacoblee93:
Yes, have tested end to end

aravind-segu (approved):
Thanks for the change, lgtm!

aravind-segu (approved):

aravind-segu:
Thanks for the clarifications offline!

integrations/langchain/src/databricks_langchain/chat_models.py
```diff
@@ -366,6 +385,18 @@ def async_client(self) -> AsyncOpenAI:
     def __init__(self, **kwargs: Any):
         super().__init__(**kwargs)
 
+        if self.workspace_client is None and (
```
annzhang-db, line 388: can we raise an error if both `workspace_client` and either of `databricks_host` or `databricks_token` is provided?

annzhang-db (approved):
can you rebase this pr so the tests can run again? after that I can help you merge! @jacoblee93

PR #455, opened 2026-07-21
fix: Fix returned Anthropic reasoning blocks

jacoblee93:
Preserve structured content blocks returned by Databricks chat models instead of JSON-serializing them.

This fixes Anthropic reasoning responses where the initial streamed chunk contains a structured reasoning block and subsequent chunks contain answer text. Previously, LangChain combined these into a mixed list:

```python
[reasoning_block, "answer text"]
```

## Changes

- Preserve list-based structured content in non-streaming and streaming message conversion.
- After a structured streaming chunk is received, convert subsequent text deltas into indexed text blocks:
```python
  [
      reasoning_block,
      {"type": "text", "text": "answer text", "index": 1},
  ]
```

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -902,6 +902,7 @@ class ChatDatabricks(BaseChatModel):
                 usage_chunk_emitted = True
         else:
             first_chunk_role = None
+            text_block_index = None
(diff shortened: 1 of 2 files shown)


annzhang-db (commented):
generally looks good, can we introduce something like `output_version` that langchain uses to make this behavior opt-in? just want to avoid breaking anything given the original [pr](https://github.com/databricks/databricks-ai-bridge/pull/184?timeline_per_page=5) for the string serialization mentions some downstream consumers

Issue #446, opened 2026-06-23
ChatDatabricks: UsageMetadata inflates input_tokens when prompt caching is used


anguyen:
`UsageMetadata` emitted by `ChatDatabricks` double-counts cached input tokens in `UsageMetadata.input_tokens`. An example of this (adapted from the [LangChain docs](https://docs.langchain.com/oss/python/integrations/chat/anthropic#explicit-cache-breakpoints)) is:

```python
import os

import requests
from databricks.sdk import WorkspaceClient
from databricks_langchain import ChatDatabricks
from langchain_core.messages import AnyMessage, HumanMessage, SystemMessage

def create_model() -> ChatDatabricks:
    workspace_client = WorkspaceClient(
        host=os.getenv("DATABRICKS_HOST"),
        client_id=os.getenv("DATABRICKS_CLIENT_ID"),
        client_secret=os.getenv("DATABRICKS_CLIENT_SECRET"),
        auth_type="oauth-m2m",
    )
    chat_databricks = ChatDatabricks(
        workspace_client=workspace_client,
        endpoint=os.getenv("DATABRICKS_MODEL_ENDPOINT"),
        temperature=0,
        max_tokens=1024,
    )
    return chat_databricks

def create_messages() -> list[AnyMessage]:
    get_response = requests.get("https://raw.githubusercontent.com/langchain-ai/langchain/b476fdb54aa6e6f5f0b24a68c2f4a94e43b369f9/README.md")
    get_response.raise_for_status()
    readme = get_response.text

    return [
        SystemMessage(content=[
            {
                "type": "text",
                "text": "You are a technology expert.",
            },
            {
                "type": "text",
                "text": f"{readme}",
                "cache_control": {"type": "ephemeral"},
            },
        ]),
        HumanMessage(content="What is Langchain?"),
    ]

def main():
    chat_databricks = create_model()
    messages = create_messages()
    
    databricks_response = chat_databricks.invoke(messages)
    print(databricks_response.usage_metadata)

if __name__ == "__main__":
    main()
```

When run using Claude Sonnet 4.5 as the model, the result is:
```
{'input_tokens': 5654, 'output_tokens': 399, 'total_tokens': 3232, 'input_token_details': {'cache_creation': 2821, 'cache_read': 0}}
```

As you can see, the `input_tokens` plus `output_tokens` does not equal `total_tokens`. Subtracting the `cache_creation` count from `input_tokens` fixes this discrepancy.

PR #456, opened 2026-07-23
Forward cache_control from message-level placement in ChatDatabricks

adamgurary:
## Context

This fix came out of a hands-on end-to-end evaluation of building, tracing, and improving a LangGraph agent on Databricks.

## Problem: ChatDatabricks drops cache_control unless it is inside a typed content block

`ChatDatabricks` forwards an Anthropic `cache_control` breakpoint only when it is nested inside a typed content block. A message-level key or an `additional_kwargs` placement is dropped silently, with no error, so prompt caching never activates for those placements.

## Steps to reproduce

Send the same large static system block three ways through the same `ChatDatabricks` client and check warm-call `cache_read` per placement:

```
message-level sibling key {"role","content":<str>,"cache_control":...} : cache_read = 0       -> DROPPED
SystemMessage(additional_kwargs={"cache_control":...})                 : cache_read = 0       -> DROPPED
typed content block [{"type":"text","text":...,"cache_control":...}]   : cache_read = 8803    -> FORWARDED
```

## Root cause

`_convert_message_to_dict` (v0.17.0) builds the outbound message as `{"content": message.content}` and copies nothing else from `additional_kwargs`. So `cache_control` survives only when it is nested inside a content block that lives in `message.content`.

## Fix

`_convert_message_to_dict` now forwards a message-level `cache_control` from `additional_kwargs` onto the outbound message dict. The forwarding fires only when `cache_control` is present, so the no-breakpoint path is unchanged.

```python
if (cache_control := message.additional_kwargs.get("cache_control")) is not None:
    message_dict["cache_control"] = cache_control
```

Two unit tests cover the present and absent cases.

This is a distinct bug from #446 (token-count inflation when caching is active). This PR addresses the breakpoint being dropped so caching never activates.

### DIFF
--- a/integrations/langchain/src/databricks_langchain/chat_models.py
+++ b/integrations/langchain/src/databricks_langchain/chat_models.py
@@ -1419,6 +1419,14 @@ def _convert_message_to_dict(message: BaseMessage) -> dict:
     # if (name := message.name or message.additional_kwargs.get("name")) is not None:
     #     message_dict["name"] = name
 
+    # Forward an Anthropic prompt-caching breakpoint placed at the message level.
+    # LangChain stores a message-level `cache_control` in `additional_kwargs`. Without
+    # forwarding it here, the breakpoint reaches the endpoint only when it is nested
+    # inside a typed content block, so prompt caching silently never activates for the
+    # other natural placements.
+    if (cache_control := message.additional_kwargs.get("cache_control")) is not None:
+        message_dict["cache_control"] = cache_control
+
     if isinstance(message, ChatMessage):
         return {"role": message.role, **message_dict}
     elif isinstance(message, HumanMessage):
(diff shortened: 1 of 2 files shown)


adamgurary:
Hey @fanzeyi and @aravind-segu, this PR fixes a silent drop of message-level `cache_control` in `_convert_message_to_dict`. Two unit tests added. Would appreciate a review when you have a moment.

adamgurary:
One thing on my side is blocked: the test workflow is waiting on a maintainer to click "Approve and run workflows" (first PR I've opened here, so the runs are sitting at action_required). Once that runs, CI should be green. The two new unit tests cover the present and absent `cache_control` cases and pass locally with 108 passed / 2 skipped.

adamgurary:
Hi @fanzeyi @aravind-segu, following up from Aug 3. CI here is gated on a maintainer clicking "Approve and run workflows" in the Actions tab (first-time-contributor fork gate), so all checks are sitting at action_required. Could one of you approve the workflow run so this can get a review?

PR #467, opened 2026-08-19
Generalize long-running agent recovery and replay

shivam5:

**Change:** Generalize `LongRunningAgentServer` around two explicit crash-recovery strategies while keeping durable background execution and SSE replay in the server.

- `ResumeStrategy.EVENT_LOG` reconstructs recovery context from the immediately previous attempt’s durable events and rotates the agent session to `<session-key>::attempt-N`.
- `ResumeStrategy.AGENT_SESSION` keeps the same agent session key and sends a fixed recovery prompt; the agent SDK/harness restores its own transcript.
- Heartbeats, stale-attempt detection, atomic claiming, request persistence, terminal-response persistence, and ordered SSE event persistence/replay are server-owned behavior for both strategies.
- An optional `@on_resume()` request transformer runs only after this process wins a stale-attempt claim. Without an override, `ResumeContext.default_resume_request()` applies the selected strategy. The resulting request then goes through the same `@invoke()` or `@stream()` handler mode as the original attempt; normal handlers do not contain recovery branches.
- Agent-session recovery accepts `context.conversation_id`, `custom_inputs.session_id`, or `custom_inputs.thread_id` as the harness session anchor. If all are absent, the server warns and injects the generated durable `response_id` as `context.conversation_id`.
- Event-log recovery requires `@stream()`, because the streamed events are its recovery input. Agent-session recovery preserves the original invoke/stream handler mode.
- Completed and failed terminal Responses payloads are stored directly so polling does not reconstruct authoritative results from stream fragments.
- Public OpenAI Agents SDK and LangGraph cookbooks use one shared, restart-safe App entry point with separate bundle targets for the two recovery strategies.
- Lakebase connections honor the Databricks Apps `PGDATABASE` resource binding instead of always connecting to `databricks_postgres`.
- SSE retrieval advances past sequence `0` correctly instead of replaying the first event on every polling iteration.

**Why:** The earlier implementation mixed runtime durability, SSE replay, and agent conversation recovery. This makes the ownership boundaries explicit and lets an agent harness use its native session persistence instead of forcing every recovery through generated event-log prose.

### Recovery contract

| Strategy | Input passed to the resumed agent loop | Session behavior | Event-log use |
| --- | --- | --- | --- |
| `EVENT_LOG` | Original request plus `[RECOVERY]` prose containing the previous attempt’s durable events | Rotates to `<session-key>::attempt-N` | SSE replay and recovery context |
| `AGENT_SESSION` | Fixed `[RECOVERY]` prompt | Reuses the same harness session so the SDK reloads its transcript | SSE replay only |

Recovery restarts a handler; it does not resume a suspended Python coroutine. Interrupted side effects are therefore at-least-once and tools still need idempotency.

### Three separate stores

The implementation has three logically independent stores. They may share one Lakebase database, but they have different owners and purposes.

| Store | Owner | Physical tables | Purpose |
| --- | --- | --- | --- |
| Runtime durability | `LongRunningAgentServer` | `agent_server.responses` | Durable `response_id`, original request, terminal response, status, heartbeat, attempt number, and original invoke/stream mode |
| Durable event log | `LongRunningAgentServer` | `agent_server.messages` | Ordered stream events/output items used for `starting_after` SSE replay; also read by `EVENT_LOG` recovery |
| Agent session store | Agent SDK/harness | Example: `<session schema>.agent_sessions` and `.agent_messages` | Logical conversation transcript used by `AGENT_SESSION` recovery; the server neither creates nor interprets this transcript |

`response_id` identifies the durable HTTP operation. The harness session ID identifies conversational history. They are not inherently the same. For agent-session recovery, the selected session anchor remains inside the persisted `original_request`; it is not a separate durability-table column.

### Public cookbooks

The PR contains the same small wait-tool agent in two harnesses:

1. `cookbooks/openai-sdk-agent`
2. `cookbooks/langgraph-agent`

Each cookbook has `event_log` and `agent_session` bundle targets. A root
`app.yaml` and shared App entry point keep the command and resource bindings
available after `databricks apps stop` / `start`; the stable App name selects the
same recovery strategy and SDK-owned schema on restart. Handler and agent logic
remain shared, while the small strategy-specific entry points show the direct
wiring. Deployment helpers and test-result artifacts are not part of the PR.

### Automated validation

```bash
uv run pytest -q \
  tests/databricks_ai_bridge/test_long_running_db.py \
  tests/databricks_ai_bridge/test_long_running_server.py \
  tests/databricks_ai_bridge/test_lakebase.py
# 238 passed

uv run pytest -q tests/databricks_ai_bridge \
  --ignore=tests/databricks_ai_bridge/test_model_serving_obo_credential_strategy.py
# 329 passed, 120 skipped
# The ignored OBO credential-strategy test is unrelated to this change.

uv run ruff format --check .
uv run ruff check .
uv run ty check
# All exit 0; ty reports only three pre-existing unused-ignore warnings.
```

App build logs also confirm dependencies install successfully through the Databricks PyPI proxy.

### Deployed environments

Both latest deployments are `RUNNING`, their active deployments are `SUCCEEDED`, and their dedicated Lakebase branches are `READY`.

| Strategy | Databricks App | Runtime URL | Lakebase tables |
| --- | --- | --- | --- |
| Event-log recovery | [App overview](https://eng-ml-inference.staging.cloud.databricks.com/apps-v2/app/openai-agent-framework-sse/overview?o=1653573648247579) | https://openai-agent-framework-sse-1653573648247579.staging.aws.databricksapps.com | [`durable-framework-sse`](https://eng-ml-inference.staging.cloud.databricks.com/lakebase/projects/18e258e0-0c3c-4cf4-aa99-c5a8452b25ef/branches/br-falling-rice-y2moig3m/tables) |
| Agent-session recovery | [App overview](https://eng-ml-inference.staging.cloud.databricks.com/apps-v2/app/openai-agent-session-sse/overview?o=1653573648247579) | https://openai-agent-session-sse-1653573648247579.staging.aws.databricksapps.com | [`durable-session-sse`](https://eng-ml-inference.staging.cloud.databricks.com/lakebase/projects/18e258e0-0c3c-4cf4-aa99-c5a8452b25ef/branches/br-withered-snow-y24p21r1/tables) |

Platform state was checked with:

```bash
env -u DATABRICKS_HOST -u DATABRICKS_TOKEN -u DATABRICKS_CONFIG_PROFILE \
  databricks apps get <app> --profile ml-inference-staging -o json

env -u DATABRICKS_HOST -u DATABRICKS_TOKEN -u DATABRICKS_CONFIG_PROFILE \
  databricks postgres get-branch \
  projects/shivam-openai-agent-on-apps/branches/<branch> \
  --profile ml-inference-staging -o json
```

### Client protocol exercised

All live tests used the public `POST /responses`, `GET /responses/{response_id}`, and SSE replay contract:

```bash
curl -N -X POST "$APP_URL/responses" \
  -H "Authorization: Bearer $APP_TOKEN" \
  --data '{
    "background": true,
    "stream": true,
    "input": [{"role": "user", "content": "<long tool-call request>"}],
    "custom_inputs": {"session_id": "<manual test session>"}
  }'

curl -N \
  "$APP_URL/responses/$RESPONSE_ID?stream=true&starting_after=$LAST_SEQUENCE" \
  -H "Authorization: Bearer $APP_TOKEN"

curl "$APP_URL/responses/$RESPONSE_ID" \
  -H "Authorization: Bearer $APP_TOKEN"
```

#### Happy path and disconnect/replay

- Event-log App: response `resp_b6e5d83e6b1f44549a8fae9b` completed on attempt 1. For response `resp_bba2c7517912463abf54b2ed`, the client disconnected after sequence 1; replay returned sequences 2–35, `response.completed`, and `[DONE]`.
- Agent-session App: response `resp_e2da1b63155b4280902b6409` completed on attempt 1. For response `resp_8c1da65aafc14e269fd4c87b`, the client disconnected after sequence 10; replay returned sequences 11–36, `response.completed`, and `[DONE]`.
- Polling the completed durable `response_id` returned the stored terminal Responses payload in both Apps.

#### Real App stop/start: event-log recovery

Input session: `pr467-event-crash-20260821T184027Z`  
Durable response: `resp_1b2c026380fb4fb2a7c1f3d8`

1. The client started a 30-second tool call, captured events through sequence 10, and intentionally disconnected after six seconds (`curl: (28)`, HTTP 200 with 8,579 bytes already received).
2. `databricks apps stop openai-agent-framework-sse --profile ml-inference-staging` stopped compute while the tool was still running.
3. Lakebase after the stop showed `in_progress`, attempt 1, no terminal response, 11 event rows (sequences 0–10), and one message in the original SDK session.
4. After `databricks apps start ...`, stale-heartbeat claiming advanced the row to attempt 2.
5. Reconnect from cursor 10 returned sequences 11–35. The first event was `response.resumed`; the stream ended with `response.completed` and `[DONE]`.
6. Final Lakebase state is `completed`, attempt 2. Attempt 1 has 11 event rows; attempt 2 has 25. The SDK session rotated to `pr467-event-crash-20260821T184027Z::attempt-2`.
7. The resumed loop received the original user input plus an 8,444-character `[RECOVERY]` input containing serialized prior event/tool-call evidence and produced `PR467_EVENT_CRASH_RECOVERED_20260821T184027Z`.

#### Real App stop/start: agent-session recovery

Input session: `pr467-session-crash-20260821T184221Z`  
Durable response: `resp_516816b5e26f49a4a98248f1`

1. The same 30-second tool-call scenario was stopped with attempt 1 still `in_progress`; Lakebase had 11 event rows (sequences 0–10), one SDK user message, and no terminal response.
2. Restart claimed attempt 2 without rotating the session ID.
3. Correct cursor replay:
   ```bash
   curl -N \
     "$APP_URL/responses/resp_516816b5e26f49a4a98248f1?stream=true&starting_after=10" \
     -H "Authorization: Bearer $APP_TOKEN"
   ```
   returned HTTP 200, 37 JSON events (sequences 11–47), then `[DONE]`. The first event was `response.resumed`; the last was `response.completed`.
4. Final `agent_server.responses` state is `completed`, attempt 2, with a persisted terminal response. Event rows are 11 for attempt 1 and 37 for attempt 2.
5. The same SDK session contains five messages: the original user input, the fixed 255-character `[RECOVERY]` prompt, the resumed tool call, its output, and the final assistant message. No `::attempt-2` session exists.
6. The client received `PR467_SESSION_CRASH_RECOVERED_20260821T184221Z`, demonstrating that harness-owned history was restored and the resumed loop completed.

The Lakebase checks queried `agent_server.responses`, grouped `agent_server.messages` by attempt/sequence, and inspected the SDK-owned `agent_messages` rows for the exact session IDs. No test-result files are added to the open-source diff; the evidence is recorded here.

### LangGraph durability validation — August 25, 2026

Both LangGraph cookbook targets were deployed in `ml-inference-staging` using the
exact local PR wheel and separate empty databases under
`projects/shivam-debug/branches/pr-cuj-harness`:

| Strategy | App | Database |
| --- | --- | --- |
| Event log | `langgraph-event-recovery` | `pr467-langgraph-event` |
| Agent session | `langgraph-session-recovery` | `pr467-langgraph-session` |

The test for each App used `background=true`, `stream=true`, a 60-second
`wait_for_completion` tool call, an actual `databricks apps stop` while the tool
was pending, `databricks apps start`, SSE reconnection, final polling, and a
follow-up conversational turn.

#### LangGraph EVENT_LOG recovery

Original conversation: `pr467-langgraph-event_log-20260825T184118Z`  
Durable response: `resp_8d13bf3284084485a85bffe7`

1. Attempt 1 persisted one event: sequence 0, `response.output_item.done` with a `function_call`. The App was stopped before a tool output existed.
2. Restart claimed attempt 2. Sequence 1 was `response.resumed` with rotated conversation ID `pr467-langgraph-event_log-20260825T184118Z::attempt-2`.
3. Attempt 2 persisted four rows (sequences 1–4): `response.resumed`, a new `function_call`, `function_call_output`, and the final assistant `message`. This shows prose recovery started a fresh LangGraph thread and re-invoked the interrupted tool.
4. Final polling returned `completed`, attempt 2, a persisted terminal response, and `PR467_LANGGRAPH_EVENT_LOG_20260825T184118Z`.
5. The original LangGraph thread has 3 checkpoints; the rotated attempt-2 thread has 8 checkpoints.
6. The client then sent a new turn using the rotated ID from `response.resumed`. The response was `FOLLOWUP_REMEMBERS:PR467_LANGGRAPH_EVENT_LOG_20260825T184118Z`, proving that the client must adopt the emitted rotated conversation ID for later turns. The old ID points at the abandoned attempt-1 checkpoint.
7. The live initial SSE stream emitted sequence 0 only once. This specifically verifies the sequence-zero cursor regression fixed in this update.

#### LangGraph AGENT_SESSION recovery

Conversation: `pr467-langgraph-agent_session-20260825T184625Z`  
Durable response: `resp_387945bf600244e9ae579223`

1. Attempt 1 persisted one event: sequence 0, the pending `function_call`.
2. Restart claimed attempt 2 and emitted `response.resumed` at sequence 1 with the unchanged conversation ID.
3. Attempt 2 persisted only `response.resumed`, `function_call_output`, and the final assistant `message` (sequences 1–3). There was no second function-call event, showing `astream(None)` resumed the pending LangGraph checkpoint instead of rebuilding the turn from prose.
4. Final polling returned `completed`, attempt 2, a persisted terminal response, and `PR467_LANGGRAPH_AGENT_SESSION_20260825T184625Z`.
5. Lakebase contains one LangGraph thread with 8 checkpoints; no `::attempt-2` thread was created.
6. A follow-up using the same conversation ID returned `FOLLOWUP_REMEMBERS:PR467_LANGGRAPH_AGENT_SESSION_20260825T184625Z`.

The live deployment also caught and fixed three cookbook/runtime issues before
recording these passing runs: duplicate bundle App source paths, missing
restart-time `app.yaml` resource bindings, and Lakebase clients ignoring the
Apps-provided `PGDATABASE` value.

### Durable HITL cookbook flow

Both the OpenAI Agents SDK and LangGraph cookbooks now show the same agent-managed HITL pattern:

1. The client submits a proposal with `background=true` and `stream=true`.
2. The first durable Response completes with `APPROVAL_REQUIRED`; no worker or lease is held while waiting for a person.
3. The client submits approval as a second background streamed Response using the same SDK session or LangGraph thread.
4. The second run includes a 60-second tool call, so the App can be stopped and recovered while approved work is active.
5. Both terminal Responses and both ordered event streams are persisted. The harness session/checkpointer preserves conversational state between the proposal and approval turns.

This keeps responsibilities explicit: `LongRunningAgentServer` makes each Response durable, while the agent/harness defines the HITL protocol and stores the state needed by the next turn. Tool side effects remain at-least-once and require idempotency.

| Concern | Contract in this PR |
| --- | --- |
| Background | `background=true` creates a durable Responses record. The client receives an `in_progress` response ID and polls `GET /responses/{id}`. |
| Streaming | `@stream()` yields Responses events; the server persists them and reconnects with `?stream=true&starting_after=<sequence>`. |
| OpenAI Agents SDK client | The SDK loop remains server-side. A custom deployment client adopts Responses; an existing Responses-compatible client keeps its shape and changes the base URL. |
| LangGraph SDK client | This cookbook runs LangGraph behind Responses, not LangGraph Agent Server. Native threads/runs calls change to Responses unless a LangGraph protocol adapter is added. |

Both public cookbook READMEs now include concrete before/after client snippets and the exact `background=true, stream=true`, polling, and reconnect flow.


src/databricks_ai_bridge/long_running/AGENTS.md
```diff
@@ -336,17 +359,18 @@ Postgres row locking ensures only one of N concurrent UPDATEs matches the `attem
 | Concern | Where it lives | Author-visible? |
 |---|---|---|
 | Heartbeat + claim | `LongRunningAgentServer` | No |
-| Conversation_id rotation | `LongRunningAgentServer._rotate_conversation_id` | No |
-| Prose recovery message construction | `LongRunningAgentServer._build_prose_recovery_message` | No |
+| Resume request policy | `LongRunningAgentServer.on_resume` | Optional override |
+| Conversation_id rotation | Framework-managed `on_resume` | No |
```
shivam5: when auto_recovery=True

shivam5:
@smurching - Should we keep some examples in this repo as well? 
Or just keep everything in https://github.com/databricks/app-templates?

src/databricks_ai_bridge/long_running/server.py
```diff
@@ -157,6 +167,83 @@ def _age_seconds(created_at: datetime) -> float:
     return (now - created_at).total_seconds()
 
 
+def _failed_response(response_id: str, message: str, code: str) -> dict[str, Any]:
```
jamesbxwu:
```suggestion
def _failed_response(response_id: str, error_msg: str, error_code: str) -> dict[str, Any]:
```

examples/openai-sdk-agent/agent_recovery_polling/handlers.py
jamesbxwu, line 1: while it's good to demonstrate a more complicated agent implementation such as your review agent, can we have a simpler example such just a thin LLM wrapper wiring up to the lifecycle and recovery?

src/databricks_ai_bridge/long_running/AGENTS.md
```diff
@@ -336,17 +392,18 @@ Postgres row locking ensures only one of N concurrent UPDATEs matches the `attem
...
 
 The author's `agent.py` is unchanged from a non-durable agent. They construct an `AsyncCheckpointSaver` (LangGraph) or `AsyncDatabricksSession` (OpenAI) and use it normally. Durability fires entirely above the SDK boundary — the SDK adapters themselves contain zero durability code.
 
 ### 4.2 Author-visible client cooperation (chat-UI side)
 
-For the always-rotate flow to work cross-turn, a cooperating chat UI needs to:
+For framework-managed recovery's rotated-session flow to work cross-turn, a cooperating chat UI needs to:
 
 1. **Capture the rotated `conversation_id` from the SSE `response.resumed` event** when one is emitted during a streaming retrieve.
 2. **Use the rotated value as `context.conversation_id` on subsequent requests** for the same chat.
```
elainemwang: some of the mermaid diagrams aren't formatted correctly, I don't think it's due to any changes in this pr, but can we fix it if it's an issue?

Image

examples/openai-sdk-agent/agent_recovery_polling/TEST_RESULTS.md
jamesbxwu, line 1: does this belong in the public repo?

examples/openai-sdk-agent/framework_recovery_sse/TEST_RESULTS.md
jamesbxwu, line 1: +1 on publishing this to the public repo

src/databricks_ai_bridge/long_running/server.py
```diff
@@ -340,13 +427,24 @@ def __init__(
         self._db_autoscaling_endpoint = db_autoscaling_endpoint
         self._db_project = db_project
         self._db_branch = db_branch
+        self._auto_recovery = auto_recovery
+        self._sse_replay = sse_replay
+        self._event_log_enabled = auto_recovery or sse_replay
```
elainemwang: qq: why does `auto_recovery` require the event log?

examples/openai-sdk-agent/agent_recovery_polling/README.md
```diff
@@ -0,0 +1,58 @@
...
+  request translation that happens before the same handler is called.
+- [`app.py`](./app.py) directly constructs the server with both policy flags
+  set to `False`.
+
+## Physical persistence
+
+| Logical store | Physical table | What this app writes |
+| --- | --- | --- |
+| Runtime durability | `agent_server.responses` | Original request, terminal response, status, heartbeat, attempt, handler mode |
+| Durable event log | `agent_server.messages` | No rows; the table exists only for schema compatibility |
```
jamesbxwu: is it still possible to rename these to be more accurate such as `agent_server.checkpoints` and ``agent_server.events`?

examples/openai-sdk-agent/framework_recovery_sse/README.md
```diff
@@ -0,0 +1,54 @@
+# Framework Recovery With SSE Replay
```
jamesbxwu:
```suggestion
# Durable Runtime Auto Recovery with Streaming Replay
```

src/databricks_ai_bridge/long_running/AGENTS.md
```diff
@@ -14,39 +14,55 @@ This document describes:
 `LongRunningAgentServer` extends MLflow's `AgentServer` for `ResponsesAgent` handlers with three capabilities:
 
 1. **Background execution.** A `POST /responses` request with `background: true` returns a `response_id` immediately; the agent loop runs detached from the HTTP connection. State persists to Lakebase Postgres.
-2. **Streaming retrieval.** `GET /responses/{response_id}?stream=true&starting_after=N` replays events past sequence `N` and tails new ones until the run finishes. Reconnects without losing events.
-3. **Crash-resumable execution.** If the pod running an agent loop dies, another pod atomically claims the run and finishes the work via **prose recovery**: the new attempt receives a single user message containing `json.dumps(events)` of the crashed attempt's stream-event log plus a directive asking the LLM to figure out what's done vs interrupted and continue. The handler runs on a freshly-rotated SDK session.
+2. **Optional streaming retrieval.** With `sse_replay=True`, `GET /responses/{response_id}?stream=true&starting_after=N` replays events past sequence `N` and tails new ones until the run finishes.
```
jamesbxwu: For MVP, I'd say we remove this knob and always write the event log and see if customers complain

src/databricks_ai_bridge/long_running/AGENTS.md
```diff
@@ -14,39 +14,55 @@ This document describes:
...
+3. **Crash-resumable execution.** If the pod running an agent loop dies, another pod atomically claims the run. With `auto_recovery=True`, the server reconstructs prose from stream events and rotates the SDK session. With `auto_recovery=False`, the default resume policy retains the session anchor and sends only a fixed recovery prompt so the agent's session store restores the transcript.
```
jamesbxwu:
```suggestion
3. **Crash-resumable execution.** If the worker executing a specific session run dies, another worker atomically claims the run. With `auto_recovery=True`, the server reconstructs prose from stream events and rotates the SDK session. With `auto_recovery=False`, the default resume policy retains the session anchor and sends only a fixed recovery prompt so the agent's session store restores the transcript.
```

jamesbxwu: can we explain the concept of `session_anchor`?

src/databricks_ai_bridge/long_running/AGENTS.md
```diff
@@ -14,39 +14,55 @@ This document describes:
...
 Callers see one HTTP surface; the underlying SDK (LangGraph, OpenAI Agents, others) is opaque to the server.
 
 ### Guarantees
 
 - **At-most-once durable claim.** Only one pod runs a given response at a time. The handoff uses an atomic CAS on `attempt_number`.
-- **Append-only event log.** Every SSE frame is persisted to `agent_server.messages` keyed by `(response_id, attempt_number, sequence_number)`. Clients cursor-resume from `starting_after`.
-- **SDK-agnostic recovery.** The resumed attempt receives a flat prose narrative — no provider-specific tool-pair structure, no synthetic tool events, no per-SDK adapter code.
+- **Conditional append-only event log.** Events are persisted when framework recovery or SSE replay needs them: `auto_recovery or sse_replay`.
+- **Durable terminal response.** Polling reads the final response from `agent_server.responses.response`; it does not require event rows.
+- **Pluggable resume input.** An optional `@on_resume()` handler can transform the stored request; `ResumeContext.default_request()` exposes the framework-managed or agent-session-managed default.
```
jamesbxwu:
```suggestion
- **Pluggable resume input.** An optional `@on_resume()` handler can transform the stored request; `ResumeContext.default_request()` exposes the runtime-managed or agent-sdk-managed default.
```

src/databricks_ai_bridge/long_running/AGENTS.md
```diff
@@ -14,39 +14,55 @@ This document describes:
...
+### Configuration modes
+
+| `auto_recovery` | `sse_replay` | Recovery context | Event rows |
+|---|---|---|---|
+| `True` | `True` | Prior event log, rotated session | Recovery + client replay |
+| `True` | `False` | Prior event log, rotated session | Recovery only |
+| `False` | `True` | Agent session transcript, same session | Client replay only |
+| `False` | `False` | Agent session transcript, same session | None |
+
+Heartbeat, stale scanning, CAS claim, and handler restart apply to every mode.
```
jamesbxwu: as mentioned above, let's simplify by removing the sse_replay knob.

src/databricks_ai_bridge/long_running/models.py
```diff
@@ -37,6 +37,10 @@ class Response(Base):
         Integer, nullable=False, server_default="1", default=1
     )
     original_request: Mapped[str | None] = mapped_column(Text, nullable=True)
+    response: Mapped[str | None] = mapped_column(Text, nullable=True)
```
jamesbxwu: what is this used for?

src/databricks_ai_bridge/long_running/resume.py
```diff
@@ -0,0 +1,57 @@
...
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+_on_resume_function: Callable[..., Any] | None = None
+
+
+@dataclass(frozen=True)
+class ResumeContext:
+    """Metadata and default behavior available to an ``@on_resume`` handler."""
```
jamesbxwu: can we improve the documentation here to cover
1. what are previous events
2. what does _default_request include and why should this seemingly generic concept of a `ResumeContext` need to differentiate default vs not?

src/databricks_ai_bridge/long_running/server.py
```diff
@@ -47,12 +48,18 @@
...
     update_response_trace_id,
 )
+from databricks_ai_bridge.long_running.resume import ResumeContext, get_on_resume_function
 from databricks_ai_bridge.long_running.settings import LongRunningSettings
 from databricks_ai_bridge.utils.annotations import experimental
 
 logger = logging.getLogger(__name__)
 
 BACKGROUND_KEY = "background"
+AGENT_MANAGED_RECOVERY_PROMPT = (
```
jamesbxwu: what does agent managed mean? is it the same as framework managed mentioned in the readme?

src/databricks_ai_bridge/long_running/server.py
```diff
@@ -157,6 +167,83 @@ def _age_seconds(created_at: datetime) -> float:
     return (now - created_at).total_seconds()
 
 
+def _failed_response(response_id: str, message: str, code: str) -> dict[str, Any]:
+    return {
+        "id": response_id,
+        "object": "response",
+        "status": "failed",
```
jamesbxwu: nit: can status be defined as an enum?

src/databricks_ai_bridge/long_running/server.py
```diff
@@ -157,6 +167,83 @@ def _age_seconds(created_at: datetime) -> float:
...
+    if not isinstance(request_context, dict):
+        request_context = {}
+    request_context["conversation_id"] = response_id
+    request_dict["context"] = request_context
+    logger.warning(
+        "[durable] agent-managed request has no session anchor; using response_id=%s "
+        "as context.conversation_id",
+        response_id,
+    )
+    return request_dict
```
jamesbxwu: can we document what an anchor is?

jamesbxwu (commented):
Went through first pass and here are a few structural comments:
1. Can we get rid of the `sse_replay` knob and keep the existing behavior of always writing to the event log and support streaming replay?
2. Should the `examples` folder be defined at the top level vs inside of `src/long_running`? It doesn't seem that `examples` as they are today covers everything inside of the `databricks-ai-bridge` repo
3. Consider renaming `examples` to `cookbooks`
4. I found the examples to be incredibly hard to follow and their corresponding README.md to be difficult to compare.  Can we have a top level README.md in the `examples` folder that explains the different configurations for the long running agent server and have a single summary table comparing them? 
5. I feel the review agent is too complicated to understand, let's keep the initial examples/cookbook very simple and easy to follow.  At this time, we are focused on demonstrating top level wiring, lifecycle hooks and etc and less about showing a complicated review agent

src/databricks_ai_bridge/long_running/server.py
elainemwang, line 542: can we track this in a follow up to allow background mode, GET /responses etc. with an in memory database for the local dev loop?

src/databricks_ai_bridge/long_running/models.py
shivam5: In cases, where we will not store the event log, we need to store the final response to return the respone in background mode.
If it is a background non-streaming agent with agent managed recovery, we won't be storing the event logs, but need to store the final result.

src/databricks_ai_bridge/long_running/server.py
shivam5: By auto-recovery I mean the creating a new conversation id with the event logs, and not depending on session store / agent managed recovery.
I will better name it.

src/databricks_ai_bridge/long_running/server.py
shivam5: for sure, added in [standup doc](https://docs.google.com/document/d/1HVvSRd2Gy5m17qEirLpOcIbHoC8PDtEzgTje1Vzfrac/edit?tab=t.0)

src/databricks_ai_bridge/long_running/AGENTS.md
shivam5: auto_recovery was removed and replaced with explicit EVENT_LOG and AGENT_SESSION strategies

src/databricks_ai_bridge/long_running/server.py
shivam5: Renamed to EVENT_LOG recovery and AGENT_SESSION

examples/openai-sdk-agent/agent_recovery_polling/README.md
shivam5: We can, although there would be backward incompatibility. Should we think about this in a later PR and keep the names as is for now?

src/databricks_ai_bridge/long_running/AGENTS.md
shivam5: Replaced by `session key`

src/databricks_ai_bridge/long_running/server.py
shivam5: Those names were removed and replaced by EVENT_LOG and AGENT_SESSION

src/databricks_ai_bridge/long_running/server.py
shivam5: Added ResponseStatus in server.py

src/databricks_ai_bridge/long_running/server.py
shivam5: Replaced with and documented as “session key"

src/databricks_ai_bridge/long_running/resume.py
shivam5: added documentation. The default request is use the server default. If it still sounds confusing I can take a stab to simplify this.

cookbooks/openai-sdk-agent/openai_sdk_agent_shared/handlers.py
```diff
@@ -0,0 +1,48 @@
...
+from openai_sdk_agent_shared.agent import invoke_agent, stream_agent
+from openai_sdk_agent_shared.sessions import create_session
+
+
+def _session_id(request: ResponsesAgentRequest) -> str:
+    custom_inputs = dict(request.custom_inputs or {})
+    if custom_inputs.get("session_id"):
+        return str(custom_inputs["session_id"])
+    if request.context and request.context.conversation_id:
+        return request.context.conversation_id
```
jamesbxwu, line 23: this method needs a bit more explanation

src/databricks_ai_bridge/long_running/resume.py
```diff
@@ -0,0 +1,77 @@
...
+
+from mlflow.types.responses import ResponsesAgentRequest
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+_on_resume_function: Callable[..., Any] | None = None
+
+
+class ResumeStrategy(str, Enum):
```
jamesbxwu, line 19: let's rename this to "recovery" so that it is differentiated from planned pause/resume such as HITL? Let's update all references from resume -> recovery in that case. 
```suggestion
class RecoveryStrategy(str, Enum):
```

src/databricks_ai_bridge/long_running/db.py
```diff
@@ -91,6 +91,9 @@ def _set_statement_timeout(dbapi_conn, connection_record, connection_proxy):
         f"ALTER TABLE {AGENT_DB_SCHEMA}.responses "
         "ADD COLUMN IF NOT EXISTS attempt_number INTEGER NOT NULL DEFAULT 1",
         f"ALTER TABLE {AGENT_DB_SCHEMA}.responses ADD COLUMN IF NOT EXISTS original_request TEXT",
+        f"ALTER TABLE {AGENT_DB_SCHEMA}.responses ADD COLUMN IF NOT EXISTS terminal_response TEXT",
```
jamesbxwu, line 94: agent responses can be quite long and verbose, any concerns on db row/column size limits if we store this?

src/databricks_ai_bridge/long_running/server.py
jamesbxwu, line 1: this server file is becoming quite long (1500+ lines) and combines invoke entry points, heartbeat + recovery.  I personally had a hard time tracking all the changes. Can we refactor them into such logical parts for better modularity? I recognize this may be out of scope for this PR but let's discuss whether this should be deferred.

src/databricks_ai_bridge/long_running/server.py
```diff
@@ -157,6 +177,85 @@ def _age_seconds(created_at: datetime) -> float:
...
+    response["attempt_number"] = attempt_number
+    return response
+
+
+def _agent_session_recovery_message() -> dict[str, Any]:
+    return {
+        "type": "message",
+        "role": "user",
+        "content": AGENT_SESSION_RECOVERY_PROMPT,
+    }
```
jamesbxwu, line 211: instead of defining a method, why not just make this a constant similar to `AGENT_SESSION_RECOVERY_PROMPT`?

src/databricks_ai_bridge/long_running/server.py
```diff
@@ -267,15 +361,15 @@ class LongRunningAgentServer(AgentServer):
...
-    Durable resume: when ``GET /responses/{id}`` sees an ``in_progress`` run
-    whose owning pod has stopped heartbeating for more than
-    ``heartbeat_stale_threshold_seconds``, the retrieving pod atomically claims
-    the run and re-invokes the registered handler with a rotated
-    ``conversation_id`` (so the agent SDK resolves to a fresh thread/session),
-    the original request's ``input`` enriched with the prior attempt's already
-    emitted tool calls / outputs / narrative, and an ``[INTERRUPTED]`` synthetic
-    output paired with any tool call that didn't finish. Completed work is
-    preserved; only the interrupted step re-runs.
+    When a heartbeat becomes stale, another pod atomically claims the execution
```
jamesbxwu, line 364:
```suggestion
    When a heartbeat becomes stale, another worker atomically claims the execution
```

src/databricks_ai_bridge/long_running/server.py
```diff
@@ -892,9 +1018,14 @@ async def _do_background_stream(
                     stream_event={"trace_id": span.trace_id},
                     attempt_number=attempt_number,
                 )
+            if return_trace_id:
```
jamesbxwu, line 1021: is this redundant with line 1014 above?
```suggestion
```

cookbooks/openai-sdk-agent/README.md
```diff
@@ -0,0 +1,132 @@
...
+`@stream()` handler used by the original attempt.
+
+```python
+from databricks_ai_bridge.long_running import ResumeContext, on_resume
+
+
+@on_resume()
+async def resume(request, context: ResumeContext):
+    resumed = await context.default_resume_request(request)
+    return resumed
```
jamesbxwu, line 94: i didn't see an example of this in the cookbook, did I miss it?

src/databricks_ai_bridge/long_running/resume.py
```diff
@@ -0,0 +1,77 @@
...
+    async def default_resume_request(self, request: ResponsesAgentRequest) -> ResponsesAgentRequest:
+        """Apply the configured strategy's built-in request transformation."""
+        return await self._build_default_resume_request(request)
+
+
+def get_on_resume_function() -> Callable[..., Any] | None:
+    return _on_resume_function
+
+
+def on_resume() -> Callable[[Callable[_P, _R]], Callable[_P, _R]]:
```
jamesbxwu, line 58: it's not immediately clear that on_resume() is used to reconstruct an incoming request during crash recovery.  
1. if that is indeed the hook we are exposing, then we should rename this to `build_recovery_request`
2. right now the decorator here does not restrict the input and output, should we do that?

cookbooks/openai-sdk-agent/README.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+The default transformation depends on `resume_strategy`:
+
+- `ResumeStrategy.EVENT_LOG` appends the immediately previous attempt's
+  durable events as recovery prose and changes the session key to
+  `<original>::attempt-N`.
+- `ResumeStrategy.AGENT_SESSION` keeps the original session key and replaces
+  the request input with a recovery prompt. The agent SDK supplies the history.
```
jamesbxwu, line 103: how should the developer reason about using `AGENT_SESSION` and implementing `on_resume`? My understanding is that setting `ResumeStrategy.AGENT_SESSION` means I get automatic recovery if my harness/sdk has a session store so when should I implement `on_resume()`?

jamesbxwu (commented):
had a few more questions on the combination of using `ResumeStrategy.AGENT_SESSION` and `on_resume()` together, let's chat offline

shivam5:
Interface comparison set:

shivam5:
Keeping long running agent server as is. New durable runtime sdk merged in https://github.com/databricks/databricks-ai-bridge/pull/476

PR #471, opened 2026-08-24
Enable databricks-mason TestPyPI development releases

smurching:
## Summary
- set the initial databricks-mason scaffold version to 0.1.0.dev0
- enable only the generated databricks-mason release workflow
- retain TestPyPI as the default destination for tag-triggered releases

## Verification
- regenerated all release workflows and confirmed only the Mason workflow changed
- verified release metadata, OIDC permission, and TestPyPI environment selection

This development release validates packaging and trusted publishing. The first functional production release remains 0.1.0. Production publishing also requires a separately configured PyPI trusted publisher and an explicitly approved production workflow dispatch.

### DIFF
--- a/.github/workflows/generate_release_workflows.py
+++ b/.github/workflows/generate_release_workflows.py
@@ -8,16 +8,16 @@ from pathlib import Path
 @dataclass
 class Package:
     name: str
+    release_enabled: bool
     working_dir: str | None = None  # None means root level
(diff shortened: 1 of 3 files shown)


smurching:
Closing because package publishing has moved to databricks/secure-public-registry-releases-eng. The source package remains at 0.0.1.dev0; release enablement will be handled in the secure release repository.

PR #472, opened 2026-08-24
Add the Mason CLI

smurching:
## Summary
- add the `mason` console entry point and authenticated `agents/v1` API client
- add commands for memory stores/entries, session stores/sessions/items, tracing, and Databricks Apps deployments
- add saved-profile login/logout, JSON output, Rich terminal rendering, and optional MLflow tracing support
- add unit coverage for authentication, command layout, REST paths, rendering, deployment wiring, and tracing
- mark the functional preview as `0.1.0.dev0`

This is based on Ankit Mathur’s experimental Mason CLI prototype, adapted for this OSS package.

## OSS scope
Only generic client/CLI functionality is included. Internal application/mock code, internal documentation, staging profiles and observations, service implementation paths, corporate URLs, and rollout details are excluded.

## Verification
- Mason unit tests pass with lowest-direct and highest dependency resolution
- Mason Ruff lint, formatting, and type checks pass
- package metadata, console entry point, AST parsing, byte compilation, and secret scanning pass

PR #474, opened 2026-08-25
Build out the Mason CLI: init, dev, and store-aware deploy

elainemwang:
Builds out the Mason CLI beyond the initial skeleton (#472) into the full local-to-deployed agent workflow: scaffold a template, run it locally, and deploy it with its stores wired up.

## `mason init`
Sparse-clones a basic template (`--framework openai|langgraph`) into a local directory, ready to run. `--repo`/`--ref` override the source so a template can be fetched from a fork/branch before it merges. `--profile` seeds a local `.env` (`DATABRICKS_CONFIG_PROFILE`) so the scaffolded project runs immediately; the profile/`.env` step is surfaced in Next steps when no profile is set.

## `mason dev`
Runs a scaffolded agent locally by wrapping `databricks apps run-local` — same `app.yaml` command + env as a deployment, so local behavior matches what ships. Builds the venv on first run and reuses it after (`--prepare-environment` forces a rebuild). Strips the deploy-only package-index env vars for the local build so it uses the machine's own index.

## `mason deploy` — store access
When a deployment uses a managed session or memory store, the app runs as its service principal while the store's Lakebase tables are owned by the deploying user, and the SP has no access by default. `deploy` now grants it automatically (best-effort, two steps):
1. binds each store's per-store database as a `postgres` app resource (Lakebase role + CONNECT, past the project ACL), then
2. issues an owner-scoped `psql` GRANT on the store's tables (`public.sessions`/`session_items` for sessions; `memory.memory_entries` for memory).

Both stores' resources go in a single `apps update` (it replaces the whole resource array). The store-agnostic mechanics live in `store_access.py` (`LakebaseBackend` + `apply_postgres_resources` + `grant_tables`); `session_store_access` / `memory_store_access` supply per-store specifics. `AGENT_MEMORY_STORE` carries the bare store id (the runtime re-adds the `memory-stores/` prefix).

## Templates
`--framework langgraph` points at the merged `integrations/mason/templates/agent-langgraph` template.

Tested: unit suite green; store grants verified end-to-end against a deployed app (remember/recall round-trips through the memory store, durable sessions across restart).

### DIFF
--- a/integrations/mason/src/databricks_mason/init.py
+++ b/integrations/mason/src/databricks_mason/init.py
@@ -1,11 +1,12 @@
-"""`mason init` — scaffold a local agent project from an app-templates template.
+"""`mason init` — scaffold a local agent project from a mason template.
 
-Fetches one template directory out of the `databricks/app-templates` repo (a sparse,
-blobless clone so only the chosen template is materialized) and drops it into a local
-target directory. That directory is then ready for `mason deploy --source <dir>`.
+Fetches one template directory out of its git repo (a sparse, blobless clone so only the
+chosen template is materialized) and drops it into a local target directory, ready for
+`mason deploy --source <dir>`.
 
-`--framework` selects which basic template to lay down; the repo and ref are overridable
-so the command can target a fork/branch before a template has merged to the canonical repo.
+`--framework` selects which template to lay down; each framework knows its own repo, ref, and
+path (see `_TEMPLATES`). `--repo` / `--ref` override those, e.g. to pull from a fork or branch
+before a template has merged to its canonical repo.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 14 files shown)


junchoi-db (approved):

integrations/mason/src/databricks_mason/init.py
```diff
@@ -0,0 +1,153 @@
...
+import subprocess
+import tempfile
+from typing import Optional
+
+import click
+
+from databricks_mason import render
+from databricks_mason.errors import AgentCliError
+
+_DEFAULT_REPO = "https://github.com/databricks/app-templates.git"
```
shivam5: should this be now https://github.com/databricks/databricks-ai-bridge?

integrations/mason/src/databricks_mason/init.py
```diff
@@ -0,0 +1,153 @@
...
+    if wrote_env:
+        fields["Profile (.env)"] = env_profile
+    else:
+        # No profile resolved, so no .env was seeded — call out the auth step explicitly rather
+        # than burying it, since `uv run start-server` fails without a Databricks profile.
+        steps += [
+            "cp .env.example .env",
+            "Set DATABRICKS_CONFIG_PROFILE in .env (or re-run `mason init --profile <profile>`)",
+        ]
+    steps += ["uv run start-server        # run locally", f"mason deploy <name> --source {dest}"]
```
shivam5: can we add a TODO here to change this step to `mason dev`?

integrations/mason/src/databricks_mason/init.py
```diff
@@ -0,0 +1,167 @@
...
+    default="openai",
+    show_default=True,
+    help="Which basic agent template to scaffold.",
+)
+@click.option(
+    "--profile",
+    default=None,
+    help="Seed a local .env with this DATABRICKS_CONFIG_PROFILE so `uv run start-server` works "
+    "immediately (defaults to the profile from -p / `mason login`).",
+)
```
shivam5, line 119: I think --profile should not be part of init if it is actually not needed. - assuming no databricks dependency for init.

Ideally just a part of `deploy`, but okay to be part of `dev` also for now.

shivam5 (approved):

integrations/mason/src/databricks_mason/init.py
elainewang-db: agreed, we should be able to set up other model provider auth

PR #475, opened 2026-08-26
Prototype transport-neutral durable runtime library

shivam5:

This is a **draft interface prototype** for the transport-neutral library option. It is intentionally an alternative to the provided app/server interfaces, not a proposed final API.

### Developer contract

```python
runtime = DatabricksDurableRuntime(run_agent)
await runtime.start()
await runtime.submit(run_id, request)
```

The library stores requests, results, heartbeats, attempts, and ordered events. The developer owns FastAPI lifecycle, request mapping, `202` responses, polling, SSE formatting, cursor handling, and recovery behavior inside `run_agent`.

- A background streamed proposal completes with `requires_action`.
- The client polls the persisted result and submits approval as a second run with the same session ID.
- OpenAI Agents SDK conversation state is stored in `AsyncDatabricksSession`.
- SDK events are persisted through `context.emit()` and exposed by developer-authored SSE routes.
- The approved run includes a 60-second tool call so the process can be stopped and recovered with `context.is_recovery=True`.

| Capability | Result |
| --- | --- |
| Existing request/response protocol | Developer can preserve it |
| Background execution | Runtime primitive provided; HTTP contract developer-owned |
| Pod/process recovery | Supported |
| Durable stream replay | Storage provided; SSE route developer-owned |
| Final result | Stored by the runtime; polling route developer-owned |

This prototype makes the cost of the library approach explicit: it has the least protocol coupling and the most developer integration work.

## How do you know it works?

- `23` focused runtime/store tests pass.
- Ruff check and format pass for the cookbook files.

| --- | --- |
| Background | The runtime stores and schedules work, but the developer defines `202`, run IDs, polling routes, and result mapping. |
| Streaming | `context.emit()` and `runtime.events()` provide durable storage; the developer defines SSE, reconnect cursors, and client behavior. |
| OpenAI Agents SDK client | The developer may preserve an existing route or adopt the cookbook's `/runs` envelope. The library does not mandate the migration. |
| LangGraph SDK client | `langgraph_sdk` can remain unchanged only if the developer implements its native routes over the runtime; the cookbook instead uses `/runs`. |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Improvements
 - databricks-openai: `DatabricksOpenAI` and `AsyncDatabricksOpenAI` clients now follow HTTP redirects by default, configurable via the new `follow_redirects` parameter (#445)
+- databricks-ai-bridge: Add a transport-neutral `DatabricksDurableRuntime` with Lakebase request, response, event persistence, and stale-attempt recovery
 
 ### Bug Fixes
 - databricks-ai-bridge: Genie now returns the full answer text aggregated from all text attachments (#432)
--- a/README.md
+++ b/README.md
@@ -24,6 +24,20 @@ For frameworks without dedicated integration packages:
 pip install databricks-ai-bridge
 ```
 
+## Durable Runtime
+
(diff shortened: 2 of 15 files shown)


shivam5:
Interface comparison set:

shivam5:
Just for prototype

PR #476, opened 2026-08-26
Prototype SDK-provided durable entrypoint

shivam5:
## What changed

This PR adds a framework-neutral durable runtime to `databricks-mason` and one `durable-langgraph-agent` template with a real Databricks chat model, LangGraph tools, and persisted streaming events.

```python
from databricks_mason import DurableAgentApp, DurableAgentContext

app = DurableAgentApp()


@app.invoke
@app.on_recovery
async def invoke(input: object, context: DurableAgentContext) -> object:
    return await run_agent(input, context)
```

The SDK implementation stays in Mason's existing runtime package:

```text
integrations/mason/src/databricks_mason/runtime/durability/
  app.py
  attempt.py
  recovery.py
  runtime.py
  store.py
  types.py
```

- `runtime.py`: public execution facade and polling.
- `attempt.py`: attempt claiming, execution, heartbeats, fencing, and completion.
- `recovery.py`: local task scheduling and stale-attempt recovery scanning.
- `store.py`: in-memory development store and Lakebase production store.
- `app.py`: thin FastAPI adapter and instance-scoped decorators.
- `types.py`: callback, context, execution, event, and store contracts.

`DurableAgentApp` subclasses FastAPI directly. Public imports are available from `databricks_mason` and `databricks_mason.runtime`.

## HTTP contract

The app exposes only:

- `POST /api/invocations`
- `GET /api/invocations/{invocation_id}`
- `GET /api/invocations/{invocation_id}/events?after={cursor}`

The client supplies a UUID `id`, which is the idempotency key in every mode. Retrying the same ID and input returns the existing run; reusing the ID with different input returns `409`.

| Mode | Request flags | Response |
|---|---|---|
| Foreground sync | default | `200` with `output` |
| Background sync | `background: true` | `202` with status URL |
| Foreground stream | `stream: true` | `200` SSE |
| Background stream | both flags | `202` with status/event URLs |

Inputs and outputs may be any JSON value, including `null`. Transport fields are not passed to the callback. There is no request-level `resume`, protocol-level `attempt`, health endpoint, duplicate non-`/api` route, or nested server object.

The intended session/routing source is the `__Host-databricks-app-router` cookie. The Apps proxy currently consumes that cookie before forwarding deployed requests, so the adapter deterministically falls back to the invocation ID as the session ID. The code has a TODO to adopt a standard session header when Databricks Apps supports one. Polling is keyed by invocation ID and relies on Databricks Apps authentication rather than treating the session ID as authorization.

The store persists `run.started`, `run.completed`, and `run.failed` lifecycle events. Terminal events and terminal state changes are committed atomically. The heartbeat task is always cancelled and awaited in `finally`, including when execution fails.

## Recovery

`@app.invoke` handles the first attempt. `@app.on_recovery` handles a stale attempt claimed after worker loss. Omitting `@app.on_recovery` logs a startup warning and disables crash-recovery scanning; applications may register the same function for both decorators when replaying the initial invocation is safe.

The internal attempt remains for ownership fencing, but it is not exposed as protocol metadata. Agent checkpoint restoration and idempotent external side effects remain application responsibilities.

## Lakebase durability

A deployed durability app uses exactly one Lakebase database:

1. Reuse the configured Session Store database.
2. Otherwise reuse or provision the dedicated `<app>-durability` project.

The Memory Store database is not used as a durability fallback. Each app owns a schema named `databricks_mason_runtime_<12-char-sha256(app-name)>` and adds only its `executions` and `execution_events` tables.

Mason injects the selected endpoint and schema through:

```text
DATABRICKS_MASON_RUNTIME_ENDPOINT
DATABRICKS_MASON_RUNTIME_SCHEMA
```

Local execution always uses the in-memory store. A Databricks App using `DurableAgentApp` fails startup if no durability endpoint was attached.

## Template and CLI scope

`mason init --framework langgraph --durability` scaffolds the new API-only `durable-langgraph-agent` and records `[durability] enabled = true` in `agent.toml`. There is no public `mason durability enable/bind` command.

Explicit `--framework langgraph` and `--framework openai` keep the existing templates and chat UI overlays unchanged. Migrating those templates is intentionally deferred to a separate PR.

## Current validation

Validated commit `7a80262d52ebcd1478b714a206e4bbc152d1ecc4` on the latest `origin/main`:

<details>
<summary>Show local validation commands and results</summary>

```bash
cd integrations/mason
uv run --exact --group tests pytest tests/unit_tests
uv run --exact --group tests pytest tests/unit_tests/test_durable_app.py
PYTHONPATH=templates/durable-langgraph-agent uv run --extra runtime pytest templates/durable-langgraph-agent/tests
uv run --exact --all-extras ruff check
uv run --exact --all-extras ruff format --check --diff
uv run --exact --all-extras ty check
uv lock --check
uv build --wheel --out-dir <temporary-directory>
git diff --check
```

Results:

- Mason unit tests: `390 passed`
- Focused runtime application tests: `24 passed`
- Standalone durability template tests: `4 passed`
- Ruff, format, type check, lock check, and `git diff --check`: passed
- Mason wheel: built successfully
- Wheel contents include `lakebase_durability_store.py` and `runtime/durability/{app,attempt,recovery,runtime,store,types}.py`
- Existing LangGraph, OpenAI, and UI template directories have no effective diff
- Fresh GitHub override scaffold pinned `databricks-mason` to commit `7a80262` through `tool.uv.sources`; `uv sync`, public SDK imports, `mason dev`, a real model/tool call, background polling, and SSE replay passed

</details>

## Live end-to-end validation (`e2-dogfood`, 2026-09-08)

- All four invocation modes passed against the deployed real LangGraph agent.
- Idempotent retry reused the original execution without creating another attempt.
- A running invocation survived app stop/start and completed as recovery attempt 2.
- Lakebase retained lifecycle and model events from both attempts.

<details>
<summary>Show scaffold, deployment, and authentication commands</summary>

The exact current commit was scaffolded and deployed with one app instance:

```bash
cd integrations/mason
uv run mason --profile e2-dogfood init \
  --framework langgraph \
  --durability \
  --profile e2-dogfood \
  --repo https://github.com/shivam5/databricks-ai-bridge.git \
  --ref poc/durable-entrypoint-interface \
  /tmp/pr476-live-agent-7a80262

cd /tmp/pr476-live-agent-7a80262
uv sync --refresh
uv run mason --profile e2-dogfood deploy durable-langgraph-agent \
  --source /tmp/pr476-live-agent-7a80262 \
  --instances 1
```

`mason init` resolves that branch and pins the generated project's `databricks-mason` source to exact commit `7a80262d52ebcd1478b714a206e4bbc152d1ecc4`.

Deployment result:

```text
App:      mason-durable-langgraph-agent
URL:      https://mason-durable-langgraph-agent-6051921418418893.staging.aws.databricksapps.com
Postgres: projects/mason-durable-langgraph-agent-durability/branches/production/databases/databricks-postgres
```

Requests and app lifecycle operations used the Apps workspace profile `e2-dogfood-apps`:

```bash
export PROFILE=e2-dogfood-apps
export APP=mason-durable-langgraph-agent
export APP_URL=https://mason-durable-langgraph-agent-6051921418418893.staging.aws.databricksapps.com
export TOKEN=$(databricks auth token --profile "$PROFILE" -o json | jq -r .access_token)
export ROUTING_KEY=$(uuidgen)
```

The similarly named singular profile `e2-dogfood-app` targets another workspace and returned `401`; `e2-dogfood-apps` is the profile for this deployed app.

</details>

<details>
<summary>Show foreground, idempotency, streaming, and background commands</summary>

### Foreground, idempotency, streaming, and background

Foreground real model/tool request:

```bash
export FG_ID=b7353b92-8a94-42f9-86d2-929c2eccf52d
curl -sS "$APP_URL/api/invocations" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Cookie: __Host-databricks-app-router=$ROUTING_KEY" \
  -d "{\"id\":\"$FG_ID\",\"input\":[{\"role\":\"user\",\"content\":\"What time is it? Use the get_current_time tool.\"}]}"
```

Result: HTTP `200`, `status=completed`, `recovered=false`; output contained AI tool-call, tool-result, and final AI messages. Repeating the identical request returned HTTP `200` with the same persisted output and did not create another attempt.

Foreground SSE:

```bash
export STREAM_ID=ef76ec01-2dae-4c5e-806b-d7e1ea85f350
curl -N -sS "$APP_URL/api/invocations" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Cookie: __Host-databricks-app-router=$ROUTING_KEY" \
  -d "{\"id\":\"$STREAM_ID\",\"stream\":true,\"input\":[{\"role\":\"user\",\"content\":\"Reply exactly: deployed streaming works\"}]}"
```

Result: HTTP `200` SSE persisted and returned `run.started`, token `delta` events, the complete `message`, and `run.completed`.

Background polling:

```bash
export BG_ID=83aff9e4-f099-406d-8e7d-986482d87897
curl -sS "$APP_URL/api/invocations" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Cookie: __Host-databricks-app-router=$ROUTING_KEY" \
  -d "{\"id\":\"$BG_ID\",\"background\":true,\"input\":[{\"role\":\"user\",\"content\":\"Reply exactly: background complete\"}]}"

curl -sS "$APP_URL/api/invocations/$BG_ID" \
  -H "Authorization: Bearer $TOKEN"
```

Result: submit returned HTTP `202` with only `status_url`; polling observed `queued -> active -> completed` and the final result had `recovered=false`.

</details>

<details>
<summary>Show pod-loss recovery commands and Lakebase evidence</summary>

### Pod-loss recovery and Lakebase evidence

A background stream was kept active inside `wait_for_seconds(90)`:

```bash
export CRASH_ID=298764ae-4261-40b2-b44d-09e94ba3b021
curl -sS "$APP_URL/api/invocations" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Cookie: __Host-databricks-app-router=$ROUTING_KEY" \
  -d "{\"id\":\"$CRASH_ID\",\"background\":true,\"stream\":true,\"input\":[{\"role\":\"user\",\"content\":\"You must call wait_for_seconds with exactly 90 seconds. After the tool returns, say recovery test complete.\"}]}"

curl -N --max-time 5 -sS "$APP_URL/api/invocations/$CRASH_ID/events" \
  -H "Authorization: Bearer $TOKEN"
```

The submit returned HTTP `202` with both `status_url` and `events_url`. Before the crash, SSE and Lakebase contained attempt 1's `run.started` plus the model's `wait_for_seconds(seconds=90)` tool call, and the execution row was:

```text
status=ACTIVE, attempt=1, response=NULL
```

The only app instance was then stopped and restarted:

```bash
databricks apps stop "$APP" --profile "$PROFILE" -o json

databricks psql \
  projects/mason-durable-langgraph-agent-durability/branches/production/endpoints/primary \
  --profile "$PROFILE" -- -d databricks_postgres -X -P pager=off -c \
  "SELECT execution_id,status,attempt,heartbeat_at,response
   FROM databricks_mason_runtime_19e6e525be98.executions
   WHERE execution_id='$CRASH_ID';"

databricks apps start "$APP" --profile "$PROFILE" -o json

curl -sS "$APP_URL/api/invocations/$CRASH_ID" \
  -H "Authorization: Bearer $TOKEN"
```

Observed state transitions:

```text
After stop:    compute=STOPPED, execution=ACTIVE, attempt=1, response=NULL
After restart: compute=ACTIVE,  execution=COMPLETED, attempt=2, recovered=true
```

Lakebase event replay retained both attempts in order:

```text
attempt 1: run.started, message(wait_for_seconds tool call)
attempt 2: run.started, delta..., message, run.completed
```

The final status endpoint returned HTTP `200`, `status=completed`, `output.recovered=true`, and the original invocation/session ID. A terminal `GET /events` replay returned all 11 events across both attempts and closed normally. App logs showed the new process reaching application startup successfully with no recovery exception.

Lakebase used exactly the app's one attached database and app-owned schema:

```text
Endpoint: projects/mason-durable-langgraph-agent-durability/branches/production/endpoints/primary
Database: databricks_postgres
Schema:   databricks_mason_runtime_19e6e525be98
Tables:   executions, execution_events
```

</details>


shivam5:
Interface comparison set:

integrations/mason/src/databricks_mason/durable_server/app.py
```diff
@@ -0,0 +1,342 @@
...
+AgentHook = Callable[[JsonObject, "DurableAgentContext"], Awaitable[JsonObject]]
+
+
+@dataclass(frozen=True)
+class DurableAgentContext:
+    """Invocation metadata and durable event emission for an agent callback."""
+
+    run_id: str
+    session_id: str
+    actor: str
```
shivam5: what is actor here?

integrations/mason/src/databricks_mason/durable_server/app.py
```diff
@@ -0,0 +1,342 @@
...
+    DurableRequestConflictError,
+    JsonObject,
+)
+
+RUN_ID_HEADER = "X-Databricks-Run-Id"
+SESSION_ID_HEADER = "X-Databricks-Session-Id"
+
+_ROUTING_COOKIE = "__Host-databricks-app-router"
+_LOCAL_SESSION_COOKIE = "mason-local-session"
+_USER_HEADERS = ("X-Forwarded-Email", "X-Forwarded-User")
```
shivam5: we just need __Host-databricks-app-router

integrations/mason/src/databricks_mason/durable_server/app.py
```diff
@@ -0,0 +1,342 @@
...
+
+@dataclass(frozen=True)
+class DurableAgentContext:
+    """Invocation metadata and durable event emission for an agent callback."""
+
+    run_id: str
+    session_id: str
+    actor: str
+    attempt: int
+    _execution_context: DurableExecutionContext
```
shivam5: why is this nested?

integrations/mason/src/databricks_mason/durable_server/app.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+class DurableAgentApp:
+    """Expose an agent callback through Mason's durable HTTP protocol."""
+
+    def __init__(
+        self,
+        invoke: AgentHook,
+        *,
+        on_resume: AgentHook | None = None,
```
shivam5: does this being optional mean we will not do auto-recovery if this hook is not provided?
This is a valid behaviour, but we should document this

integrations/mason/src/databricks_mason/durable_server/app.py
```diff
@@ -0,0 +1,342 @@
...
+        durability_store: DurabilityStore,
+        heartbeat_seconds: float = 3.0,
+        stale_seconds: float = 10.0,
+        scan_seconds: float = 3.0,
```
shivam5: does long running agent server also have scan_seconds and poll_seconds?

integrations/mason/src/databricks_mason/durable_server/app.py
```diff
@@ -0,0 +1,342 @@
...
+            self.app.add_api_route(
+                f"{prefix}/invocations/{{run_id}}", self._get_invocation, methods=["GET"]
+            )
+            self.app.add_api_route(
+                f"{prefix}/invocations/{{run_id}}/events",
+                self._get_events,
+                methods=["GET"],
+            )
+            self.app.add_api_route(f"{prefix}/health", self._health, methods=["GET"])
+        self.app.add_api_route("/api/session/new", self._new_session, methods=["POST"])
```
shivam5: what does new session do? should this be part of runtime? shouldn't this be developer controlled?

integrations/mason/src/databricks_mason/durable_server/app.py
```diff
@@ -0,0 +1,342 @@
...
+        self.app.add_api_route("/api/healthz", self._health, methods=["GET"])
```
shivam5: what are the 2 health endpoints? why do we need 2, and what does this tell?

integrations/mason/src/databricks_mason/durable_server/app.py
```diff
@@ -0,0 +1,342 @@
...
+            raise RuntimeError("persisted runtime request is invalid")
+
+        context = DurableAgentContext(
+            run_id=execution_context.execution_id,
+            session_id=session_id,
+            actor=actor,
+            attempt=execution_context.attempt,
+            _execution_context=execution_context,
+        )
+        hook = self._on_resume if context.is_recovery and self._on_resume else self._invoke
```
shivam5: if on_resume not provded then we should not recover

integrations/mason/src/databricks_mason/runtime/__init__.py
```diff
@@ -36,10 +24,23 @@
...
-    return getattr(importlib.import_module(f"{__name__}.{module}"), name)
+        return getattr(durable_server, name)
+    if module_name := _MODULE_BY_NAME.get(name):
+        from importlib import import_module
+
+        return getattr(import_module(f"{__name__}.{module_name}"), name)
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+
+__all__ = [
```
shivam5: why unecessary code movement in this file?

integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -14,19 +14,33 @@
...
+    LakebaseBackend,
+    _databricks,
+    apply_postgres_resources,
+    grant_tables,
+)
 from databricks_mason.tracing import TRACES_DEST_ENV, TRACES_EXPERIMENT_ENV, default_experiment
 
 _MEMORY_ENV = "AGENT_MEMORY_STORE"
 _SESSION_ENV = "AGENT_SESSION_STORE"
+_RUNTIME_ENDPOINT_ENV = "DATABRICKS_MASON_RUNTIME_ENDPOINT"
```
shivam5: call this agent_durability_store?

integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -282,38 +296,41 @@ def resolve_store_env(
     return env
 
 
-def _grant_store_access(
-    app: str,
+def _grant_store_tables(
```
shivam5: uneeded refactoring, just add durability store to be used if session store is there, otherwise provision it by a default name

integrations/mason/src/databricks_mason/deploy.py
shivam5: Renamed the provisioning helper to `agent_durability_store.py` and its tests accordingly in `9b9e3c9`.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/deploy.py
shivam5: Simplified in `9b9e3c9`: deployment now reuses the Session Store Lakebase backend when configured; otherwise it ensures the default `<app>-durability` backend. The broader metadata/replacement refactor was removed and the existing managed-store grant flow was restored.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/durable_server/app.py
shivam5: Addressed in `9b9e3c9`: the server is now the thin `runtime/app.py` layer, `DurableAgentContext` has no actor, the durable app uses only `__Host-databricks-app-router`, the extra headers/cookie are removed, recovery scanning starts only when `on_resume` exists, timing/schema knobs are no longer exposed by `DurableAgentApp`, session creation is owned by the optional UI overlay, and only `/api/healthz` remains.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/runtime/__init__.py
shivam5: Restored the existing lazy-import structure in `9b9e3c9`; this file now only adds the two new public names to `__all__`, `TYPE_CHECKING`, and `_MODULE_BY_NAME`.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/durable_server/app.py
shivam5: The remaining server feedback is also addressed in `9b9e3c9`: only the Apps routing cookie remains, the app moved to `runtime/app.py`, recovery is disabled without `on_resume`, constructor timing knobs and runtime-owned session creation were removed, and the app exposes only `/api/healthz`.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/durable_server/app.py
shivam5: Moved the application into `databricks_mason.runtime.app`; there is no nested `durable_server` package in `9b9e3c9`.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/durable_server/app.py
shivam5: Correct: auto-recovery now runs only when `on_resume` is provided, and that behavior is documented in `9b9e3c9`.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/durable_server/app.py
shivam5: Removed `scan_seconds`, `poll_seconds`, heartbeat, stale, and schema configuration from the public `DurableAgentApp` constructor in `9b9e3c9`; those remain internal runtime defaults.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/durable_server/app.py
shivam5: Removed session creation from the SDK runtime. The optional chat UI overlay owns `/api/session/new`, so non-UI developers control sessions themselves.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/durable_server/app.py
shivam5: Removed the duplicate health route. The SDK app now exposes only `/api/healthz`.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/durable_server/app.py
shivam5: Implemented in `9b9e3c9`: without `on_resume`, startup does not scan or claim recoverable executions.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/templates/agent-langgraph/tests/test_runtime.py
```diff
@@ -1,22 +1,20 @@
...
-    assert paths["/invocations/{invocation_id}"]["get"]
-    assert paths["/api/invocations/{invocation_id}"]["get"]
-    assert paths["/health"]["get"]
-    assert paths["/api/health"]["get"]
+    assert paths["/invocations/{run_id}"]["get"]
+    assert paths["/api/invocations/{run_id}"]["get"]
+    assert paths["/invocations/{run_id}/events"]["get"]
+    assert paths["/api/invocations/{run_id}/events"]["get"]
+    assert paths["/api/healthz"]["get"]
+    assert "/api/session/new" not in paths
```
shivam5: remove this

elainewang-db (commented):
Can we make a new template with using our new durable runtime for now? I think we should keep the old templates with the purpose of illustrating how people can think about writing their existing agents with our `databricks_mason` sdk. we can point `mason init` to pull in the new template by default, wdyt? 

In a follow up, we really should spend some time organizing the templates and pulling out the common and sdk agnostic pieces, so that the template folders themselves only have the minimum of what they need and the common pieces are all shared and copied in by `mason init`.

integrations/mason/templates/agent-langgraph/tests/test_runtime.py
shivam5: Removed `/api/healthz` and the corresponding template assertion in `17dcb23`. The durable app now exposes only the invocation, polling, and event routes.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

shivam5:
Addressed Elaine’s template-scope feedback: this PR now leaves all existing LangGraph/OpenAI/chat templates unchanged and adds only the minimal `durability-app` default. I moved the existing-template migration and cleanup into draft follow-up #550.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/templates/durability-app/agent/agent.py
```diff
@@ -0,0 +1,56 @@
...
+builder.add_edge("process", END)
+graph = builder.compile()
+
+
+async def run_agent(payload: dict, context: DurableAgentContext) -> dict:
+    agent_input = payload.get("input", {})
+    if not isinstance(agent_input, dict):
+        raise ValueError("input must be an object")
+
+    message = agent_input.get("message", "hello")
```
elainewang-db: qq: why is there a default message here?

integrations/mason/templates/durability-app/agent/agent.py
```diff
@@ -0,0 +1,56 @@
...
+    delay_seconds = agent_input.get("delay_seconds", 0)
```
elainewang-db: what is `delay_seconds` for, why do we need it?

integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -314,6 +317,13 @@ def _grant_store_access(
     return None
 
 
+def _uses_durable_runtime(source_dir: pathlib.Path) -> bool:
+    """Whether this project was scaffolded from Mason's durability template."""
```
elainewang-db: hmm instead of coupling the durability setup to the template, can we introduce a new `mason durability bind` or something like `mason sessions/memory bind`?

integrations/mason/templates/durability-app/agent/agent.py
shivam5: Removed the implicit `"hello"` fallback in `313a3e7`. The template now requires callers to provide `input.message`, and the standalone test suite covers the missing-message validation.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -0,0 +1,193 @@
...
+        *,
+        on_resume: AgentHook | None = None,
+        durability_store: DurabilityStore | None = None,
+    ) -> None:
+        self._invoke = invoke
+        self._on_resume = on_resume
+        self._runtime = DurableRuntime(
+            self._execute,
+            durability_store=(
+                durability_store if durability_store is not None else default_durability_store()
```
junchoi-db, line 61: qq: is default lakebox?

integrations/mason/templates/durability-app/agent/agent.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+builder = StateGraph(AgentState)
+builder.add_node("process", _process)
+builder.add_edge(START, "process")
+builder.add_edge("process", END)
+graph = builder.compile()
+
+
+async def run_agent(payload: dict, context: DurableAgentContext) -> dict:
```
junchoi-db, line 26: qq: is this util function to be used by templates/ clients?

integrations/mason/templates/durability-app/agent/agent.py
shivam5: Agreed — `delay_seconds` was only a demo aid for making an in-flight request easy to observe, and it does not belong in the minimal template. It has been removed; the template now runs the graph directly from `input.message`.

integrations/mason/src/databricks_mason/deploy.py
shivam5: Yes. This is now explicit via `mason durability bind` / `mason durability unbind`, stored as `[durability] enabled = true` in `agent.toml`. Deploy checks that binding rather than template identity. Bare `mason init` binds the new default durability template; existing projects opt in explicitly.

integrations/mason/src/databricks_mason/runtime/app.py
shivam5: The deployed default is Lakebase; the local default is in-memory. `mason deploy` reuses the Session Store Lakebase database first, then the Memory Store database, or provisions a dedicated fallback, and sets `DATABRICKS_MASON_RUNTIME_ENDPOINT`. `default_durability_store()` then selects `LakebaseDurabilityStore`; without that environment variable, local development uses `InMemoryDurabilityStore`. Callers can also pass `durability_store=` explicitly.

integrations/mason/templates/durability-app/agent/agent.py
shivam5: `run_agent` is template-owned sample application code, not an SDK utility. It demonstrates the callback contract accepted by `DurableAgentApp`; users replace its graph/body with their own agent logic. `runtime/main.py` simply passes it as both the initial invocation and recovery callback.

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,321 @@
...
+    async def _execute_attempt(self, execution_id: str) -> None:
+        try:
+            claimed = await self.durability_store.claim(execution_id, self.stale_seconds)
+        except Exception:
+            logger.exception("Failed to claim durable execution: %s", execution_id)
+            return
+        if claimed is None:
+            return
+
+        heartbeat = asyncio.create_task(
```
junchoi-db, line 222: qq: when execution fails -- do we also cleanup the heartbeat task?

junchoi-db (approved):

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: Yes. `_execute_attempt` creates the heartbeat after claiming the attempt and always cancels and awaits it in `finally`, so success, execution failure, ownership loss, and task cancellation all clean it up. This is covered by `test_failed_execution_stops_heartbeat`, and the current implementation is in commit `0024933`.

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+
+        @asynccontextmanager
+        async def lifespan(_: FastAPI):
+            if self._invoke_hook is None:
+                raise RuntimeError("register an invocation handler with @app.invoke")
+            recover = self._on_recovery_hook is not None
+            if not recover:
+                logger.warning(
+                    "No @app.on_recovery handler is registered; crash recovery is disabled."
```
jamesbxwu:
```suggestion
                    "No @app.on_recovery handler is registered; automatic crash recovery is disabled."
```

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -0,0 +1,219 @@
...
+                )
+            await self._runtime.start(recover=recover)
+            try:
+                yield
+            finally:
+                await self._runtime.stop()
+
+        super().__init__(
+            title="Databricks Durable Agent",
```
jamesbxwu:
```suggestion
            title="Databricks Durable Agent Runtime",
```

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -0,0 +1,219 @@
...
+            lifespan=lifespan,
+            docs_url=None,
+            redoc_url=None,
+            openapi_url=None,
+        )
+        self.middleware("http")(self._bind_session)
+        self.add_api_route(_API_ROOT, self._invoke_request, methods=["POST"])
+        self.add_api_route(f"{_API_ROOT}/{{run_id}}", self._get_request, methods=["GET"])
+        self.add_api_route(f"{_API_ROOT}/{{run_id}}/events", self._events, methods=["GET"])
```
jamesbxwu:
```suggestion
        self.add_api_route(f"{_API_ROOT}/{{invocation_id}}", self._get_request, methods=["GET"])
        self.add_api_route(f"{_API_ROOT}/{{invocation_id}}/events", self._events, methods=["GET"])
```

integrations/mason/src/databricks_mason/runtime/types.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+DurableExecutor = Callable[[JsonValue, DurableExecutionContext], Awaitable[JsonValue]]
+
+
+@dataclass(frozen=True)
+class DurableAgentContext:
+    """Invocation metadata and durable event emission for an agent callback."""
+
+    run_id: str
```
jamesbxwu:
```suggestion
    invocation_id: str
```

integrations/mason/src/databricks_mason/runtime/types.py
```diff
@@ -0,0 +1,147 @@
+"""Public types for durable request execution."""
```
jamesbxwu: can we document the relationship between all the entities defined here?

integrations/mason/src/databricks_mason/runtime/durability/types.py
```diff
@@ -0,0 +1,147 @@
...
+        return self.attempt > 1
+
+    async def emit(self, event: JsonObject) -> int:
+        return await self._execution_context.emit(event)
+
+
+AgentHook = Callable[[JsonValue, DurableAgentContext], Awaitable[JsonValue]]
+
+
+class DurabilityStore(Protocol):
```
jamesbxwu, line 110: add docs for all methods below

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,324 @@
...
+from __future__ import annotations
+
+import asyncio
+import copy
+import json
+import logging
+from datetime import datetime, timezone
+from typing import cast
+
+from databricks_mason.runtime.types import (
```
jamesbxwu: can we introduce a `databricks_mason.runtime.durability` package/source folder to achieve a clean separation of a generic runtime from the durability specific extension?

integrations/mason/src/databricks_mason/runtime/durability/runtime.py
jamesbxwu, line 1: Similar to https://github.com/databricks/databricks-ai-bridge/pull/476/changes#r3954520336, can we layer the SDK here such that we have the generic app/runtime as the baseline with the durable runtime/app as an extension?  The generic baseline should handle common things such as 1) defining the route for `/invocations` and 2) helper utilities for extracting the session id.

integrations/mason/templates/durable-langgraph-agent/agent/agent.py
jamesbxwu, line 1: 1. file name should include langgraph?
2. the template name should avoid mentioning "app" and be more specific such as `durable-langgraph-agent`?
3. do we need separate agent and runtime folders in this template? Why not just have `agent.py` and `main.py` at the top level?

integrations/mason/src/databricks_mason/runtime/durability/types.py
```diff
@@ -0,0 +1,147 @@
...
+
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass, field
+from datetime import datetime
+from enum import Enum
+from typing import Protocol, TypeAlias
+
+JsonValue: TypeAlias = None | bool | int | float | str | list["JsonValue"] | dict[str, "JsonValue"]
+JsonObject = dict[str, JsonValue]
+DurableEventEmitter = Callable[[JsonObject], Awaitable[int]]
```
jamesbxwu, line 23: this needs more explanation

integrations/mason/templates/durability-app/pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[project]
+name = "durability-app"
```
jamesbxwu: similar to earlier comment, this name feels way too generic.  How about something like `durable-langgraph-agent`

integrations/mason/src/databricks_mason/agent_durability_store.py
```diff
@@ -0,0 +1,78 @@
...
+        branch=_BRANCH,
+        endpoint_id=_ENDPOINT,
+        database=_DATABASE,
+        schema=runtime_schema(app),
+        tables=(),
+        resource_name=_RESOURCE_NAME,
+    )
+
+
+def ensure_backend(app: str, profile: Optional[str], *, create: bool) -> LakebaseBackend:
```
jamesbxwu:
```suggestion
def get_or_create_backend(app: str, profile: Optional[str], *, create: bool) -> LakebaseBackend:
```

integrations/mason/src/databricks_mason/agent_durability_store.py
```diff
@@ -0,0 +1,78 @@
...
+
+def _project_id(app: str) -> str:
+    normalized = re.sub(r"[^a-z0-9-]+", "-", app.lower()).strip("-")
+    normalized = normalized or "mason-app"
+    if not normalized[0].isalpha():
+        normalized = f"mason-{normalized}"
+    return f"{normalized}-durability"[:63].rstrip("-")
+
+
+def runtime_schema(app: str) -> str:
```
jamesbxwu:
```suggestion
def get_lakebase_schema(app: str) -> str:
```

integrations/mason/src/databricks_mason/lakebase_durability_store.py
jamesbxwu, line 1: this is lakebsase specific, can we rename to `lakebase_durability_store.py`?

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,324 @@
+"""Transport-neutral durable request runtime."""
```
jamesbxwu: this class warrants better document on what is the mechanism to achieve durability, any persistent storage required and etc

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,324 @@
...
+        if not execution_id:
+            raise ValueError("execution_id must not be empty")
+        state = await self.durability_store.accept(
+            execution_id,
+            _copy_json_value(request, "request"),
+        )
+        self._ensure_scheduled(state)
+        return state
+
+    async def invoke(
```
jamesbxwu: difference between this and `execute` above is unclear to me

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,324 @@
...
+class DurableRuntime:
+    """Execute idempotent JSON requests with Lakebase-backed crash recovery.
+
+    The runtime persists only request execution state. The executor owns agent
+    sessions, checkpoints, tools, and any recovery-specific prompt or behavior.
+    """
+
+    def __init__(
+        self,
+        executor: DurableExecutor | None = None,
```
jamesbxwu:
```suggestion
        execute_fn: DurableExecutor | None = None,
```

integrations/mason/src/databricks_mason/runtime/types.py
```diff
@@ -0,0 +1,147 @@
...
+        return self.attempt > 1
+
+    async def emit(self, event: JsonObject) -> int:
+        """Persist an ordered event and return its replay cursor."""
+        if self._emit is None:
+            raise RuntimeError("event emission is not available for this execution context")
+        return await self._emit(event)
+
+
+DurableExecutor = Callable[[JsonValue, DurableExecutionContext], Awaitable[JsonValue]]
```
jamesbxwu:
```suggestion
DurableExecutorFn = Callable[[JsonValue, DurableExecutionContext], Awaitable[JsonValue]]
```

integrations/mason/src/databricks_mason/runtime/durability/runtime.py
```diff
@@ -0,0 +1,324 @@
...
+            raise ValueError("scan_seconds must be positive")
+        if poll_seconds <= 0:
+            raise ValueError("poll_seconds must be positive")
+
+        self._executor = executor
+        self.durability_store = durability_store
+        self.heartbeat_seconds = heartbeat_seconds
+        self.stale_seconds = stale_seconds
+        self.scan_seconds = scan_seconds
+        self.poll_seconds = poll_seconds
```
jamesbxwu, line 56: this class feels bloated right now, can we refactor/decompose it into the following but keep `runtime` as the facade layer that delegates to the following:
1. `AttemptRunner` (attempt.py) - _execute_attempt, _heartbeat_loop, and the emit closure.
2. `RecoveryScheduler` (recovery.py) - _tasks, _ensure_scheduled, _scan_loop, _discard_task, and recoverability.

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,324 @@
...
+        execution_id: str,
+        request: JsonValue,
+        *,
+        timeout: float | None = None,
+    ) -> JsonValue:
+        """Accept a request and wait for its persisted terminal response."""
+        await self.submit(execution_id, request)
+        return await self.wait(execution_id, timeout=timeout)
+
+    async def get(self, execution_id: str) -> DurableExecution | None:
```
jamesbxwu:
```suggestion
    async def get_execution(self, execution_id: str) -> DurableExecution | None:
```

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,324 @@
...
+                await asyncio.sleep(self.poll_seconds)
+
+        if timeout is None:
+            return await poll()
+        try:
+            return await asyncio.wait_for(poll(), timeout=timeout)
+        except asyncio.TimeoutError as exc:
+            raise TimeoutError from exc
+
+    async def events(
```
jamesbxwu:
```suggestion
    async def get_events(
```

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def invoke(self, function: AgentHook) -> AgentHook:
+        """Register the handler for a run's first attempt."""
+        if self._invoke_hook is not None:
+            raise ValueError("an invocation handler is already registered")
+        self._invoke_hook = function
+        return function
+
+    def on_recovery(self, function: AgentHook) -> AgentHook:
```
jamesbxwu: to be symmetrical with `invoke`
```suggestion
    def recover(self, function: AgentHook) -> AgentHook:
```

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -0,0 +1,219 @@
...
+    async def _execute(
+        self,
+        request: JsonValue,
+        execution_context: DurableExecutionContext,
+    ) -> JsonValue:
+        if not isinstance(request, dict):
+            raise TypeError("persisted request must be an object")
+        session_id = request.get("session_id")
+        if not isinstance(session_id, str) or "input" not in request:
+            raise TypeError("persisted request must contain session_id and input")
```
jamesbxwu: what does "persisted request" mean? should this app layer know about what's persisted?

integrations/mason/src/databricks_mason/runtime/types.py
```diff
@@ -0,0 +1,147 @@
...
+
+    @property
+    def is_recovery(self) -> bool:
+        return self.attempt > 1
+
+    async def emit(self, event: JsonObject) -> int:
+        return await self._execution_context.emit(event)
+
+
+AgentHook = Callable[[JsonValue, DurableAgentContext], Awaitable[JsonValue]]
```
jamesbxwu: is this durable runtime specific? if so, `AgentHook` feels too generic

integrations/mason/src/databricks_mason/runtime/app.py
shivam5: Addressed in `51feb39`: the warning now explicitly says that automatic crash recovery is disabled.

integrations/mason/src/databricks_mason/runtime/app.py
shivam5: Addressed in `51feb39`: the FastAPI title is now `Databricks Durable Agent Runtime`.

integrations/mason/src/databricks_mason/runtime/app.py
shivam5: Keeping `run_id` intentionally. The public contract uses the request `id` as a run ID and idempotency key across foreground, background, and streaming modes, so the status and event routes use the same terminology.

integrations/mason/src/databricks_mason/runtime/types.py
shivam5: Keeping `run_id` intentionally to match the public request/route contract. It identifies the durable run and serves as its idempotency key; `invocation_id` would introduce a second term for the same value.

integrations/mason/src/databricks_mason/runtime/types.py
shivam5: Addressed in `51feb39`: the module documentation now explains how the app, runtime, executor, contexts, store, execution snapshots, events, and JSON payload types relate.

integrations/mason/src/databricks_mason/runtime/durability/types.py
shivam5: Addressed in `51feb39`: every `DurabilityStore` protocol method now documents its contract and ownership semantics.

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: Keeping the approved flat SDK structure: `runtime/{app,runtime,store,types}.py`. Durability is the runtime introduced by this PR, and another nested package would make the import and implementation shape more complex without adding a second runtime abstraction yet.

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: Not adding a generic baseline app/runtime in this PR. The intended scope is the durable runtime and its thin FastAPI adapter; introducing a generic superclass before another SDK app implementation needs it would add abstraction and broaden the change. Existing Mason template runtimes remain unchanged.

integrations/mason/templates/durability-app/agent/agent.py
shivam5: Keeping the requested `durability-app` template name and the standard Mason `agent/agent.py` plus `runtime/main.py` separation. The agent file owns the LangGraph workload while the runtime file owns only SDK app wiring and the server entrypoint; this also stays consistent with existing Mason template structure.

integrations/mason/src/databricks_mason/runtime/durability/types.py
shivam5: Addressed in `51feb39`: the module documentation now explains that `JsonValue`/`JsonObject` are the recursive payload boundary shared by the runtime and both store implementations.

integrations/mason/templates/durability-app/pyproject.toml
shivam5: Keeping `durability-app` as the intentionally selected template/project name for this initial durability CUJ. Framework selection is already explicit through `mason init --framework langgraph --durability`.

integrations/mason/src/databricks_mason/agent_durability_store.py
shivam5: Addressed in `51feb39`: renamed to `get_or_create_backend` and updated callers/tests.

integrations/mason/src/databricks_mason/agent_durability_store.py
shivam5: Addressed in `51feb39`: renamed to `get_lakebase_schema` and updated callers/tests.

integrations/mason/src/databricks_mason/lakebase_durability_store.py
shivam5: Addressed in `51feb39`: renamed the module and test to `lakebase_durability_store.py` / `lakebase_durability_store_test.py`.

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: Addressed in `51feb39`: `DurableRuntime` now documents immutable request acceptance, atomic claims, heartbeats, stale recovery, attempt fencing, event replay, and the in-memory/Lakebase storage model.

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: Addressed in `51feb39`: the constructor parameter is now `execute_fn`; the callable alias is `DurableExecutorFn` per the adjacent naming suggestion.

integrations/mason/src/databricks_mason/runtime/types.py
shivam5: Addressed in `51feb39`: renamed the callable type to `DurableExecutorFn`.

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,324 @@
...
+            raise ValueError("scan_seconds must be positive")
+        if poll_seconds <= 0:
+            raise ValueError("poll_seconds must be positive")
+
+        self._executor = executor
+        self.durability_store = durability_store
+        self.heartbeat_seconds = heartbeat_seconds
+        self.stale_seconds = stale_seconds
+        self.scan_seconds = scan_seconds
+        self.poll_seconds = poll_seconds
```
shivam5: Keeping this as one runtime facade for now. We intentionally removed separate attempt/recovery modules to preserve the approved simple four-file SDK shape; the attempt and scanner logic are private and cohesive here. We can extract them later if independent implementations or materially more behavior justify the extra surface.

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: Addressed in `51feb39`: renamed the runtime method to `get_execution` and updated callers/tests.

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: Addressed in `51feb39`: renamed the runtime method to `get_events` and updated callers/tests.

integrations/mason/src/databricks_mason/runtime/app.py
shivam5: Keeping `on_recovery` intentionally. It names the lifecycle hook explicitly and avoids conflating crash recovery with HITL resume/recover operations. The API remains symmetrical as instance-scoped decorators: `@app.invoke` and `@app.on_recovery`.

integrations/mason/src/databricks_mason/runtime/app.py
shivam5: Addressed in `51feb39`: app-layer variables and errors now call this the `execution request`; persistence remains an internal runtime/store concern.

integrations/mason/src/databricks_mason/runtime/types.py
shivam5: Addressed in `51feb39`: renamed the app-specific callback type to `DurableAgentHook`.

integrations/mason/src/databricks_mason/runtime/app.py
shivam5: on_recovery feels to denote more clearly the intent of “register a handler for recovery attempts.”
recover sounds like calling recovery directly.

integrations/mason/src/databricks_mason/runtime/durability/runtime.py
shivam5: This acts like a general runtime as well if someone doesn't implement `on_recovery`

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: This acts both as the durability and general package

integrations/mason/templates/durable-langgraph-agent/agent/agent.py
shivam5: 3. following the current template structure, and it separate agent code from runtime cleanly. we can restructure all of them, but I'll let it be the same for this pr

integrations/mason/src/databricks_mason/runtime/app.py
shivam5: Follow-up in `4cdbe18`: agreed after further discussion and renamed the public route parameters to `invocation_id`; the lower-level runtime keeps `execution_id`.

integrations/mason/src/databricks_mason/runtime/types.py
shivam5: Follow-up in `4cdbe18`: renamed the public `DurableAgentContext` field to `invocation_id` for consistency with the invocation API.

integrations/mason/templates/durable-langgraph-agent/agent/agent.py
shivam5: Follow-up in `4cdbe18`: renamed the template to `durable-langgraph-agent`. I retained `agent/agent.py` and `runtime/main.py` because that separation matches the existing LangGraph template and keeps workload code separate from server wiring.

integrations/mason/templates/durability-app/pyproject.toml
shivam5: Follow-up in `4cdbe18`: renamed the project package to `durable-langgraph-agent`.

integrations/mason/src/databricks_mason/runtime/durability/runtime.py
shivam5: Follow-up in `4cdbe18`: decomposed the implementation into `AttemptRunner` (`attempt.py`) and `RecoveryScheduler` (`recovery.py`), with `DurableRuntime` remaining the facade.

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: Follow-up in `f97d74c`: agreed after revisiting the structure. The six durability-specific modules now live under `databricks_mason.runtime.durability`, while `DurableAgentApp` and `DurableAgentContext` remain re-exported from `databricks_mason.runtime` and `databricks_mason`.

PR #477, opened 2026-08-26
Prototype generic JSON durable agent server

shivam5:

This is a **draft interface prototype** for a supplied generic JSON server. It has the same runtime capabilities as the decorator app in #476, but the developer constructs and configures the server explicitly.

### Developer contract

```python
async def agent(payload, context): ...
async def resume_agent(payload, context): ...

server = DatabricksDurableServer(
    agent,
    on_resume=resume_agent,
)
app = server.app
```

The server owns the fixed JSON `/invocations` protocol, background and blocking execution, heartbeat recovery, final-result storage, polling, and cursor-based SSE replay.

- A request uses `background=true` and `stream=true`.
- A streamed proposal completes with a persisted `requires_action` result.
- The client submits approval as a second durable run with the same session ID.
- OpenAI Agents SDK state is stored in `AsyncDatabricksSession`.
- SDK events are persisted through `context.emit()`.
- The approved run includes a 60-second tool call so the process can be stopped and recovered through the explicit `on_resume` callback.

| Capability | Result |
| --- | --- |
| Request/response shape | Fixed generic JSON `/invocations` protocol |
| Background execution | Provided |
| Pod/process recovery | Provided through `on_resume` |
| Durable stream replay | Provided through `context.emit()` |
| Final result | Stored in the durable run row |

## How do you know it works?

- `27` focused runtime/store/server tests pass.
- Tests cover recovery callback selection and background stream submission.
- Ruff check and format pass for source, tests, and cookbook files.

| --- | --- |
| Background | `background=true, stream=false` returns `202`; the server stores status/output and the client polls `GET /invocations/{id}`. |
| Streaming | The developer calls `context.emit()`; `stream=true` returns SSE and the client reconnects with `?after=<event-id>`. |
| OpenAI Agents SDK client | An app-specific client changes to the generic `/invocations` envelope; the in-process `Runner.run_streamed()` loop remains inside the registered handler. |
| LangGraph SDK client | Native threads/runs calls change to `/invocations`; `thread_id` is passed as `session_id`. Preserving `langgraph_sdk` requires a protocol adapter. |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ### Improvements
 - databricks-openai: `DatabricksOpenAI` and `AsyncDatabricksOpenAI` clients now follow HTTP redirects by default, configurable via the new `follow_redirects` parameter (#445)
+- databricks-ai-bridge: Add a transport-neutral `DatabricksDurableRuntime` with Lakebase request, response, event persistence, and stale-attempt recovery
+- databricks-ai-bridge: Add a generic JSON `DatabricksDurableServer` prototype
 
 ### Bug Fixes
 - databricks-ai-bridge: Genie now returns the full answer text aggregated from all text attachments (#432)
--- a/README.md
+++ b/README.md
@@ -24,6 +24,21 @@ For frameworks without dedicated integration packages:
 pip install databricks-ai-bridge
 ```
 
+## Durable Runtime
+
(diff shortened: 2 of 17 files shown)


shivam5:
Interface comparison set:

shivam5:
Prototype. Actual durable runtime sdk merged in https://github.com/databricks/databricks-ai-bridge/pull/476

PR #480, opened 2026-08-27
Prototype header-based durable agent entrypoint

shivam5:

This is a **draft interface prototype** for preserving an application's existing JSON request body while adding durable execution metadata through HTTP headers. It is an alternative to the wrapped JSON contract in #476 and the explicit server in #477.

### Developer contract

```python
app = DatabricksDurableApp()

@app.entrypoint
async def agent(payload, context): ...

@app.on_resume
async def resume_agent(payload, context): ...
```

The client keeps its existing `/invocations` body. It supplies small execution metadata separately:

```http
POST /invocations
Idempotency-Key: approval-1
Databricks-Agent-Session-Id: approval-session-1
Databricks-Background: true
Databricks-Stream: true

{"action":"publish the release notes","decision":"approve","wait_seconds":60}
```

The server validates and persists the original body plus normalized session metadata. It constructs `context.run_id`, `context.session_id`, `context.attempt`, `context.is_recovery`, and `context.emit()` for the agent. After a crash there are no request headers to replay; the server reconstructs context from the durable run record.

- A background streamed proposal completes with `requires_action`.
- The client polls the persisted result and submits approval as a second run with the same session header.
- OpenAI Agents SDK state is stored under `context.session_id`.
- SDK events are persisted through `context.emit()` and replayed with cursors.
- The approved run includes a 60-second tool call so the process can be stopped and recovered through `@app.on_resume`.

| Capability | Result |
| --- | --- |
| Existing JSON request body | Preserved |
| Foreground response body | Preserved |
| Background execution | Supported through headers |
| Pod/process recovery | Supported |
| Durable stream replay | Supported through `context.emit()` |
| Final result | Stored in the durable run row |

## How do you know it works?

- `29` focused runtime/store/app tests pass.
- Tests cover header-to-context mapping, unchanged foreground response bodies, recovery callback selection, and background stream submission.
- Ruff check and format pass for source, tests, and cookbook files.

| --- | --- |
| Background | The existing body is unchanged. Headers select background mode; the app stores status/result and the client polls `GET /invocations/{run_id}`. |
| Streaming | The developer calls `context.emit()`. `Databricks-Stream: true` returns SSE; reconnect uses `?after=<event-id>`. |
| OpenAI/custom client | The route and body can remain unchanged; the client adds durability headers plus polling/replay logic. |
| LangGraph SDK client | Headers alone cannot preserve LangGraph's multi-route threads/runs protocol. Without an adapter, the client changes to the generic invocation route and passes `thread_id` in the session header. |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ### Improvements
 - databricks-openai: `DatabricksOpenAI` and `AsyncDatabricksOpenAI` clients now follow HTTP redirects by default, configurable via the new `follow_redirects` parameter (#445)
+- databricks-ai-bridge: Add a transport-neutral `DatabricksDurableRuntime` with Lakebase request, response, event persistence, and stale-attempt recovery
+- databricks-ai-bridge: Add an AgentCore-style `DatabricksDurableApp` entrypoint prototype
 
 ### Bug Fixes
 - databricks-ai-bridge: Genie now returns the full answer text aggregated from all text attachments (#432)
--- a/README.md
+++ b/README.md
@@ -24,6 +24,21 @@ For frameworks without dedicated integration packages:
 pip install databricks-ai-bridge
 ```
 
+## Durable Runtime
+
+[`DatabricksDurableRuntime`](./src/databricks_ai_bridge/durable_runtime/README.md)
+adds Lakebase-backed request, result, and event persistence with heartbeat-based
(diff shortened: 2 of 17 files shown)


shivam5:
Prototype. Actual durable runtime sdk merged in https://github.com/databricks/databricks-ai-bridge/pull/476

shivam5:
Prototype. Actual durable runtime sdk merged in https://github.com/databricks/databricks-ai-bridge/pull/476

PR #481, opened 2026-08-27
feat(mason): add downscoped sandbox command

junchoi-db:
## What

Adds `mason add-sandbox`, a focused source-generation command that attaches only
`system.ai.sandbox` to a Mason agent project and fixes its resource downscope in
MCP `_meta`, outside model-controlled tool arguments.

The command supports repeatable Unity Catalog volume, table, and Workspace path
scopes, read-only/read-write permissions, idempotent reruns, strict template
validation, and atomic updates to `agent/mcps.py`.

The generated client targets the public managed MCP route
`/ai-gateway/mcp-services/system.ai.sandbox`, uses a 120-second response timeout,
and accepts both advertised tool names during rollout: production `sandbox` and
older preview `run_code`.

## Template structure

The generated implementation now lives in the valid Python resource
`templates/sandbox_mcp.py`. That file owns its imports, fixed-downscope wrapper,
public endpoint, and server factory, so the implementation is readable and can be
syntax-checked directly.

The CLI reads the resource once and uses Python AST nodes only to locate the
template imports, the marked downscope assignment, and safe edit points in
`agent/mcps.py`. The large embedded Python string is gone.

This is still source generation today: the app framework calls its MCP runtime,
which calls `build_mcp_servers()`, which calls the generated
`_build_sandbox_mcp_server()`. The follow-up `agent.toml` direction can remove
per-MCP Python generation: a shared framework loader reads declarative MCP and
downscope entries at runtime, constructs the downscoped server, and appends it
before tool discovery. Mason would then edit TOML only; the generic loader/factory
would ship once.

## Why

Agent authors need a small, explicit path to add the shared sandbox MCP without
exposing downscope selection as an LLM-controlled tool argument.

## Design doc

[Mason CLI CUJ](https://docs.google.com/document/d/1Oi8k63r8-wJbJn5TFdDqjndDiveOkRm5aGlqE6e2AlQ)

## Proof of work

[df1 E2E metric report](https://metric-reports-viewer-6051921418418893.staging.aws.databricksapps.com/20260828_mason-add-sandbox-df1-e2e.html)

Production `df1` flow against the existing OpenAI agent app:

1. Built the `99cf0da` Mason wheel and verified that it contains and can read/compile
   `databricks_mason/templates/sandbox_mcp.py`.
2. Reset only the existing app's generated sandbox section; did not rerun
   `mason init`.
3. Ran the packaged `mason add-sandbox --scope table:samples.nyctaxi.trips`;
   generated `agent/mcps.py` compiled successfully.
4. Public `system.ai.sandbox` `tools/list` returned exactly `["sandbox"]`.
5. Called `sandbox` with
   `SELECT COUNT(*) AS trip_count FROM samples.nyctaxi.trips`.
6. The live result was `is_error=false`, content `[[21932]]`.

This proves the positive read path for the fixed read-only table scope. An
out-of-scope negative denial test was not run, so the report does not claim proof
of denial behavior.

The first df1 attempt also caught rollout skew: production advertises `sandbox`
while Tilefood previously advertised `run_code`. Incremental commit `99cf0da`
allows both names while keeping the endpoint fixed to `system.ai.sandbox`.

## Testing

- PR head: `99cf0da`; base: `0c53897` on `main`.
- 25/25 focused sandbox and CLI tests pass (JUnit: 0 errors, 0 failures).
- 60/60 broader Mason tests pass; one unrelated environment-dependent tracing
  test was deselected because this environment loads MLflow and attempts live setup.
- Ruff check, Ruff format check, and `ty check` pass.
- Wheel resource inclusion plus installed-resource read/compile pass.
- Production df1 `tools/list` and downscoped SQL `tools/call` pass.
- GitHub Actions: 49/50 checks pass, including every Mason check. The only
  failure is Ty typechecking—not Ruff lint—in untouched `integrations/openai`
  files ([job](https://github.com/databricks/databricks-ai-bridge/actions/runs/33130367239/job/98718211223)).
  The exact CI command reproduces identically on PR head `99cf0da` and base SHA
  `0c53897`: both resolve `openai 3.3.1`, whose client types use `httpx2`, while
  the existing integration passes `httpx 0.28.1`. This PR changes only
  `integrations/mason/**`.

### DIFF
--- a/integrations/mason/README.md
+++ b/integrations/mason/README.md
@@ -59,7 +59,38 @@ mason [-p <profile>] [-o text|json]
   tracing
     setup      --catalog C --schema S [--experiment E]
     list | get | instrument
+  add-sandbox  --scope SCOPE [--scope SCOPE ...]
+               [--permission read_only|read_write] [--source PATH]
   deploy       <name> --source PATH [--with-memory-store N]
                [--with-session-store N] [--with-traces C.S] [--create-stores]
   deployments  list | get | logs | start | stop | delete
 ```
+
+## Add a downscoped sandbox
+
+From a Mason agent project, add the `system.ai.sandbox` MCP server and fix its
+allowed resources at configuration time:
+
+```sh
+mason add-sandbox --scope catalog.schema.volume
+```
+
+The command updates `agent/mcps.py`. Every sandbox call carries the selected
+downscope in MCP `_meta`, outside the tool arguments controlled by the model.
+Scopes default to read-only access. Repeat `--scope` to allow more than one
+resource, use `/Workspace/...` for a workspace path, or prefix a table with
+`table:`:
+
+```sh
+mason add-sandbox \
+  --scope catalog.schema.volume \
+  --scope /Workspace/Users/alice@example.com \
+  --scope table:catalog.schema.table
+```
+
+Pass `--permission read_write` to grant write access to every supplied scope,
+or `--source /path/to/project` when running outside the project root. Re-running
+the command with the same policy leaves the existing configuration unchanged;
+a different policy fails without modifying the file so it cannot silently report
+stale access. Edit or remove the generated block before intentionally changing
+the sandbox policy.
--- a/integrations/mason/src/databricks_mason/sandbox.py
+++ b/integrations/mason/src/databricks_mason/sandbox.py
@@ -1,0 +1,581 @@
+"""``mason add-sandbox`` — attach a downscoped ``system.ai.sandbox`` MCP server."""
+
+from __future__ import annotations
+
+import ast
+import json
+import pathlib
+import stat
+import tempfile
+import textwrap
+from collections.abc import Sequence
+
+import click
+
+from databricks_mason import render
+from databricks_mason.errors import AgentCliError
+
+_BEGIN_MARKER = "# BEGIN: mason add-sandbox"
+_END_MARKER = "# END: mason add-sandbox"
+_SERVER_FACTORY = "_build_sandbox_mcp_server()"
+
+_GENERATED_TYPING_IMPORT = "from typing import Any"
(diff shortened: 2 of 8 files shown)


integrations/mason/src/databricks_mason/sandbox.py
```diff
@@ -0,0 +1,583 @@
...
+
+def _find_server_list(tree: ast.Module) -> tuple[ast.FunctionDef, ast.List]:
+    functions = [
+        node
+        for node in tree.body
+        if isinstance(node, ast.FunctionDef) and node.name == "build_mcp_servers"
+    ]
+    if len(functions) != 1:
+        raise AgentCliError(
+            "Could not find one build_mcp_servers() function in agent/mcps.py.",
```
elainewang-db, line 110: in the future we should make a common util in the template or something to set up system.ai uc mcps instead of having to parse the mcps.py file and edit it, like the agent.toml as you mentioned before. looks like a lot of the logic here is brittle file parsing?

elainewang-db (approved):

integrations/mason/src/databricks_mason/sandbox.py
junchoi-db: yes - will follow up!

integrations/mason/README.md
```diff
@@ -59,7 +59,38 @@ mason [-p <profile>] [-o text|json]
...
+
+```sh
+mason add-sandbox --scope catalog.schema.volume
+```
+
+The command updates `agent/mcps.py`. Every sandbox call carries the selected
+downscope in MCP `_meta`, outside the tool arguments controlled by the model.
+Scopes default to read-only access. Repeat `--scope` to allow more than one
+resource, use `/Workspace/...` for a workspace path, or prefix a table with
+`table:`:
```
jamesbxwu, line 82: should we go with prefix or make it more explicit via a `type` parameter?

PR #478, opened 2026-08-27
[templates] Add agent-langgraph-scratch template

elainemwang:
A from-scratch FastAPI LangGraph agent backend for Databricks Apps: native LangGraph wire on both ends, in-memory checkpointer, streaming + background, MCP loading, and human-in-the-loop tool approval (HumanInTheLoopMiddleware).

PR #483, opened 2026-08-27
Add LangGraph chat app option to `mason init`

shivam5:
## Summary

- add `mason init --framework langgraph --enable-chat-app`
- keep the chat overlay framework-specific under [`integrations/mason/templates/ui/agent-langgraph`](https://github.com/shivam5/databricks-ai-bridge/tree/mason-add-ui/integrations/mason/templates/ui/agent-langgraph)
- always include Start App and Stop App; there are no separate stop/crash flags
- keep durability, recovery, heartbeat, and long-running tool components in `integrations/mason/templates/agent-langgraph/agent/mason`
- demonstrate sync, SSE streaming, background polling, session history, HITL resume, managed memory, and managed Session Store
- use `__Host-databricks-app-router` as both the Apps routing key and application session id
- do not send `session_id` in request bodies; localhost uses an HTTP-only `mason-local-session` fallback cookie
- add a TODO to move to `X-Routing-Key` when Databricks Apps supports it
- remove the unrelated OpenAI client and Mason workspace-client changes

## Demo flow

```bash
mason init --framework langgraph --enable-chat-app \
  --profile e2-dogfood \
  --repo https://github.com/shivam5/databricks-ai-bridge.git \
  --ref mason-add-ui \
  ~/my-agent

cd ~/my-agent
uv run start-server
```

For the managed-state and stop/start demo:

```bash
mason --profile e2-dogfood deploy mason-agent-demo --source . \
  --with-session-store mason-demo-sessions \
  --with-memory-store mason-demo-memory \
  --actor-id alice \
  --create-stores
```

## Client contract

The generated README documents the exact requests made by the browser:

- `GET /api/demo/config` discovers the routing-backed session id and capability state
- `POST /invocations` performs sync, SSE streaming, or background submission
- `GET /invocations/{id}` polls background runs with the same routing cookie
- `POST /api/demo/sessions` resolves the current managed session
- `GET /api/demo/session/items` loads messages and pending interrupts into the chat
- `POST /api/demo/session/items` mirrors transcript items into Session Store
- memory entry list/create/search routes use the configured Memory Store and actor
- HITL resume posts only the native `resume` decision with the same routing cookie
- Start App runs `tool_step_1` through `tool_step_4`; Stop App schedules `os._exit(86)`

Capability colors are automatic from `/api/demo/config`. Only the sync/streaming/background transport selector is manual.

## Stop/start durability

- each completed tool output is checkpointed before the next node runs
- attempts, owner ids, heartbeats, completion, and failure events are persisted in Session Store
- heartbeats default to every 3 seconds and become stale after 10 seconds
- Databricks Apps restarts the process after Stop App; locally, restart the server manually or with a supervisor
- the browser keeps the same routing cookie, waits for a new process, detects the stale owner, and starts a new attempt
- completed nodes are restored and skipped; an interrupted node can retry if its output was not committed

The lease is intentionally demo-grade and last-writer-wins (`atomic_claim: false`). Production durability still needs atomic ownership, server-side stale scanning, idempotent side effects, and durable event replay.

## Validation — August 28, 2026

- Mason unit tests: `108 passed`
- Mason Ruff check and format check: passed
- Mason `ty check`: passed
- generated chat app: `21 passed, 1 skipped`
- JavaScript syntax: passed
- real sparse-clone `mason init --enable-chat-app`: passed
- built-wheel install and init: passed

## Context

Rebased directly on current `main` after #474 and #478 merged.


integrations/mason/templates/agent-langgraph/agent/mason/memory.py
```diff
@@ -14,30 +14,52 @@
...
 
 
 def _actor_id() -> str:
     return os.getenv("AGENT_MEMORY_ACTOR_ID", "agent")
 
 
 def _store_path() -> str:
-    return f"{_AGENTS_V1}/memory-stores/{os.environ['AGENT_MEMORY_STORE']}"
+    store = os.environ["AGENT_MEMORY_STORE"].strip().strip("/")
+    store = store.removeprefix("memory-stores/")
```
elainewang-db: qq: is this due to the way `mason deploy --with-memory-store` is? My pr should fix this so the template doesn't need to do this.

integrations/mason/templates/agent-langgraph/README.md
```diff
@@ -244,7 +273,9 @@ replicas, over RPCs only. No agent code changes; the checkpointer swap is the on
 | `DATABRICKS_CONFIG_PROFILE` | `DEFAULT` | Auth profile used to call the model (local dev) |
 | `PORT` | `8000` | Port the server listens on |
 | `AGENT_MEMORY_STORE` | _unset_ | Managed memory store name → registers `remember`/`recall` long-term-memory tools |
+| `AGENT_MEMORY_ACTOR_ID` | `agent` | Actor partition used by memory tools and the optional UI |
 | `AGENT_SESSION_STORE` | _unset_ | Managed Session Store name → durable checkpointer (REST-backed); unset = in-process `InMemorySaver` |
+| `AGENT_SESSION_ACTOR_ID` | session id | Actor used by the durable saver and optional UI managed session |
```
elainewang-db: can we try to make the template not mention the UI?

integrations/mason/templates/agent-langgraph/runtime/runtime.py
```diff
@@ -6,8 +6,9 @@
     stream_handler(request: dict) -> AsyncGenerator[dict]
 
 Nothing here is SDK-specific — the agent lives entirely behind those handlers (``agent/agent.py``).
-The session id (used for multi-turn and to resume a paused run) travels in the ``X-Routing-Key``
-header, not the body; the runtime reads it and hands it to the handlers as ``session_id``.
+The session id (used for multi-turn and to resume a paused run) travels in the JSON body as
+``session_id``. Databricks Apps replica affinity is handled separately by the
```
elainewang-db: I thought we had decided on `X-Routing-Key`? https://databricks.slack.com/archives/C0BL69HA4DQ/p1787767718991539?thread_ts=1787345723.987959&cid=C0BL69HA4DQ

integrations/mason/templates/agent-langgraph/agent/agent.py
```diff
@@ -60,19 +71,43 @@ def _check_databricks_auth() -> None:
...
     return create_agent(
-        model=ChatDatabricks(endpoint=MODEL),
+        model=ChatDatabricks(endpoint=MODEL, workspace_client=_workspace_client()),
         tools=tools,
         middleware=middleware,
         checkpointer=checkpointer(),
     )
 
 
+async def session_history(session_id: str) -> dict:
```
elainewang-db: would it be possible to construct the session history ui side? can you call the session store apis?

integrations/mason/templates/ui/agent-langgraph/ui/styles.css
jamesbxwu, line 1: can we move all this new ui templates to `integrations/mason/templates/ui` alongside of the agent templates?

integrations/mason/src/databricks_mason/client.py
jamesbxwu, line 1: can we include a README for this UI template?

integrations/mason/src/databricks_mason/add.py
```diff
@@ -0,0 +1,259 @@
...
+    "--enable-crash",
+    "legacy_enable_crash",
+    is_flag=True,
+    hidden=True,
+)
+@click.option(
+    "--force",
+    is_flag=True,
+    help="Overwrite Mason-managed UI files to refresh an existing installation.",
+)
```
jamesbxwu: `--force` feels weird, shouldn't this be part of `mason dev` or `mason deploy` if local source changes are made?

integrations/mason/templates/agent-langgraph/runtime/runtime.py
jamesbxwu: we are still waiting on the changes on apps side to land but we should add a TODO here

jamesbxwu (commented):
A few high level comments on the CLI changes
1. do we need to define a separate `add ui` command instead of just adding a parameteter to the existing `init` command such as `mason init --framework langgraph --enable-chat-app`
2. I noticed extra knobs to the `add ui` command such as `--enable-stop` and `--enable-crash`,  can we just always include them?
3. can you add a link to the sample ui app in the PR description for reference?

integrations/mason/templates/agent-langgraph/agent/mason/memory.py
shivam5: yeah, this was because deploy was passing the full resource name. fixed it so deploy now writes the bare store id, and removed the prefix handling from the template.

integrations/mason/templates/agent-langgraph/README.md
shivam5: yep, removed the ui references from the base template. the ui docs are under `integrations/mason/templates/ui/README.md` now.

integrations/mason/templates/agent-langgraph/runtime/runtime.py
shivam5: we checked this again and `X-Routing-Key` is not available on apps yet. keeping the router cookie for now and added a TODO to switch once it lands.

shivam5: yep, added the TODO.

integrations/mason/templates/agent-langgraph/agent/agent.py
shivam5: yep, moved this ui side. managed sessions read items from Session Store, and local sessions read from the LangGraph checkpoint. removed `session_history` from the base agent.

integrations/mason/templates/ui/agent-langgraph/ui/styles.css
shivam5: done, moved all the ui template files under `integrations/mason/templates/ui` and verified they work from the packaged wheel.

integrations/mason/src/databricks_mason/client.py
shivam5: added `integrations/mason/templates/ui/README.md` with the template structure and how session, memory and durability work.

integrations/mason/src/databricks_mason/add.py
shivam5: changed this to `--refresh`. i do not think dev/deploy should rewrite local source - this is similar to `databricks init`

shivam5:
addressed these and pushed in `a1c5fe9`:

1. cahnged`mason add ui` to `mason init`
2. removed the enable stop/crash flags; start/stop is always included
3. added the ui template link to the pr description
4. changed `--force` to `--refresh`; dev/deploy will not rewrite local source

integrations/mason/templates/agent-langgraph/runtime/runtime.py
elainewang-db: I mean that the client should still send the session_id in the `X-Routing-Key` header. We don't rely on the apps side changes for this, our template server should just expect the session id to be in that header.

elainewang-db: like the session id should be sent like:
```
curl -s -X POST http://localhost:8000/invocations -H "Content-Type: application/json" \
  -H "X-Routing-Key: demo1" \
  -d '{"input":[{"role":"user","content":"hello! what tools do you have?"}]}' 
```
instead of in the body like 
```
curl -s -X POST http://localhost:8000/invocations -H "Content-Type: application/json" \
  -d '{"input":[{"role":"user","content":"hello! what tools do you have?", "session_id":"demo1"}]}' 
```

elainewang-db (approved):

integrations/mason/templates/agent-langgraph/runtime/runtime.py
shivam5: updated this based on the latest apps behavior. the client now uses __Host-databricks-app-router as both the sticky routing key and the app session id, and does not send session_id in the body

PR #485, opened 2026-08-28
mason deploy: fix memory-store resolution 

elainewang-db:
Two bugs in the --with-memory-store deploy path:
- Resolving an existing store listed with page_size=1000, which the memory-stores list API rejects (INVALID_PARAMETER_VALUE: page_size must be between 1 and 100). Page through with page_size=100 and the next_page_token instead.
- Without --create-stores, resolution used get_memory_store(name), which looks up by resource id and so can't find a store by its display name (NOT_FOUND). Resolve by display name (list + match) on both paths.

### DIFF
--- a/integrations/mason/src/databricks_mason/deploy.py
+++ b/integrations/mason/src/databricks_mason/deploy.py
@@ -160,8 +160,13 @@ def _ensure_session_store(client, name: str) -> dict:
 
 
 def _memory_store_database(client, memory_store: str) -> Optional[str]:
-    """Resolve the memory store's per-store Lakebase database name from its storage backend."""
-    store = client.get_memory_store(memory_store)
+    """Resolve the memory store's per-store Lakebase database name from its storage backend.
+
+    Resolves by display name (what the deploy flag carries), not get_memory_store (which is by id).
+    """
+    store = _resolve_memory_store(client, memory_store)
+    if store is None:
+        return None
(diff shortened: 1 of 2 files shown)


integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -115,17 +115,39 @@ def _upsert_manifest_env(source: pathlib.Path, updates: dict[str, str]) -> bool:
...
+def _resolve_memory_store(client, display_name: str) -> Optional[dict]:
+    """Find a memory store by display name, paging through the list, or None if none matches.
+
+    `get_memory_store` looks up by resource id (`memory-stores/<uuid>`), not the display name users
+    pass, so resolving a name means listing and matching on `display_name`. The list API caps
+    `page_size` at 100, so page through with the `next_page_token` rather than requesting all at once.
+    """
+    page_token: Optional[str] = None
+    while True:
+        listing = client.list_memory_stores(
```
junchoi-db, line 130: nit: just making sure this does not raise anything if we go end of page/sequence

junchoi-db (approved):

integrations/mason/src/databricks_mason/deploy.py
elainewang-db: good point, checked! we will only call `list_memory_stores` with the `page_token` it returned from the previous call which will be valid. if this raises an error it will be wrapped in `AgentCliError`

PR #502, opened 2026-09-01
[mason] Error hygiene: JSON-mode errors, id validation, friendly no-op update

anshul0713:
Three error-handling fixes in the Mason CLI (ML-69221, ML-69222, ML-69223).

- **ML-69221** In `-o json` mode, errors are now emitted as a JSON object `{"error": {message, code, hint}}` on stderr instead of styled text, so scripts driving `mason -o json` can parse failures. Text mode is unchanged.
- **ML-69222** Store id / resource name is validated client-side: empty or wrong-typed ids raise a clear error instead of building a URL like `.../memory-stores/` and leaking a raw `ENDPOINT_NOT_FOUND`. Covers memory stores, session stores, and memory entries.
- **ML-69223** A no-op `update` (no mutable fields) now raises a friendly "No fields to update…" message instead of surfacing the internal `update_mask is required`.

Tests: `client_test.py` (path + no-op cases), new `errors_test.py` (both output modes). Full suite green; verified live.

### DIFF
--- a/integrations/mason/src/databricks_mason/client.py
+++ b/integrations/mason/src/databricks_mason/client.py
@@ -36,15 +36,39 @@ def _as(cls: type, resp: Any) -> Any:
 
 
 def memory_store_path(name: str) -> str:
-    """Normalize a store id or name into the `memory-stores/{id}` resource segment."""
-    name = name.strip().strip("/")
-    return name if name.startswith("memory-stores/") else f"memory-stores/{name}"
+    """Normalize a store id or name into the `memory-stores/{id}` resource segment.
+
+    Validates client-side so an empty or wrong-typed argument raises a clear error
+    instead of building a URL like `memory-stores/` and leaking a raw ENDPOINT_NOT_FOUND.
+    """
+    raw = (name or "").strip()
+    if raw.startswith("memory-stores/"):
+        raw = raw[len("memory-stores/") :]
+    raw = raw.strip().strip("/")
+    if not raw:
+        raise AgentCliError("A memory store id or resource name is required.")
+    if "/" in raw:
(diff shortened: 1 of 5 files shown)


integrations/mason/src/databricks_mason/client.py
```diff
@@ -36,15 +36,39 @@ def _as(cls: type, resp: Any) -> Any:
...
+    """
+    raw = (name or "").strip()
+    if raw.startswith("memory-stores/"):
+        raw = raw[len("memory-stores/") :]
+    raw = raw.strip().strip("/")
+    if not raw:
+        raise AgentCliError("A memory store id or resource name is required.")
+    if "/" in raw:
+        raise AgentCliError(f"Invalid memory store id or resource name: {name!r}")
+    return f"memory-stores/{raw}"
```
akansal2, line 52: do we want to include this check in the session store path method as well?

integrations/mason/tests/unit_tests/client_test.py
```diff
@@ -167,3 +170,57 @@ def test_path_helpers():
...
+    # Empty / whitespace-only ids must not build a "memory-stores/" URL (ML-69222).
+    for bad in ["", "   ", "/", "memory-stores/"]:
+        with pytest.raises(AgentCliError):
+            memory_store_path(bad)
+    # A wrong-typed resource name is rejected rather than nested into the URL.
+    with pytest.raises(AgentCliError):
+        memory_store_path("session-stores/abc")
+
+
+def test_session_store_path_rejects_empty():
```
akansal2, line 185: shall we include the wrong type check that we do in the memory store path test method above, in this session store path method as well?

akansal2 (approved):
overall lgtm - few open comments to address though

PR #504, opened 2026-09-01
[mason] memory entries create: add --content-file for large/multiline content

anshul0713:
Adds `--content-file PATH` to `mason memory entries create` (ML-69229). `--content` is bounded by the shell ARG_MAX and there was no file/stdin alternative (while `sessions items append` has `--file`). The new flag reads content from a file and is mutually exclusive with `--content`.

Note: ML-69228 (`--write-mode` OVERWRITE/APPEND) is intentionally **not** included — live testing showed the server does not honor a `write_mode` field in the create body (plain or full-enum values still returned `ALREADY_EXISTS`), so the API contract needs confirmation from the owners first (see ML-69228).

Tests: `entries_input_test.py`. Full suite green; `--content-file` verified live.


integrations/mason/src/databricks_mason/memory.py
```diff
@@ -263,9 +273,13 @@ def _render_entry_detail(entry: dict) -> None:
...
 @click.pass_obj
 def entries_create(
-    obj, store, actor_id, path, content, description, session_id, source_type
+    obj, store, actor_id, path, content, content_file, description, session_id, source_type
 ) -> None:
     """Create a memory entry."""
+    if content is not None and content_file is not None:
+        raise AgentCliError("Pass either --content or --content-file, not both.")
+    if content_file is not None:
+        content = pathlib.Path(content_file).read_text()
```
akansal2, line 282: nit: should we pass in `encoding="utf-8"` into `read_text()`?

akansal2 (approved):
lgtm, just a small nit

PR #510, opened 2026-09-02
mason: grant store access over psycopg instead of the psql binary

elainewang-db:
Removes the system psql dependency for store deploys by issuing the owner GRANT through psycopg (a bundled dependency), so no Postgres client needs to be on PATH.


shaotong-db (approved):
Thank you!

PR #511, opened 2026-09-02
mason: add --min-instances to `mason deploy` for manual horizontal scaling

elainewang-db:
Wires `--min-instances` through to `databricks apps` compute scaling: passed to `apps create` for a new app, and applied via `apps update` before re-deploying an existing app's source (scaling isn't settable on `apps deploy`). Apps doesn't autoscale yet, so this is a fixed instance count.


elainewang-db:
superseded by https://github.com/databricks/databricks-ai-bridge/pull/515?timeline_per_page=5

PR #512, opened 2026-09-02
mason: add an OpenAI Agents SDK template mirroring agent-langgraph

elainewang-db:
Adds a second basic framework — `mason init --framework openai` — built on the OpenAI Agents SDK, mirroring the agent-langgraph template file-for-file:

- databricks_mason.openai adapter (sessions/mcp/memory/tracing) alongside the langgraph one, behind a new `runtime-openai` extra.
- templates/agent-openai + the templates/ui/agent-openai chat overlay. The agent translates Agents SDK stream events into the same {delta,message,interrupt} envelope, so the SDK-agnostic runtime and the browser UI are reused unchanged.
- init.py points the openai framework at this repo (versioned in lockstep with the CLI, chat app on by default).

HITL is in-process only: an Agents SDK Session persists the transcript, not paused RunState, so a paused run does not survive restarts/replicas even with a Session Store. Durable HITL is left for a follow-up, as is de-duping the copied chat UI.

### DIFF
--- a/integrations/mason/templates/agent-openai/agent/agent.py
+++ b/integrations/mason/templates/agent-openai/agent/agent.py
@@ -1,0 +1,252 @@
+import logging
+import os
+from collections.abc import AsyncGenerator
+from contextlib import AsyncExitStack
+from typing import Any
+
+from agents import Agent, RunResultStreaming, Runner, RunState
+from agents.items import ToolApprovalItem
+from databricks_openai import AsyncDatabricksOpenAI
+from openai.types.responses import ResponseTextDeltaEvent
+
+from databricks_mason import tag_session, workspace_client
+from databricks_mason.openai import configure_tracing, mcp_servers, memory_tools, session_store
+
+from agent.mcps import build_mcp_servers
+
+# Importing the tools package auto-registers every tool module.
+from agent.tools import all_tools
+
+logger = logging.getLogger(__name__)
+
+MODEL = "databricks-gpt-5-2"
+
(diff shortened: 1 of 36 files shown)


integrations/mason/templates/agent-openai/agent/agent.py
```diff
@@ -0,0 +1,252 @@
...
+    # Route the Agents SDK's default OpenAI client at the Databricks model endpoint (account-host
+    # routing and auth handled by the SDK), so `Agent(model=MODEL)` resolves to a Databricks model.
+    from agents import set_default_openai_api, set_default_openai_client
+
+    set_default_openai_client(AsyncDatabricksOpenAI())
+    set_default_openai_api("chat_completions")
+    configure_tracing()
+
+
+def _check_databricks_auth() -> None:
```
shivam5, line 49: can we move all databricks specific auth to mason probably as a util function?
We can add a todo to do this.

shivam5 (approved):

integrations/mason/templates/agent-openai/agent/agent.py
elainewang-db: yep, sounds good! there's quite of bit of template overlap that we should consolidate next.

junchoi-db (approved):

integrations/mason/templates/agent-openai/agent/agent.py
elainewang-db: https://databricks.atlassian.net/browse/ML-69337

PR #513, opened 2026-09-02
[mason] deploy/dev: rename store flags and create stores by default

elainewang-db:
Addresses CLI-ergonomics review feedback on `mason deploy` / `mason dev`:

- `--with-memory-store` → `--memory`, `--with-session-store` → `--session-store`. The old names stay as hidden, deprecated aliases so existing scripts don't break.
- `--create-stores` is gone as an opt-in flag; create-if-not-exists is now the default. `--no-create-stores` opts back into the fail-if-missing behavior.

Updates the CLI help, the mason README, the langgraph template README, and the chat-app "set it with --…" error messages to the new flags.

### DIFF
--- a/integrations/mason/src/databricks_mason/deploy.py
+++ b/integrations/mason/src/databricks_mason/deploy.py
@@ -210,9 +210,10 @@ def resolve_store_env(
     """Resolve store/trace references to the AGENT_*/MLFLOW_* env vars that wire them in.
 
     Shared by `mason deploy` and `mason dev` so both wire an agent's stores into app.yaml the same
-    way. With `create_stores`, missing stores are created (idempotent); otherwise they must already
-    exist. The memory store resolves to its bare id (the runtime re-adds the `memory-stores/` prefix
-    when building the entries URL); the session store and trace destination are used verbatim.
+    way. With `create_stores` (the default), missing stores are created (idempotent); when it is off
+    they must already exist. The memory store resolves to its bare id (the runtime re-adds the
+    `memory-stores/` prefix when building the entries URL); the session store and trace destination
+    are used verbatim.
     """
     env: dict[str, str] = {}
     if memory_store:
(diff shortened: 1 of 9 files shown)


integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -300,17 +302,19 @@ def _grant_store_access(
...
     help="Memory store display name to wire in via AGENT_MEMORY_STORE.",
 )
 @click.option(
-    "--with-session-store",
+    "--session-store",
     "session_store",
     default=None,
     help="Session store name to wire in via AGENT_SESSION_STORE.",
 )
+@click.option("--with-memory-store", "memory_store", default=None, hidden=True)
```
shivam5: we don't need to keep this around - there are no active users and we can make breaking changes

integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -300,17 +302,19 @@ def _grant_store_access(
...
 @click.option(
-    "--with-memory-store",
+    "--memory",
     "memory_store",
     default=None,
     help="Memory store display name to wire in via AGENT_MEMORY_STORE.",
 )
 @click.option(
-    "--with-session-store",
+    "--session-store",
```
shivam5: did you consider doing just --session?
also can we have short forms, such as -s for session store, -m for memory -> can be a follow up

shivam5 (approved):
lets remove the old options - approving to unblock

integrations/mason/src/databricks_mason/deploy.py
elainewang-db: yeah, but in the end just followed the notes https://docs.google.com/spreadsheets/d/1OawecuJMaSBRllCPDC3fAWHLeGZXX7lINkrt4tRF0XE/edit?gid=0#gid=0, we can do that too

PR #520, opened 2026-09-03
Make Mason login configure Databricks authentication

annzhang-db:
## Summary

Make `mason login --profile <name>` validate existing credentials and, when credentials are missing or rejected in an interactive terminal, delegate browser authentication to `databricks auth login`, revalidate, and save the profile. Other validation failures return without launching a browser, child login output stays on stderr so Mason JSON remains parseable, and the auth hint points users to Mason commands for either one-off profile selection or saved login.

## Testing

Added regression coverage for valid credentials skipping browser login, credential failures invoking and retrying Databricks login, clean JSON stdout during the fallback, non-auth failures avoiding reauthentication, non-interactive behavior, a missing Databricks CLI, and the profile-selection hint. The complete Mason unit suite passes (230 tests), along with Ruff lint/format checks and focused `ty` checks.

Live E2E against `https://eng-ml-inference.staging.cloud.databricks.com/` used a disposable `eng-ml-inference` profile containing only the host. `mason login --profile eng-ml-inference` invoked Databricks browser authentication, revalidated the profile, and saved it; a subsequent `mason mcp list` succeeded and returned six services. The temporary Databricks and Mason auth files were removed afterward.

Abridged terminal view (user and service rows omitted):

```text
$ mason login --profile eng-ml-inference
[Databricks CLI opens browser authentication]
[profile saved and revalidated]
Logged in as <user>
Profile  eng-ml-inference
Host     https://eng-ml-inference.staging.cloud.databricks.com

$ mason mcp list
MCP Services
Available in system.ai
...
6 items
```

PR #522, opened 2026-09-03
[mason] Flag a stray DATABRICKS_TOKEN in the auth-init error

akansal2:

**Context:** From the custom-agent bug bash, the P1 finding *"Auth errors are confusing/looping and suggest the wrong command (`databricks auth login` vs `mason login`)."* Since this PR was opened, most of that has landed on `main` through other PRs:

- the misleading fixed hint (`Check your profile … or pass --profile`) is gone;
- auth-init failures now surface the SDK message verbatim **and** point at the right command — `mason login --profile <name>` (in `_api_client.py`);
- `mason login` configures auth on its own (#520), so a separate `databricks auth login` is no longer needed.

So this PR is **rescoped to the one gap that didn't land**: my own bug-bash finding that a stray `DATABRICKS_TOKEN` silently overrides the selected profile ("had to unset my databricks token … not intuitive from the error messages").

**Change:** when workspace-client initialization fails and `DATABRICKS_TOKEN` is set in the environment, append a note to the auth-error hint explaining that the env token takes precedence over the profile and should be unset. The SDK's default auth resolution prefers the env token, so auth can fail against the wrong host even when the profile is correct — and nothing in the SDK error makes that obvious.

## How do you know it works?

Unit tests in `client_test.py` cover the token-set case (hint flags `DATABRICKS_TOKEN` and says to unset it) and the token-unset case (no note; base guidance intact). Also exercised `_MasonApiClient` directly against both environment states.

### DIFF
--- a/integrations/mason/src/databricks_mason/_api_client.py
+++ b/integrations/mason/src/databricks_mason/_api_client.py
@@ -124,10 +124,21 @@ class _MasonApiClient:
         try:
             self._w = workspace_client or _workspace_client(profile)
         except Exception as exc:  # noqa: BLE001 - surfaced as a clean CLI error
+            hint = (
+                "Select an existing profile with `mason --profile <name> <command>` "
+                "or authenticate and save it with `mason login --profile <name>`."
+            )
+            # A stray DATABRICKS_TOKEN takes precedence over the profile in the SDK's default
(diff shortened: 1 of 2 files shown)


akansal2:
Pairs with #523, which makes `mason login` the single auth command these hints point to.

elainewang-db (approved):

PR #523, opened 2026-09-03
[mason] mason login authenticates you (no separate databricks auth login)

akansal2:

**Change:** `mason login` now authenticates a profile on its own. If the profile has no usable credentials yet, it runs the Databricks OAuth sign-in for you (by shelling `databricks auth login`, already a requirement for `dev`/`deploy`), then validates and remembers it. An already-authenticated profile is just validated — no browser opens. Adds `--host` for first-time profile creation.

**Why:** From the custom-agent bug bash, users had to run *two* commands — `databricks auth login` then `mason login` — and error messages that said "run `databricks auth login`" were confusing when the fix that worked was `mason login` (Ann's report). `mason login` is now the only auth command needed.

Behavior:
- Try existing credentials first (cheap, no browser); only sign in if that fails, then retry once.
- Credentials still live in `~/.databrickscfg` exactly as the Databricks CLI writes them; Mason only stores the *selected* profile name in `~/.mason/config.json`.
- README updated to the single-command flow.

Pairs with #522 (clearer auth error messages), whose hints all point at this single `mason login`.

## How do you know it works?

**Testing:** `python -m py_compile` on the changed modules (clean). Added unit tests in `auth_test.py`: `mason login` runs the sign-in exactly once when the profile has no credentials (and persists the selection), and skips the sign-in entirely when the profile already authenticates. Existing `auth_test.py` cases are unchanged. Full suite runs in CI. Draft until CI is green.

### DIFF
--- a/integrations/mason/README.md
+++ b/integrations/mason/README.md
@@ -29,20 +29,21 @@ pip install 'databricks-mason[tracing]'
 ## Authentication
 
 Mason uses [Databricks authentication](https://docs.databricks.com/aws/en/dev-tools/cli/authentication).
-If you do not already have credentials, authenticate a named profile first. You can
-then ask Mason to validate and remember that profile:
+`mason login` is the only auth command you need: it authenticates the profile — running the
+browser sign-in for you if that profile has no credentials yet — and remembers it as the
+default so later commands can omit `-p`:
 
 ```sh
-databricks auth login --profile <profile>
-mason login --profile <profile>
+mason login --profile <profile> --host https://<workspace>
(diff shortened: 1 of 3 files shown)


akansal2:
Superseded by #520 ("Make Mason login configure Databricks authentication"), which already landed the same `mason login` authenticates-you feature on `main` with a more complete flow (interactive-terminal check + validate→login→revalidate retry). Closing as a duplicate. The only unique piece here was `--host` support for creating a brand-new profile during login; can be added as a small standalone follow-up if wanted.

PR #528, opened 2026-09-03
[mason] Retry transient API errors and stop rendering `None` on message-less failures

annzhang-db:
`mason dev -s <store>` (and other agents/v1 commands) could fail with an unhelpful `Error [CANCELLED]: None` that only succeeded on a second try. Two independent causes:

- The backend returned a transient `CANCELLED` with no message. The databricks SDK stringifies a message-less error as the literal `"None"`, and our fallback treated that non-empty string as a real message and printed it verbatim.
- Nothing retried the transient failure — the SDK only retries connection/timeout errors, not these error-code responses.

This PR:
- Retries transient gRPC-style codes (`CANCELLED`, `UNAVAILABLE`, `DEADLINE_EXCEEDED`, `ABORTED`) once for requests that are safe to replay: reads, read-only search, and exclusive store creation. Stateful operations such as session-item append and pop are never retried because the first attempt may have committed even when its response was lost. Duplicate store creation surfaces `ALREADY_EXISTS`, which the ensure-store callers already reconcile as success.
- Stops printing the literal `None`: falls back to the error code and adds a "usually transient — re-run" hint for these codes.

Found during the Custom Agents bug bash.

### DIFF
--- a/integrations/mason/src/databricks_mason/client.py
+++ b/integrations/mason/src/databricks_mason/client.py
@@ -25,10 +25,9 @@ from databricks_mason.errors import TRANSIENT_ERROR_CODES, AgentCliError, wrap_a
 _BASE = "/api/agents/v1"
 _MCP_SERVICES_PATH = "/api/2.1/unity-catalog/mcp-services"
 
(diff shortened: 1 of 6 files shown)


elainewang-db (approved):
thanks!

PR #540, opened 2026-09-03
[mason] deploy/dev read agent.toml store bindings

elainewang-db:
###  What

  Make `agent.toml` the single source of truth for an agent's memory and session stores, end to end.

  - Stores are declared only via `mason memory bind` / `mason sessions bind`. Binding creates the store (idempotent) and records it in `agent.toml` — memory also records the store id, which the entries API is keyed by.
  - The runtime resolves stores straight from agent.toml (`resolve_memory_store` / `resolve_session_store`); precedence is explicit arg → `AGENT_*_STORE` env → `agent.toml` binding.
  - `mason deploy / mason dev` drop the --memory/-m and --session/-s flags.
  - Stores are no longer written into `app.yaml` env — the runtime reads `agent.toml` at request time, so there's nothing to sync.
  - `mason deploy / mason dev` now only validate the bound stores exist (erroring toward bind if not) and, for deploy, grant the app's service principal access. Store creation lives in bind alone; `--no-create-stores` is removed from deploy/dev (it stays on bind).
  - Adds a spinner over deploy/dev's previously silent phases (app-compute wait, store checks, access grant).

### Why

`mason memory/sessions bind` wrote the store into agent.toml, but `mason deploy` keyed both env-wiring and the SP access grant off the -s/-m flags — bool(`session_store` or `memory_store`). So bind + deploy (no flags) deployed an app whose service principal was never granted store access, and every runtime store call failed with `PERMISSION_DENIED`. The app was effectively broken unless you re-passed the store as a flag on deploy, duplicating what bind already recorded.

Routing everything through agent.toml fixes that and removes the `app.yaml`-vs-`agent.toml` split-brain: one place declares stores (bind), one place reads them (the runtime), and deploy just validates + grants.

### Tests

277 unit tests pass; ruff + ty clean. Coverage added/updated for:
- `mason memory bind` records the store id; resolve_memory_store returns the id (falling back to the display name for a hand-written binding).
- `store_binding(prefer="id")` and the resolver precedence.
- `mason deploy / mason dev` write no store env into app.yaml.
- `mason deploy errors` (pointing at bind) when a bound store doesn't exist.

### DIFF
--- a/integrations/mason/README.md
+++ b/integrations/mason/README.md
@@ -86,12 +86,15 @@ mason [-p <profile>] [-o text|json]
   init         [--framework openai|langgraph] [--disable-chat-app]
                [--profile P] [--repo URL] [--ref REF] [directory]
   dev          [--source PATH] [--prepare-environment] [--app-port PORT]
-               [--memory/-m N] [--session/-s N]
-               [--with-traces C.S] [--no-create-stores]
+               [--with-traces C.S]
   memory
+    bind         STORE [--source PATH] [--no-create-stores]
+    unbind       [--source PATH]
     stores     create | list | get | update | delete
     entries    create | get | list | search | update | delete
   sessions     create | list | get | update | delete | fork
+    bind         STORE [--source PATH] [--no-create-stores]
+    unbind       [--source PATH]
     stores     create | list | get | update | delete
     items      list | append | pop | clear
   tracing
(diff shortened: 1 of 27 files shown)


elainewang-db (commented):
@qinxinw-db do you think we can just remove mason dev/deploy -m -s now that we have `bind`?

qinxinw-db (approved):
thanks this makes sense!

PR #541, opened 2026-09-03
[mason] Fix profile auth when profiles share a host

annzhang-db:

Raise the Mason minimum `databricks-sdk` version from 0.49 to 0.94.0, the first release that forwards a configured profile to `databricks auth token` instead of resolving credentials only by host. This fixes `mason --profile ...` failing when multiple `.databrickscfg` profiles share the same workspace host.

Add a regression test with two same-host profiles that verifies the selected profile reaches the Databricks CLI without a `--host` fallback.

## How do you know it works?

- Lowest-direct/all-extras Python 3.10 suite with `databricks-sdk==0.94.0`: 271 passed
- Highest/all-extras Python 3.10 suite with `databricks-sdk==0.134.0`: 271 passed
- Targeted duplicate-host profile regression: passed
- Ruff lint and format checks passed in both dependency environments
- Highest/all-extras `ty check` passed
- Live read-only smoke test with same-host profiles: the selected profile authenticated and `mcp list` succeeded

### Before / after terminal behavior

Before (`databricks-sdk==0.93.0`, allowed by the old dependency floor):

```console
$ mason --profile <selected-profile> deploy <app-name> --session <session-store>
Error: Could not initialize Databricks auth: default auth: databricks-cli: cannot get access token: Error:
<selected-profile> and <other-profile> match <workspace-host> in
~/.databrickscfg. Use --profile to specify which profile to use. Config:
host=<workspace-host>, account_id=<account-id>, profile=<selected-profile>,
auth_type=databricks-cli
Select an existing profile with `mason --profile <name> <command>` or authenticate and save it with
`mason login --profile <name>`.
```

Mason accepts the outer `--profile`, but the older SDK drops it from the nested credential command:

```console
$ databricks auth token --host <workspace-host>
```

After (`databricks-sdk==0.94.0`, the new minimum), the selected profile reaches the nested CLI and the ambiguity is removed:

```console
$ databricks auth token --profile <selected-profile>
$ mason --profile <selected-profile> --output json mcp list >/dev/null && echo "Authenticated and listed MCP services."
Authenticated and listed MCP services.
```

### DIFF
--- a/integrations/mason/pyproject.toml
+++ b/integrations/mason/pyproject.toml
@@ -9,7 +9,7 @@ readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
     "click>=8.1",
-    "databricks-sdk>=0.49",
+    "databricks-sdk>=0.94.0",
     "psycopg[binary]>=3.1",
     "PyYAML>=6.0",
     "rich>=13.7",
--- a/integrations/mason/tests/unit_tests/client_test.py
+++ b/integrations/mason/tests/unit_tests/client_test.py
@@ -2,6 +2,9 @@
 
 from __future__ import annotations
 
+import pathlib
+import subprocess
+import types
 from unittest import mock
 
 import pytest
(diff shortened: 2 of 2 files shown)


akansal2 (approved):
thanks for making the fix! ran into this issue a bunch myself

PR #543, opened 2026-09-04
mason: show a spinner during deploy's silent phases

elainewang-db:
mason deploy ran silently through store provisioning, the wait for agent compute, and the store-access grants — the friction-log "runs ~15s without any feedback" gaps. Wrap each in the existing render.status spinner (skipped under -o json so machine output stays clean). The sync / apps deploy steps still stream the CLI's own output as before.


elainewang-db:
superseded by https://github.com/databricks/databricks-ai-bridge/pull/540?timeline_per_page=5

PR #547, opened 2026-09-04
[mason] Add a model picker to the local chat UI (Unity Catalog AI Gateway)

ankit-db:
Adds a model picker to the local chat UI (both templates) so you can switch models without editing code and restarting.

- Lists the workspace's Unity Catalog AI Gateway chat models (`system.ai.*`) and routes the agent through the gateway (`use_ai_gateway=True`).
- Default is `system.ai.claude-opus-4-8`; pick any listed model per turn from the chat header.


integrations/mason/templates/agent-langgraph/agent/agent.py
```diff
@@ -24,7 +24,11 @@
 
 logger = logging.getLogger(__name__)
 
-MODEL = "databricks-gpt-5-2"
+# A Unity Catalog AI Gateway model, served from the `system.ai` schema and queried through the
+# gateway (see `use_ai_gateway=True` below). Swap for any `system.ai.*` model your workspace exposes
+# — the demo chat app's picker lists what's available. `mason dev`'s chat UI can also override this
+# per request.
+MODEL = "system.ai.claude-opus-4-8"
```
elainewang-db, line 32: When I tested claude with langgraph, I saw some strange interaction with claude's reasoning and the UI rendering, with errors like 
```
 PydanticSerializationUnexpectedValue(Expected `str` - serialized value may not be as expected [field_name='content', input_value=[{'type': 'reasoning', 's... '', 'signature': ''}]}], input_type=list])
    return self.__pydantic_serializer__.to_python(
```
and some rendering issues in the chat UI too.

Did you end up seeing this too?

elainewang-db: Screenshot 2026-09-03 at 2 13 35 PM - like this

integrations/mason/src/databricks_mason/runtime/models.py
```diff
@@ -0,0 +1,71 @@
...
+    ``supported_api_types`` looks like ``["openai/v1/chat/completions"]``. Be lenient when it's
+    absent (offer the model), but drop services that only advertise embeddings.
+    """
+    types = [str(t).lower() for t in (supported_api_types or [])]
+    if not types:
+        return True
+    return any(("chat" in t or "completions" in t or "responses" in t) for t in types)
+
+
+def list_ai_gateway_models(client: WorkspaceClient) -> list[str]:
```
elainewang-db, line 58: should we retry transient failures?

elainewang-db:
whoops my PR https://github.com/databricks/databricks-ai-bridge/pull/538?timeline_per_page=5 got merged

integrations/mason/src/databricks_mason/runtime/models.py
ankit-db: Good call — added transient-failure retry. `list_ai_gateway_models` now retries the model-services list a few times (`_LIST_ATTEMPTS`, 0.5s apart) before surfacing the error, and `_discover_chat_models` still falls back to just the default if it ultimately fails. Covered by a new retry test in `runtime_models_test.py`.

integrations/mason/templates/agent-langgraph/agent/agent.py
ankit-db: Thanks for flagging — this is directly relevant now: after merging main + switching to the UC AI Gateway, the default `MODEL` is `system.ai.claude-opus-4-8`, so Claude is what a fresh scaffold hits first. That `PydanticSerializationUnexpectedValue` comes from Claude returning `content` as a list of blocks (incl. `reasoning`), which `_serialize_events`' `msg.model_dump()` and the UI's text extraction don't special-case. I'd rather fix it than ship a Claude default that renders oddly: I'm planning a focused follow-up that strips/handles reasoning blocks in `_serialize_events` (and the UI's `extractText`) so only user-visible text renders. Did your testing show a clean way to keep the reasoning out of the rendered turn, or just drop those blocks? Leaving this thread open until that's in.

ankit-db: ^this was my agent...

ankit-db: Fixed in 32ee242. `_serialize_events` now strips `reasoning`/`thinking` content blocks from both completed messages and streamed deltas (a reasoning-only delta is skipped), and suppresses that benign `PydanticSerializationUnexpectedValue` warning around `model_dump`. The message object itself is left untouched so Claude's reasoning continuity across turns still works. Added hermetic tests in `test_agent.py` for both the completed-message and delta paths.

ankit-db: Update (ac776bd): reworked this — instead of dropping the reasoning, it's now shown as a collapsible **"Thinking"** element above the answer (dimmed, streams live, then collapses), kept out of the answer bubble. Backend splits reasoning into its own `{"type":"reasoning"}` stream events + a `message["reasoning"]` field (so sync/background transports keep it too). Felt wrong to throw the chain-of-thought away entirely. Tests updated for the split; the answer bubble stays clean and the pydantic warning is still suppressed.

ankit-db: Fixed.

ankit-db:
Elaine, fixed this - requested again

elainewang-db (approved):

PR #548, opened 2026-09-04
[mason] Bug-bash fixes: deploy store-created messaging + AGENTS.md wording

ankit-db:
Two smaller fixes reported in the Custom Agents bug bash.

**`mason deploy` now clearly logs when it creates a store.** As each store is provisioned it prints a standalone line — `✓ Created new memory store '<name>'` — and the final summary marks each store `(created)` or `(existing)`. Previously the summary listed the store but never said whether it was newly created or reused. (`stores_created` is also in the `--output json` payload.)

**Scaffolded `AGENTS.md` no longer frames the agent as a "Databricks Apps" backend** (both templates). The intro now reads "…agent backend served from a FastAPI app" instead of "…for Databricks Apps".

(The third reported item — `mason login`'s next steps — was already fixed on main, so it's not included here.)

Tested: `ruff` / `ty` clean; 307 CLI unit tests pass (incl. new coverage for the created-vs-reused logging).

### DIFF
--- a/integrations/mason/README.md
+++ b/integrations/mason/README.md
@@ -115,12 +115,15 @@ mason [-p <profile>] [-o text|json]
   init         [--framework openai|langgraph] [--disable-chat-app]
                [--profile P] [--repo URL] [--ref REF] [directory]
   dev          [--source PATH] [--prepare-environment] [--app-port PORT]
-               [--memory/-m N] [--session/-s N]
-               [--with-traces C.S] [--no-create-stores]
+               [--with-traces C.S]
   memory
+    bind         STORE [--source PATH] [--no-create-stores]
+    unbind       [--source PATH]
     stores     create | list | get | update | delete
     entries    create | get | list | search | update | delete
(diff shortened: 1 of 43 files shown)


elainewang-db (approved):
thanks! surfacing one related decision, we're thinking to remove `mason dev/deploy -m .. -s ..` entirely in favor of `mason sessions/memory bind ..`. then `mason dev/deploy` should never take any memory or session store flags and always just read from the `agent.toml` file that `bind` writes to. wdyt?

ankit-db:
> thanks! surfacing one related decision, we're thinking to remove `mason dev/deploy -m .. -s ..` entirely in favor of `mason sessions/memory bind ..`. then `mason dev/deploy` should never take any memory or session store flags and always just read from the `agent.toml` file that `bind` writes to. wdyt?

It's not a bad idea yeah

ankit-db:
Merging anyways here and we can fix that forward

PR #552, opened 2026-09-04
[mason] Show progress during deploy's silent phases

elainewang-db:
`mason deploy` runs for minutes with no output during its slow phases — provisioning
compute and waiting for it to become ACTIVE — so it looks hung. This adds durable
progress feedback around those phases.

- New `render.progress(message)`: prints a **persistent** `• message` line, then runs a bare spinner beneath it. Unlike `render.status` (which clears itself and only animates on a TTY), the line survives — so there's visible feedback even in terminals where the spinner doesn't render. Skipped under `-o json`.
- Wrap the two silent, minutes-long phases in it:
  - `apps create` (it provisions + waits for compute; its output is captured/relabeled, so nothing streamed before) → "Creating the agent and starting its compute…"
  - the compute-ACTIVE wait → "Waiting for agent compute to start…"
- Run the compute wait on **every** deploy, not just the first (an existing app may be STOPPED/starting). `apps create` stays gated to new apps (it errors on an existing one).
- Fix a rendering bug: the deployed-app URL was shown as a `$ open <url>` command; it's now prose (`Open the deployed app: <url>`).
- Say "agent" not "app" in the progress messages, matching Mason's agent branding.

## Why

The real gap was that `apps create` blocks for minutes with `capture=True` (so its output is buffered, not streamed) before any feedback line was reached — the CLI looked frozen. `render.status` alone wasn't enough: it clears on exit and doesn't animate on every terminal, so a long wait could leave nothing on screen. The persistent line guarantees feedback regardless.

## Tests

318 unit tests pass; ruff + ty clean. Added coverage for `render.progress` (prints the persistent line in text mode, silent under json) and for the compute wait running on both first-deploy and re-deploy.

Screenshot 2026-09-04 at 4 16 37 PM

### DIFF
--- a/integrations/mason/src/databricks_mason/deploy.py
+++ b/integrations/mason/src/databricks_mason/deploy.py
@@ -438,7 +438,7 @@ def deploy(
     #    in progress (persistent line + spinner) so the CLI isn't silent for the whole provision.
     if not _deployment_exists(name, obj.profile):
         with render.progress(
-            "Creating the app and starting its compute (this can take a few minutes)…"
+            "Creating the agent and starting its compute (this can take a few minutes)…"
         ):
             result = _databricks(
                 ["apps", "create", name, *instance_args],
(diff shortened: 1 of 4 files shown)


jamesbxwu (approved):

integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -515,7 +524,7 @@ def deploy(
         (f"mason deployments logs {name}", "Tail its logs"),
     ]
     if app_url:
-        steps.insert(0, (f"open {app_url}", "Open the deployed app"))
+        steps.insert(0, f"Open the deployed app: {app_url}")
```
junchoi-db: super nit: should we say `deployed agent`?

junchoi-db (approved):

PR #554, opened 2026-09-08
[mason] tracing: managed MLflow tracing, on by default, experiment_id-centric

jamesbxwu:
## Summary
- Make tracing **default on** for mason dev / mason deploy and writing to a default mlflow experiment derived from the current agent 
- Introduced `mason tracing configure` to bind to an existing experiment
- Remove support for UC based tracing due to its setup complexity (will address in a follow up)

## What changed

- **`mason tracing`**: `configure [--experiment <id>]` (on / rebind, id-only override), `disable` (off), `list`, `get`. Dropped `setup --catalog/--schema/--warehouse-id`, the UC `set_experiment_trace_location` linking, and `instrument`.
- **On by default**: `mason dev` and `mason deploy` send traces to a per-app experiment (`/Users/<user>/mason-traces/<app>`), auto-created on first run. No `--with-traces` flags.
- **One identifier**: the runtime binding is just `MLFLOW_TRACKING_URI=databricks` + `MLFLOW_EXPERIMENT_ID=<id>`.
- **`agent.toml`**: a single `[tracing]` table (`experiment_id` / `disabled`).
- **Deploy grant**: the experiment is declared as an `AppResourceExperiment(experiment_id, CAN_EDIT)` app resource so the app's service principal can write traces with no manual grant.
- Provisioning is best-effort: dev/deploy never block on tracing (no auth, no mlflow, or offline all degrade to running without traces).

## Managed-only for now (UC-backed rejected)

`mason` manages **managed** (non-UC) experiments. If `configure --experiment <id>` is pointed at a **UC-backed** experiment, it errors clearly (`UC-backed MLflow tracing is not supported by mason.`) rather than silently wiring a config that would fail at read/deploy - reading UC traces needs a SQL warehouse and a deployed app's SP needs UC grants mason doesn't set up. Supporting UC-backed experiments (require a warehouse + grant the app SP `USE_CATALOG`/`USE_SCHEMA`/`SELECT`/`MODIFY` on the schema) is deferred to a **follow-up**.

## Testing done

Unit: `401 passed`, `ty` clean, `ruff` clean.

Live (e2-dogfood, fresh env - installed the CLI, scaffolded, `dev`, `deploy`):
- **`mason dev`**: the per-app experiment was auto-created and the two MLflow env vars wired into `app.yaml`; invoking the local agent produced a trace that `mason tracing list` shows.
- **`mason deploy`**: the deploy reported `Trace access - granted to app service principal` (the experiment app-resource), and invoking the **deployed** app via OAuth produced a second trace in the same experiment - written by the app's service principal with **no manual grant**. Both traces (dev + deployed) appear in `mason tracing list`.
- **UC-backed guard**: created a UC-linked experiment and confirmed `configure --experiment <it>` errors cleanly; a managed experiment still configures fine.

Isaac Review found three MAJOR bugs that unit tests and the initial live run missed - dev/deploy keying the experiment on different names, a non-`AgentCliError` from provisioning aborting the deploy, and an offline `mason dev` regression - each now fixed and covered by a regression test.

## Note

This supersedes the UC-only tracing PR #545 (different branch); that approach is abandoned in favor of this simpler managed model.

### DIFF
--- a/integrations/mason/src/databricks_mason/agent_project.py
+++ b/integrations/mason/src/databricks_mason/agent_project.py
@@ -17,7 +17,7 @@ from tomlkit.exceptions import ParseError
 from databricks_mason.errors import AgentCliError
 from databricks_mason.runtime.tool_manifest import MEMORY_STORE_TABLE, SESSION_STORE_TABLE
 
-# The tracing binding (`mason tracing configure` / `disable`). Tracing is on by default (a per-app
+# The tracing binding (`mason tracing configure` / `disable`). Tracing is on by default (a per-project
 # MLflow experiment); this table only records an explicit experiment override or a disable.
 TRACING_TABLE = "tracing"
 
@@ -329,7 +329,7 @@ class AgentProject:
         # The deployment's base name (`mason deploy` prefixes it with `mason-`); None until named.
         self.deployment_name = deployment_name
         self.durability_enabled = durability_enabled
-        # Tracing config: an explicit experiment id override (None = default per-app experiment), and
+        # Tracing config: an explicit experiment id override (None = default per-project experiment), and
         # whether tracing is disabled (tracing is on by default; this flag turns it off).
         self.trace_experiment_id = trace_experiment_id
         self.trace_disabled = trace_disabled
@@ -505,7 +505,7 @@ class AgentProject:
     def configure_tracing(self, experiment_id: str | None) -> bool:
         """Enable tracing and (optionally) pin an explicit experiment id. Returns True if changed.
 
-        ``experiment_id=None`` means "use the default per-app experiment": any prior override is
+        ``experiment_id=None`` means "use the default per-project experiment": any prior override is
         cleared. Enabling always clears a previous ``disabled`` flag (tracing is on by default, so an
         absent ``[tracing]`` table is the enabled default).
         """
--- a/integrations/mason/src/databricks_mason/agent_project.py
+++ b/integrations/mason/src/databricks_mason/agent_project.py
@@ -17,6 +17,10 @@ from tomlkit.exceptions import ParseError
 from databricks_mason.errors import AgentCliError
 from databricks_mason.runtime.tool_manifest import MEMORY_STORE_TABLE, SESSION_STORE_TABLE
 
+# The tracing binding (`mason tracing configure` / `disable`). Tracing is on by default (a per-app
+# MLflow experiment); this table only records an explicit experiment override or a disable.
+TRACING_TABLE = "tracing"
+
 _SCHEMA_VERSION = 1
 _SUPPORTED_FRAMEWORKS = {"langgraph", "openai"}
 _SUPPORTED_SCOPE_KINDS = {"table", "volume", "workspace"}
@@ -296,6 +300,8 @@ class AgentProject:
         memory_store: str | None = None,
         session_store: str | None = None,
         memory_store_id: str | None = None,
+        trace_experiment_id: str | None = None,
+        trace_disabled: bool = False,
     ) -> None:
         self.root = root
         self.path = root / "agent.toml"
@@ -307,6 +313,10 @@ class AgentProject:
         self.memory_store = memory_store
         self.session_store = session_store
         self.memory_store_id = memory_store_id
+        # Tracing config: an explicit experiment id override (None = default per-app experiment), and
+        # whether tracing is disabled (tracing is on by default; this flag turns it off).
+        self.trace_experiment_id = trace_experiment_id
+        self.trace_disabled = trace_disabled
 
     @classmethod
     def load(cls, root: pathlib.Path | str) -> "AgentProject":
@@ -347,6 +357,15 @@ class AgentProject:
         session_store = _store_name_from_manifest(
             document.get(SESSION_STORE_TABLE), SESSION_STORE_TABLE
         )
+        tracing_table = document.get(TRACING_TABLE)
+        trace_experiment_id: str | None = None
+        trace_disabled = False
+        if isinstance(tracing_table, Mapping):
+            raw_experiment = tracing_table.get("experiment_id")
+            trace_experiment_id = (
+                str(raw_experiment) if isinstance(raw_experiment, str) and raw_experiment else None
+            )
+            trace_disabled = bool(tracing_table.get("disabled"))
         return cls(
             project_root,
             document,
@@ -355,6 +374,8 @@ class AgentProject:
             memory_store,
             session_store,
             memory_store_id,
+            trace_experiment_id,
+            trace_disabled,
         )
 
     @classmethod
@@ -428,6 +449,41 @@ class AgentProject:
     def unbind_session_store(self) -> bool:
         """Remove the session store binding from agent.toml. Returns True if it was present."""
         return self._clear_store(SESSION_STORE_TABLE)
+
+    def configure_tracing(self, experiment_id: str | None) -> bool:
+        """Enable tracing and (optionally) pin an explicit experiment id. Returns True if changed.
+
+        ``experiment_id=None`` means "use the default per-app experiment": any prior override is
+        cleared. Enabling always clears a previous ``disabled`` flag (tracing is on by default, so an
+        absent ``[tracing]`` table is the enabled default).
+        """
+        if self.trace_experiment_id == experiment_id and not self.trace_disabled:
+            return False
+        table = self._document.get(TRACING_TABLE)
+        if not isinstance(table, Mapping):
+            table = tomlkit.table()
+            self._document.append(TRACING_TABLE, table)
+        if "disabled" in table:
+            del table["disabled"]
+        if experiment_id:
+            table["experiment_id"] = experiment_id
+        elif "experiment_id" in table:
+            del table["experiment_id"]
+        self.trace_experiment_id = experiment_id
+        self.trace_disabled = False
+        return True
+
+    def disable_tracing(self) -> bool:
+        """Turn tracing off (records ``[tracing] disabled = true``). Returns True if changed."""
+        if self.trace_disabled:
+            return False
+        table = self._document.get(TRACING_TABLE)
+        if not isinstance(table, Mapping):
+            table = tomlkit.table()
+            self._document.append(TRACING_TABLE, table)
+        table["disabled"] = True
+        self.trace_disabled = True
+        return True
 
     def _set_store(self, table: str, name: str, store_id: str | None = None) -> bool:
         name = _required_string(name, f"[{table}] name")
--- a/integrations/mason/src/databricks_mason/deploy.py
+++ b/integrations/mason/src/databricks_mason/deploy.py
@@ -41,8 +41,8 @@ from databricks_mason.store_access import (
 from databricks_mason.tracing import (
     TRACES_EXPERIMENT_ID_ENV,
     TRACES_TRACKING_URI_ENV,
-    default_experiment,
-    ensure_experiment,
+    create_experiment_idempotent,
+    default_experiment_name,
     experiment_url,
 )
 
@@ -321,7 +321,7 @@ def validate_stores(client, *, memory_store: Optional[str], session_store: Optio
             ) from exc
 
 
-def resolve_trace_experiment(
+def resolve_trace_experiment_id(
     source: pathlib.Path, project_name: str, client, profile
 ) -> Optional[str]:
     """The MLflow experiment id an agent traces to, or None when tracing is disabled.
@@ -349,8 +349,8 @@ def resolve_trace_experiment(
     pinned = project.trace_experiment_id if project is not None else None
     if pinned:
         return pinned
-    experiment_id = ensure_experiment(
-        profile, client, default_experiment(client.current_user, project_name)
+    experiment_id = create_experiment_idempotent(
+        profile, client, default_experiment_name(client.current_user, project_name)
     )
     # Pin the resolved default so subsequent runs reuse it by id (removes the special-cased "recompute
     # the default" path). No agent.toml (raw dir) just means nowhere to pin — still trace this run.
@@ -496,7 +496,7 @@ def deploy(
     trace_experiment_id: Optional[str] = None
     trace_setup_error: Optional[str] = None
     try:
-        trace_experiment_id = resolve_trace_experiment(
+        trace_experiment_id = resolve_trace_experiment_id(
             source_dir, source_dir.resolve().name, client, obj.profile
         )
     except Exception as exc:  # noqa: BLE001 - tracing is best-effort; never block a deploy
--- a/integrations/mason/src/databricks_mason/deploy.py
+++ b/integrations/mason/src/databricks_mason/deploy.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 import json
 import pathlib
 import time
-from dataclasses import replace
+from dataclasses import dataclass, replace
 from typing import Any, Optional
 
 import click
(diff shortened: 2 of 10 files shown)


integrations/mason/src/databricks_mason/agent_project.py
```diff
@@ -309,6 +313,8 @@ def __init__(
         memory_store_id: str | None = None,
         deployment_name: str | None = None,
         durability_enabled: bool = False,
+        trace_experiment_id: str | None = None,
+        trace_disabled: bool = False,
```
jamesbxwu: can't we rely on `trace_experiment_id=None` to be the disablement signal?

integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -317,18 +319,33 @@ def validate_stores_and_trace_env(
...
+        project = None
+    if project is not None and project.trace_disabled:
+        return None
+    pinned = project.trace_experiment_id if project is not None else None
+    if pinned:
+        return pinned
+    return ensure_experiment(profile, client, default_experiment(client.current_user, app))
+
+
+def trace_env(experiment_id: str) -> dict[str, str]:
```
jamesbxwu:
```suggestion
def mlflow_tracing_config(experiment_id: str) -> dict[str, str]:
```

jamesbxwu: can we create a dedicated class instead of dict[str,str]?

integrations/mason/src/databricks_mason/tracing.py
```diff
@@ -22,34 +24,31 @@
...
+# Installing the `tracing` extra (rather than a bare mlflow) resolves both the missing- and
+# too-old-mlflow cases: the extra carries the version floor `mason tracing` needs.
+_INSTALL_HINT = "Install the tracing extra: pip install 'databricks-mason[tracing]'"
 
 
-# Env vars the deployed agent reads (see deploy.py). MLFLOW_TRACING_DESTINATION is MLflow's own
-# "catalog.schema" convention; MLFLOW_EXPERIMENT_NAME is the standard MLflow experiment selector.
-TRACES_DEST_ENV = "MLFLOW_TRACING_DESTINATION"
-TRACES_EXPERIMENT_ENV = "MLFLOW_EXPERIMENT_NAME"
+def default_experiment(user: str, app: Optional[str]) -> str:
```
jamesbxwu:
```suggestion
def get_default_experiment_path(user: str, app: Optional[str]) -> str:
```

integrations/mason/src/databricks_mason/tracing.py
```diff
@@ -22,34 +24,31 @@
...
+    if not app:
+        raise AgentCliError("Cannot derive the default tracing experiment without an app name.")
+    return f"/Users/{user}/{_TRACES_DIR}/{app}"
 
 
-# Installing the `tracing` extra (rather than a bare mlflow) is what actually resolves both the
-# missing- and too-old-mlflow cases: the extra carries the version floor `mason tracing` needs, so
-# a stale mlflow already in the venv gets upgraded to a compatible one.
-_INSTALL_HINT = "Install the tracing extra: pip install 'databricks-mason[tracing]'"
+def experiment_url(host: Optional[str], experiment_id: str) -> Optional[str]:
```
jamesbxwu:
```suggestion
def get_experiment_url(host: Optional[str], experiment_id: str) -> Optional[str]:
```

integrations/mason/src/databricks_mason/tracing.py
```diff
@@ -145,131 +136,144 @@ def _status_str(status: Any) -> Optional[str]:
...
+
+    if experiment_id:
+        # Verify the experiment exists so a wrong id fails here, not silently when the agent runs.
+        mlflow = _mlflow()
+        _configure(mlflow, obj.profile)
+        experiment = mlflow.get_experiment(experiment_id)
+        if experiment is None:
+            raise AgentCliError(
+                f"No MLflow experiment found with id {experiment_id!r}.",
+                hint="Pass an existing experiment id, or omit --experiment for the per-app default.",
```
jamesbxwu:
```suggestion
                hint="Pass an existing experiment id, or omit --experiment to use the mason default.",
```

integrations/mason/src/databricks_mason/tracing.py
```diff
@@ -145,131 +136,144 @@ def _status_str(status: Any) -> Optional[str]:
...
+            )
+        if _is_uc_backed(experiment):
+            # Traces for a UC-backed experiment land in governed UC tables, which need a SQL warehouse
+            # to read and UC grants for a deployed app's SP to write - neither of which mason sets up
+            # yet. Reject it up front rather than silently wiring a config that fails at read/deploy.
+            raise AgentCliError(
+                "UC-backed MLflow tracing is not supported by mason.",
+                hint="Pass a managed (non-UC) experiment, or omit --experiment to use the per-app "
+                "experiment mason creates.",
```
jamesbxwu:
```suggestion
                hint="Pass a managed (non-UC) experiment, or omit --experiment to use the mason default.",
```

integrations/mason/src/databricks_mason/dev.py
```diff
@@ -77,10 +66,11 @@ def dev(
     ``mason deploy``. The environment is built on first run and reused after; pass
     ``--prepare-environment`` to force a rebuild (e.g. after changing dependencies).
 
-    Stores bound with ``mason memory/sessions bind`` are validated here and read from agent.toml at
-    runtime (not written to app.yaml); the ``--with-traces`` flag wires tracing env into app.yaml,
-    exactly as ``mason deploy`` does. Locally the store owner (you) already has access, so no
-    service-principal grant is needed here; that grant happens at ``mason deploy`` time.
+    Tracing is on by default: dev sends the agent's traces to a per-app MLflow experiment (the same
```
jamesbxwu:
```suggestion
    Tracing is on by default: dev sends the agent's traces to the default mason experiment based on the agent name (the same
```

integrations/mason/src/databricks_mason/dev.py
```diff
@@ -90,24 +80,29 @@ def dev(
...
+    # offline, no mlflow, permission) degrades to running without traces rather than aborting a purely
+    # local run.
+    try:
+        experiment_id = resolve_trace_experiment(
+            source_dir, source_dir.resolve().name, obj.client(), obj.profile
+        )
+        if experiment_id:
+            env_updates.update(trace_env(experiment_id))
+    except Exception as exc:  # noqa: BLE001 - tracing must never block a local run
+        render.console().print(f"[yellow]⚠[/] Tracing not enabled: {exc}. Running without traces.")
```
jamesbxwu:
```suggestion
        render.console().print(f"[yellow]⚠[/] Tracing not enabled: {exc}. Proceeding without tracing.")
```

integrations/mason/src/databricks_mason/tracing.py
```diff
@@ -145,131 +136,144 @@ def _status_str(status: Any) -> Optional[str]:
...
-    catalog, _, schema = destination.partition(".")
-    if not catalog or not schema:
-        raise AgentCliError("--destination must be 'catalog.schema'.")
-    return catalog, schema
-
-
-def _experiment_url(host: str, experiment_id: str) -> str:
-    """Workspace URL for an experiment's Traces tab."""
-    return f"{host.rstrip('/')}/ml/experiments/{experiment_id}?compareRunsMode=TRACES"
+def _trace_json(trace: Any) -> dict:
```
jamesbxwu:
```suggestion
def _trace_to_json(trace: Any) -> dict:
```

integrations/mason/src/databricks_mason/agent_project.py
jamesbxwu, line 317: is it possible to derive `trace_disabled` from `trace_experiment_id == None`?

jamesbxwu: Not cleanly - `trace_experiment_id` and `disabled` are orthogonal, so folding one into the other loses a state:

- `experiment_id` answers *which* experiment. `None` = "use the per-app default", not "off".
- `disabled` answers *on/off*. Tracing is on by default, so a fresh/absent `[tracing]` (`experiment_id == None`) must mean **on-with-default**. If `None` also meant disabled, we could never express on-by-default. And clearing `experiment_id` to turn tracing off would not stick - the next `dev`/`deploy` just re-derives the default and turns it back on.

That said, your underlying point (the special-cased "recompute the default each run" was weird) is fixed in 107ec39: `resolve_trace_experiment` now pins the resolved default id into `agent.toml` on the first `dev`/`deploy`, so later runs reuse it by id. After first run there is no separate "default" state - just a pinned id or `disabled = true`.

integrations/mason/src/databricks_mason/deploy.py
jamesbxwu: Done in 9c49d6b - added a frozen `MlflowTracingConfig` dataclass (`experiment_id` + `tracking_uri`, with an `.env()` method) and a `mlflow_tracing_config(...)` factory, replacing the bare `dict[str, str]`.

jamesbxwu: Superseded by the dedicated-class request on this line - went with a `MlflowTracingConfig` dataclass instead of typing it as `dict[str, str]` (9c49d6b).

integrations/mason/src/databricks_mason/tracing.py
jamesbxwu: Applied in 9c49d6b.

integrations/mason/src/databricks_mason/tracing.py
jamesbxwu: Applied in 9c49d6b.

integrations/mason/src/databricks_mason/dev.py
jamesbxwu: Applied in 9c49d6b (and lightly extended in 107ec39 to note the default experiment is now pinned into agent.toml on first run).

integrations/mason/src/databricks_mason/dev.py
jamesbxwu: Applied in 9c49d6b.

integrations/mason/src/databricks_mason/tracing.py
jamesbxwu: Applied in 9c49d6b - renamed to `_trace_to_json`.

integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -286,21 +297,12 @@ def _resolve_deployment_name(project, name: Optional[str]) -> str:
...
-    traces_experiment: Optional[str],
-) -> dict[str, str]:
-    """Validate the agent's bound stores exist and build the MLFLOW_* trace env to wire in.
-
-    Shared by `mason deploy` and `mason dev`. Stores are created by `mason memory/sessions bind` and
-    read from agent.toml at runtime, so this neither creates them nor writes them to app.yaml — it
-    only checks a bound store still exists (a typo or unbound clone fails here, not at runtime) and
-    returns the trace env.
+def validate_stores(client, *, memory_store: Optional[str], session_store: Optional[str]) -> None:
+    """Validate the agent's bound stores exist. Shared by `mason deploy` and `mason dev`.
```
elainewang-db, line 301: follow up: move this and _upsert_manifest_env to some shared util file and `resolve_trace_experiment` to mason.tracing?

integrations/mason/src/databricks_mason/tracing.py
```diff
@@ -22,34 +26,35 @@
...
-    Keeps each agent's traces in their own experiment (and out of other users' view), instead of the
-    single shared bucket. `mason tracing setup`, `dev`, and `deploy` all derive the same path for a
-    given agent, so the experiment `setup` links is the one the agent logs to. Falls back to the
-    shared default only when no app name is available.
-    """
-    if not app:
-        return _DEFAULT_EXPERIMENT
-    return f"/Users/{user}/mason-traces/{app}"
 
+def default_experiment(user: str, project: Optional[str]) -> str:
```
elainewang-db: can this be used in `agent_project.py` as the default experiment?

integrations/mason/src/databricks_mason/agent_project.py
elainewang-db: can we create the toml with the `trace_experiment_id` set to the default experiment path always? and dev and deploy never overwrite `trace_experiment_id` in `agent.toml`? then we can remove `disabled`, if `trace_experiment_id` is unset then the user must have disabled it

integrations/mason/src/databricks_mason/tracing.py
jamesbxwu:
```suggestion
def default_experiment_name(user: str, project: Optional[str]) -> str:
```

integrations/mason/src/databricks_mason/tracing.py
```diff
@@ -60,72 +65,64 @@ def _mlflow():
...
+    """True if the experiment stores traces in Unity Catalog rather than the managed MLflow backend."""
+    return _UC_TRACE_TAG in (getattr(experiment, "tags", None) or {})
 
-def _configure(mlflow, profile: Optional[str], warehouse_id: Optional[str]) -> None:
-    """Point MLflow at the workspace (honoring mason's --profile) for UC-backed tracing."""
-    mlflow.set_tracking_uri(f"databricks://{profile}" if profile else "databricks")
-    if warehouse_id:
-        os.environ["MLFLOW_TRACING_SQL_WAREHOUSE_ID"] = warehouse_id
 
+def ensure_experiment(profile: Optional[str], client, name: str) -> str:
```
jamesbxwu:
```suggestion
def create_experiment_idempotent(profile: Optional[str], client, name: str) -> str:
```

integrations/mason/src/databricks_mason/agent_project.py
jamesbxwu: the question here is _when_ to create the default experiment and get back the id to write to the toml file.  While we can derive the experiment name (path) from the mason project name, we cannot get the id until we actually create the experiment which is done in mason dev / deploy and not mason init.

integrations/mason/src/databricks_mason/deploy.py
jamesbxwu: will do as a follow up

integrations/mason/src/databricks_mason/tracing.py
jamesbxwu: see above comment that we store the experiment_id in the project but this default is the experiment_name (that is derived from the mason project name)

integrations/mason/src/databricks_mason/agent_project.py
elainewang-db: ah I see... makes sense. What about setting the experiment name in toml file? as the agent author it would be nice to see the experiment name rather than the id in my configuration file.

elainewang-db: also i think similar motivation to your first comment, having both `trace_disabled` and `trace_experiment_id` is a little confusing to me.

elainewang-db (approved):

integrations/mason/src/databricks_mason/agent_project.py
jamesbxwu: > What about setting the experiment name in toml file

I chose id since it's the canonical identifier as mlflow experiment names can be changed.  

>  having both trace_disabled and trace_experiment_id is a little confusing to me

Good call out. I went back and forth on this and open to ideas.  The reason I kept both is I wanted to provide a knob to disable tracing even if it was bound to an experiment previously.

PR #558, opened 2026-09-08
mason: move CLI modules into a databricks_mason.cli subpackage

elainewang-db:
The CLI command modules sat flat at the package top mixed with the SDK and shared infra. Collect them under databricks_mason/cli/ (auth, deploy, dev, help, init, mcp, memory, sessions, tools, tracing, sandbox, render, agent_project, plus the root command tree as cli/app.py). cli/__init__.py re-exports `main`, so the console entrypoint `databricks_mason.cli:main` is unchanged. The old cli.py became cli/app.py — named `app` (not `main`) so the entry function doesn't shadow the submodule. errors.py and the SDK/runtime packages stay put; runtime never imports from cli.


jamesbxwu (approved):
LGTM, to avoid merge conflicts with ongoing work, would be good to do this right before we release

PR #576, opened 2026-09-10
Fix Mason UI runtime capability reporting

shivam5:
## Summary

- rename the UI configuration endpoint from `/api/demo/config` to `/api/ui/config`
- report the effective run store for both streaming and background execution
- show `In-process run store` for local/in-memory execution and `Durable run store` for Lakebase
- prevent stale UI assets after redeploy and replace indefinite `Checking…` states with `Unavailable` on configuration errors

## Validation

- `ruff check` and formatting passed
- Mason unit and functional suite: `462 passed`
- verified custom-server templates do not include the Mason chat UI or capability endpoint


shivam5:
Done in https://github.com/databricks/databricks-ai-bridge/pull/592

PR #577, opened 2026-09-10
mason: grant store access via the conversation-store permissions API

elainewang-db:
Replace the direct Lakebase grant (postgres app-resource attach + owner-issued SQL GRANT) with the conversation-store permissions:grant RPC for session and memory stores. The store service performs the Lakebase role provisioning and GRANTs server-side, so deploy no longer requires the caller to own the store or hold MANAGE on its (service-managed) Lakebase project — fixing store grants for non-admin deployers. Durability keeps its dedicated app-owned Lakebase resource.

Removes the now-unused owner-grant machinery (grant_tables, _mint_token, _resolve_pg_host, _memory_store_database).

see https://mason-grant-test-del-6051921418418893.staging.aws.databricksapps.com


integrations/mason/src/databricks_mason/_api_client.py
```diff
@@ -405,6 +405,40 @@ def update_session_store(
     def delete_session_store(self, name: str) -> dict:
         return self._do("DELETE", f"{_BASE}/{session_store_path(name)}")
 
+    # --- store permission grants --------------------------------------------
+
+    def grant_session_store_permission(
```
elainewang-db, line 410: @shaotong-db btw do we have `WorkspaceClient.mason` support that can replace this whole file?

integrations/mason/src/databricks_mason/app_resources.py
```diff
@@ -0,0 +1,125 @@
+"""Lakebase/Apps resource plumbing for a deployed app.
```
elainewang-db, line 1: this is `store_access.py` moved and renamed

integrations/mason/src/databricks_mason/_api_client.py
elainewang-db, line 22: should this be v2? @shaotong-db

qinxinw-db: the new standardized api path is `/2.0/agents/`

integrations/mason/src/databricks_mason/_api_client.py
shaotong-db: Nope didn't get a chance to add it yet. Do you wanna add something to `integrations/mason/src/databricks_mason/session_store.py` and `integrations/mason/src/databricks_mason/memory_store.py` in this PR? I can also do that a bit later.

qinxinw-db (approved):

integrations/mason/src/databricks_mason/_api_client.py
shaotong-db: Initially we used the API group style for this, but then the deco team advised us to use `/api/2.0/agents` since the API group is not mature yet. I believe both API paths support all RPCs now.

@annzhang-db Are we good to switch to `/api/2.0/agents` given you have the best context here?

shaotong-db (approved):
Thanks!

integrations/mason/src/databricks_mason/_api_client.py
shaotong-db: Actually I'm not sure. Is `WorkspaceClient.mason` generated by databricks SDK from our api proto? If so then the new api might be already included.

shaotong-db: Had Claude doing a quick investigation. Seems the databricks SDK is not ready yet so we have to call the raw REST API.

It would be good to add the permission grant API in integrations/mason/src/databricks_mason/session_store.py and integrations/mason/src/databricks_mason/memory_store.py so it's easier to switch in the future.

elainewang-db: done, also will do a follow up pr to have `mason deploy` etc use the public `session_store.py` / `memory_store.py` potentially

integrations/mason/src/databricks_mason/_api_client.py
elainewang-db: ok will do a follow to flip to `/api/2.0/agents` everywhere in mason, not just here.

PR #550, opened 2026-09-04
Add configurable Mason server and durable runtime modes

shivam5:
## What changed

- Adds `databricks_mason.AgentApp`, the Mason HTTP server for foreground, streaming, and background invocations. `AgentApp()` uses process-local state; `AgentApp(durable_runtime=True)` adds Lakebase persistence, heartbeats, and crash recovery.
- Simplifies `mason init` to three user-facing modes for both LangGraph and OpenAI:

| Command | Server | Runtime state |
| --- | --- | --- |
| `mason init my-agent` | Mason `AgentApp` | Durable Lakebase runtime after deploy (default) |
| `mason init my-agent --no-durable-runtime` | Mason `AgentApp` | Process-local background/event state |
| `mason init my-agent --server custom` | Plain FastAPI | Foreground-only; no Mason `AgentApp` |

- Removes the previous automatic-recovery CLI/configuration knob. Recovery is a capability of the durable runtime: the generated app registers `@app.on_recovery` only when `DURABLE_RUNTIME` is enabled.
- Adds minimal custom FastAPI templates for LangGraph and OpenAI while retaining the full existing agents, tools, HITL, sessions, memory, and chat UI in the Mason templates.
- Reuses `databricks_mason.workspace_client()` and `workspace_headers()` in custom templates for account-host workspace routing without using `AgentApp`.
- Provisions/reuses a Lakebase durability store only when generated `agent.toml` contains `[durability] enabled = true`. The process-local and custom modes do not provision it.
- Keeps `runtime/main.py` thin: configure the agent, construct the app, register hooks, and install the optional UI.

## Validation

Validated through commit `fe86e628e6f3297b056504674c458a964af9afea`.

- Mason package: `400 passed`
- `ruff format`, `ruff check`, `ty check`, `uv build`, and `git diff --check`: passed
- Generated and synced all six framework/mode combinations using the installed `mason` command
- Every generated project passed its own tests and package build
- Every generated project started with `mason dev` and completed a real model invocation using explicit profile `e2-dogfood`

| Framework | Mode | Generated tests | Live local invocation |
| --- | --- | --- | --- |
| LangGraph | Durable Mason | `26 passed, 1 skipped` | HTTP 200, `langgraph durable ok` |
| LangGraph | Process-local Mason | `26 passed, 1 skipped` | HTTP 200, `langgraph-local ok` |
| LangGraph | Custom FastAPI | `2 passed` | HTTP 200, `langgraph-custom ok` |
[...]

PR #578, opened 2026-09-10
[mason] tracing: fix default tracing in mason template 

jamesbxwu:
## Summary

Deployed and `mason dev` agents stopped emitting MLflow traces - the experiment was created but stayed empty. This restores tracing, in the framework layer (`AgentApp` stays a generic invocation/durability contract, untouched).

## Root cause

Two agent-side causes:

**1. The per-request span was dropped by the durable-runtime refactor (#550).**
Before #550, each template shipped a `runtime/runtime.py` that wrapped every request in an MLflow span (`with mlflow.start_span(name="stream_handler")`). #550 deleted that file when moving to the SDK `AgentApp`, and the span was not carried over. That span is what made tracing work:

```
BEFORE #550                                  AFTER #550 (broken)
runtime/runtime.py                           runtime/main.py -> AgentApp
  with mlflow.start_span("stream_handler"):    (no span opened anywhere)
      graph.astream(...)                          graph.astream(...)
          |                                           |
   autolog nests under the span               no active trace -> autolog no-ops
          v                                           v
   trace recorded                              0 traces
```

Framework autolog records spans only when a trace is already active, and does **not** auto-start one for streaming invocations (`graph.astream` / `Runner.run_streamed`) - which is what the durable runtime uses to emit incremental events. Verified for both frameworks: with autolog bound but no root span, a streamed invocation produces **0 traces** (and `tag_session` warns `No active trace found`).

**2. langchain autolog was never bound in the langgraph template (regression from de-vendor #496).**
The langgraph template imported `configure_tracing` from top-level `databricks_mason` - the framework-neutral runtime version (`autolog=None`) - instead of `databricks_mason.langgraph`, whose `configure_tracing` binds `mlflow.langchain.autolog`. So even with a root span it would capture the root but no LLM/graph children.

## Fix

- **New SDK helper `root_span()`** in `databricks_mason.runtime.tracing`: opens a root MLflow span (no-op when tracing is disabled), re-exported from `runtime`, `langgraph`, and `openai`. This restores the per-request span #550 removed, in a place a future runtime refactor won't silently drop.
- **langgraph template**: import `configure_tracing` from `databricks_mason.langgraph` (binds autolog); wrap the invocation in `root_span`; move `tag_session` inside the span.
- **openai template**: same `root_span` wrap + `tag_session` inside (its import was already correct).

## Testing

Unit + static (from `integrations/mason`): **469 tests pass**, `ty` + `ruff` clean, both templates compile. Adds `runtime_tracing_test.py`.

End-to-end against e2-dogfood - each trace has the root `invoke` span with the framework/LLM spans nested, and zero `No active trace found` warnings:

| Scenario | Result |
| --- | --- |
| langgraph `mason dev` (durable) | trace, **5 spans**: `invoke -> LangGraph -> model -> _RoutedChatDatabricks -> HumanInTheLoopMiddleware.after_model` |
| openai `mason dev` (durable) | trace, **7 spans**: `invoke -> AgentRunner.run_streamed -> Agent workflow -> Agent -> Generation -> AsyncDatabricksCompletions` |
| langgraph `mason deploy` (deployed app) | trace, **5 spans**, status OK, exported by the app service principal |

Control (both frameworks, autolog bound but **no** root span): **0 traces** + `No active trace found` - confirms the root span is required, not incidental.

* OpenAI template trace: https://dogfood.staging.databricks.com/ml/experiments/2532651972788519/traces?o=6051921418418893&selectedEvaluationId=tr-9a502b57db910ca6787a67f4d72a5ecc
* Langgraph template trace: https://dogfood.staging.databricks.com/ml/experiments/2532651972788519/traces?o=6051921418418893&selectedEvaluationId=tr-1d24a0f0d8afa312da54baad88a1d214


integrations/mason/src/databricks_mason/runtime/tracing.py
```diff
@@ -41,11 +43,33 @@ def configure_tracing(autolog: Callable[[], None] | None = None) -> None:
...
+@contextlib.contextmanager
+def request_span(name: str = "agent", inputs: Any = None) -> Iterator[Any]:
+    """Open a root MLflow span for one agent invocation (no-op when tracing is disabled).
+
+    This is why traces appear at all. Framework autolog (``mlflow.langchain.autolog`` /
+    ``mlflow.openai.autolog``) only records spans when a trace is already active — for a LangGraph /
+    OpenAI-Agents invocation it does NOT start one on its own, so without a root span the whole
+    invocation produces zero traces. Opening this span around the handler gives autolog a parent to
+    nest its LLM/graph spans under, and gives ``tag_session`` an active trace to tag. ``AgentApp``
+    opens it around every invocation, so every framework and both runtimes are covered in one place.
```
jamesbxwu: can we avoid mentioning openai/langgraph here? we can add generic guidance that use this when `autolog` requires a root span to be started

integrations/mason/templates/agent-langgraph/agent/agent.py
```diff
@@ -159,24 +165,33 @@ async def _run_agent(
...
-    interrupted = bool(outputs and outputs[-1].get("type") == "interrupt")
-    return {
-        "output": [event["message"] if event["type"] == "message" else event for event in outputs],
-        "session_id": session_id,
-        "status": "interrupted" if interrupted else "completed",
-    }
+    # Open a root MLflow span around the invocation so a trace is recorded: langchain autolog only
+    # nests spans under an active trace and does not start one for astream. tag_session tags this
+    # trace; the graph's LLM/tool spans nest under it. No-op when tracing is disabled.
+    with request_span(name="agent", inputs=agent_input) as span:
```
jamesbxwu: can we use a better name than "agent"? 
```suggestion
    with request_span(name="invoke", inputs=agent_input) as span:
```

integrations/mason/templates/agent-openai/agent/agent.py
```diff
@@ -135,19 +141,27 @@ async def on_recovery(value: Any, context: DurableAgentContext) -> dict:
...
-    interrupted = bool(outputs and outputs[-1].get("type") == "interrupt")
-    return {
-        "output": [event["message"] if event["type"] == "message" else event for event in outputs],
-        "session_id": session_id,
-        "status": "interrupted" if interrupted else "completed",
-    }
+    # Open a root MLflow span around the invocation so a trace is recorded: OpenAI autolog only nests
+    # spans under an active trace and does not start one for a streamed run. tag_session tags this
+    # trace; the agent's LLM/tool spans nest under it. No-op when tracing is disabled.
+    with request_span(name="agent", inputs=payload) as span:
```
jamesbxwu: similar to above
```suggestion
    with request_span(name="invoke", inputs=payload) as span:
```

integrations/mason/src/databricks_mason/runtime/tracing.py
```diff
@@ -41,11 +43,33 @@ def configure_tracing(autolog: Callable[[], None] | None = None) -> None:
...
     else:
-        # runtime/runtime.py wraps every request in a span regardless; without an experiment it
-        # would try to export to a missing one (INVALID_PARAMETER_VALUE), so disable.
+        # AgentApp wraps every invocation in request_span; without an experiment it would try to
+        # export to a missing one (INVALID_PARAMETER_VALUE), so disable tracing outright.
         mlflow.tracing.disable()
 
 
+@contextlib.contextmanager
+def request_span(name: str = "agent", inputs: Any = None) -> Iterator[Any]:
```
jamesbxwu: why the name 'request_span`? can this be used outside of a request potentially?

jamesbxwu: Good point - renamed to `root_span`. It is a generic "open a root trace span" helper (nothing request-specific); it can wrap any operation that needs an active trace before autolog will nest. Done in 742f75b.

integrations/mason/src/databricks_mason/runtime/tracing.py
jamesbxwu: Done in 742f75b - dropped the langchain/openai specifics; the docstring now describes the general behavior ("autolog records spans only when a trace is already active and does not always start one, notably for streaming"). The framework-specific mention stays only in the template comments, where the framework is known.

integrations/mason/templates/agent-langgraph/agent/agent.py
jamesbxwu: Applied in 742f75b (span name is now `invoke`).

integrations/mason/templates/agent-openai/agent/agent.py
jamesbxwu: Applied in 742f75b (span name is now `invoke`).

integrations/mason/src/databricks_mason/runtime/tracing.py
```diff
@@ -5,13 +5,15 @@
 ``MLFLOW_EXPERIMENT_NAME``) — whichever pair the user or the Apps resource binding provides. MLflow
 resolves the specific value itself; this only decides on/off. Requiring both halves avoids the
 half-configured case where traces silently export to a local file store instead of the workspace.
-When unconfigured, tracing is disabled outright so the per-request span ``runtime/runtime.py`` opens
-has nothing to export to. No user decision lives here — it's all driven by env — so this whole module
-is a candidate to move behind an SDK helper.
+When unconfigured, tracing is disabled outright so ``start_trace`` (which agents open around each
+invocation) has nothing to export to. No user decision lives here — it's all driven by env — so this
+whole module is a candidate to move behind an SDK helper.
```
elainewang-db: qq: is this comment 

> so this whole module is a candidate to move behind an SDK helper.

still accurate?

elainewang-db (approved):

integrations/mason/src/databricks_mason/runtime/tracing.py
jamesbxwu: this is out of date, let me remove. It makes sense to keep this here since it helps to connect the agent code with the tracing configurations from `mason tracing` cli

PR #579, opened 2026-09-10
[mason] init: pin template to matching release tag when it exists

elainewang-db:
Every published databricks-mason version is a .dev0 pre-release, so the "dev in version -> fetch main" heuristic fired for every real install, defeating the release-tag pin and letting mason init scaffold a template ahead of the installed
package. Replace it with a git ls-remote check: fetch databricks-mason-v<version>
when that tag exists, else fall back to main (locally built, never-tagged builds).

tested e2e

### DIFF
--- a/integrations/mason/src/databricks_mason/init.py
+++ b/integrations/mason/src/databricks_mason/init.py
@@ -93,15 +93,22 @@ def _template_ref(framework: str) -> str:
     except PackageNotFoundError:
         return default_ref
     tag = f"{_RELEASE_TAG_PREFIX}{installed}"
-    return tag if _remote_ref_exists(_TEMPLATES[framework]["repo"], tag) else default_ref
+    if _remote_ref_exists(_TEMPLATES[framework]["repo"], tag):
+        return tag
+    render.warning(
+        f"No release tag '{tag}' for installed databricks-mason {installed}; scaffolding from "
+        f"'{default_ref}', which may be ahead of your installed package."
+    )
+    return default_ref
 
 
 def _remote_ref_exists(repo: str, tag: str) -> bool:
     """Whether `tag` exists in `repo`, via a lightweight `git ls-remote` (no clone).
 
-    A missing tag exits non-zero (2), which is an expected answer here rather than an error, so
-    this checks the return code directly instead of going through `_git`. Any failure to confirm
-    the tag (missing, or an unreachable remote) returns False, falling back to the default ref.
+    `git ls-remote --exit-code` exits 0 when the tag is found and 2 when it is genuinely absent (an
+    untagged local build — a legitimate `main` fallback). Any other exit is a real failure (an
+    unreachable remote, auth) that we surface rather than silently treating as "absent" and fetching
+    a possibly-drifted `main`.
     """
     result = subprocess.run(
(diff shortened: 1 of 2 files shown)


integrations/mason/src/databricks_mason/init.py
```diff
@@ -89,9 +92,24 @@ def _template_ref(framework: str) -> str:
         installed = _installed_version("databricks-mason")
     except PackageNotFoundError:
         return default_ref
-    if "dev" in installed or "+" in installed:
-        return default_ref
-    return f"{_RELEASE_TAG_PREFIX}{installed}"
+    tag = f"{_RELEASE_TAG_PREFIX}{installed}"
+    return tag if _remote_ref_exists(_TEMPLATES[framework]["repo"], tag) else default_ref
```
jamesbxwu: can we log a warning that when this triggers a fall back? this will help with debugging.  Maybe consider just printing out the exact version of the template every time?

integrations/mason/src/databricks_mason/init.py
```diff
@@ -77,10 +77,13 @@
...
     """The git ref to fetch a framework's template from, absent a `--ref` override.
 
-    For a versioned framework, fetch the tag matching the installed CLI so the scaffold's pinned
-    `databricks-mason` matches what the user has. Fall back to the default ref when the version
-    isn't a published release — an editable/dev build (e.g. `0.1.0.dev0`, or a local `+`
-    local-version install) has no corresponding tag, so those keep fetching `main`.
+    A registry install pins the scaffold's `databricks-mason` at its own version, so fetch the
+    template from the matching `databricks-mason-v<version>` release tag — the template as it
+    shipped with that release — rather than `main`, which may have drifted ahead. Pre-release
+    versions (`0.1.4.dev0`) are tagged like any other release, so they pin too. Fall back to `main`
```
jamesbxwu, line 83: i noticed the release workflow is temporarily disabled: https://github.com/databricks/databricks-ai-bridge/blob/main/.github/workflows/release-databricks-mason.yml#L20. when do we plan to turn this back on?

integrations/mason/src/databricks_mason/init.py
```diff
@@ -89,9 +92,24 @@ def _template_ref(framework: str) -> str:
...
+    this checks the return code directly instead of going through `_git`. Any failure to confirm
+    the tag (missing, or an unreachable remote) returns False, falling back to the default ref.
+    """
+    result = subprocess.run(
+        ["git", "ls-remote", "--exit-code", "--tags", repo, f"refs/tags/{tag}"],
+        text=True,
+        capture_output=True,
+        check=False,
+    )
+    return result.returncode == 0
```
jamesbxwu: falling back on any error feels wrong.  Should we only fall back on error code 2 and just error on anything else?

jamesbxwu (approved):

integrations/mason/src/databricks_mason/init.py
elainewang-db: i think the release workflow is through https://github.com/databricks/secure-public-registry-releases-eng/actions/workflows/release-databricks-mason.yml cc @smurching

PR #580, opened 2026-09-10
[mason] editable dev loop: bundle templates in the package

elainewang-db:
## What changed, and why?

**Change:** Templates now ship *inside* the `databricks_mason` package (`src/databricks_mason/templates/`). `mason init` copies the template that matches the installed CLI via `importlib.resources` — no git fetch, no version-matching heuristic, no `--repo`/`--ref`. The scaffold keeps its normal `databricks-mason` **PyPI** dependency. For an editable install (`pip install -e`), `importlib.resources` resolves to the source tree, so a contributor's uncommitted **template** edits are scaffolded live. The per-framework templates are now an `_AgentTemplate` dataclass (mason-server / custom-server / chat-app) instead of three parallel dicts.

**Why:** `mason init` previously fetched templates from a git ref chosen by a heuristic that drifted from the installed package (the drift behind #476, narrowly fixed in #579). Bundling the templates makes the scaffold match the installed CLI by construction, killing the drift.

## How do you know it works?

**Tests:** init copy-flow / template-mapping / dataclass unit tests pass; ruff / format / ty clean.

**Manual (e2-dogfood, editable install):** `mason init` scaffolded from the working tree — a working-tree template edit appeared in the generated project — and the built package wheel shipped all 74 template files; `mason dev` served a live model invocation.

## Notes

- Running a scaffold against *unreleased* SDK is a documented **manual** `[tool.uv.sources]` override (editable path for `mason dev`, a pushed git ref for `mason deploy`) — see the README's "Developing Mason" section. Automatic SDK pinning and the deploy wheel/reject logic were intentionally kept out of this PR.
- Most of the diff is a pure file move: the 72 template files are relocated into `src/databricks_mason/templates/` (`R100` renames) so they ship in the wheel. The substantive changes are `init.py`, its tests, the README, and a one-line `AGENTS.md` registration.


integrations/mason/README.md
jamesbxwu:
```suggestion
    configure     -experiment E
```

integrations/mason/src/databricks_mason/init.py
```diff
@@ -24,110 +23,31 @@
...
-    if result.returncode == 0:
-        return True
-    if result.returncode == 2:
-        return False
-    raise AgentCliError(
-        f"Could not check for release tag '{tag}'.",
-        hint=(result.stderr or result.stdout or "").strip(),
-    )
+# Framework -> template name (a directory under databricks_mason/templates/).
+_TEMPLATES = {"openai": "agent-openai", "langgraph": "agent-langgraph"}
```
jamesbxwu:
```suggestion
_MASON_SERVER_TEMPLATES = {"openai": "agent-openai", "langgraph": "agent-langgraph"}
```

integrations/mason/src/databricks_mason/init.py
```diff
@@ -24,110 +23,31 @@
...
+_CUSTOM_SERVER_TEMPLATES = {"openai": "custom-agent-openai", "langgraph": "custom-agent-langgraph"}
+_CHAT_APP_TEMPLATES = {"langgraph": "ui/agent-langgraph", "openai": "ui/agent-openai"}
```
jamesbxwu: instead of a dictionary, can we define a proper data class for these and operate on those instead of having to reason across several lists?

It can be in the form of
```
class AgentTemplate:
    framework: str # openai,langgraph,etc
    server_template: str
    chat_app_template: str
```

I was also confused on what this "overlay" thing was later and only to realize it was the chat app template

integrations/mason/src/databricks_mason/init.py
```diff
@@ -358,48 +264,27 @@ def init(
...
-                selected_repo,
-                resolved_ref or selected_ref,
-                runtime_extra=mason_server,
-            )
-        template_name = pathlib.PurePosixPath(template_path).name
+        # Copy the template bundled with the installed CLI, then pin databricks-mason per how that
+        # CLI is installed (editable / git checkout, else the template's own PyPI pin).
+        _copy_packaged_template(template_name, dest, overlay_names)
+        if mason_pin is not None:
+            _pin_mason(dest, selected_framework, mason_pin, runtime_extra=mason_server)
```
jamesbxwu: this part was very confusing for me to follow.  I had to really dig into `_resolve_mason_pin()` to understand that `_resolve_mason_pin()` returns None for PyPi registry install.  Can we make this more clear by removing `_resolve_mason_pin()` and instead just inline the logic around `mason_source` such as?

```
# The generated agent still needs to *install* databricks-mason. Match how *this* CLI was installed.
if (root := _editable_checkout_root()) is not None:
    _pin_mason(dest, ..., mason_source.editable(root / "integrations" / "mason"), ...)
elif (git := _installed_git_template_source()) is not None:
    _pin_mason(dest, ..., mason_source.git(*git), ...)
# else: PyPI — keep the version already in the template's pyproject.toml
```

integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -250,6 +251,27 @@ def _load_project(source: pathlib.Path):
...
+    if not pin:
+        return
+    if isinstance(pin.get("path"), str) or str(pin.get("git", "")).startswith("file://"):
+        raise AgentCliError(
+            "This project pins databricks-mason to a local checkout, which the Apps build can't "
+            "reach.",
+            hint="Install Mason from git and re-run `mason init` to deploy your changes: "
+            "pip install 'git+<repo>@<pushed-sha>#subdirectory=integrations/mason'. "
+            "The editable checkout stays for `mason dev`.",
+        )
```
jamesbxwu: when will we run into this? are we bleeding a mason developer devloop into the agent's `pyproject.toml` which is also used as the deploy config for customers?

junchoi-db (approved):

jamesbxwu (approved):

PR #584, opened 2026-09-14
[mason] deploy: fail fast when databricks-mason is pinned to a local path (ML-69653)

anshul0713:
## Root cause (confirmed via captured Apps logs)
A `mason init` run while mason itself is installed **editable/from a local checkout** records a `file://` git source for `databricks-mason` in the scaffold's `[tool.uv.sources]`:
```toml
[tool.uv.sources]
databricks-mason = {git = "file:///home/<user>/.../databricks-ai-bridge", rev = "...", subdirectory = "integrations/mason"}
```
That resolves for local `mason dev`, but the Databricks Apps build environment has no such path. The deployed app's `uv run start-server` then fails to install `databricks-mason` and **crashes at startup**, after a ~5-minute deploy, with an opaque error. Confirmed from `databricks apps logs`:
```
Updating file:///home/<user>/.../databricks-ai-bridge (<rev>)
× Failed to download and build databricks-mason @ git+file:///...
╰─▶ fatal: '/home/<user>/.../databricks-ai-bridge' does not appear to be a git repository
```

This is primarily a **developer-workflow sharp edge** (a PyPI or `git+https` install pins a *reachable* source, so end users don't hit it) — but the failure mode is a multi-minute deploy followed by an opaque crash you have to dig through Apps build logs to diagnose.

## Fix
`mason deploy` now inspects the project's `databricks-mason` source up front and **fails fast** with a clear, actionable error before any sync/deploy work when the pin is a local `file://` git source or a filesystem `path`. Deployable pins (remote git URL, registry version, or no override) are unaffected, and `mason dev` is untouched. This matches the "fail fast + surface the real error" direction of #581.

New error (returns exit 1, in ~1s instead of ~300s):
```
Error: This project pins databricks-mason to a local path (file:///...), which won't exist in the
Databricks Apps build environment — the deployed app can't install it and will crash at startup.
This happens when mason itself was installed editable from a local checkout. Re-scaffold from a
released or git install of mason ... or edit [tool.uv.sources].databricks-mason ... before deploying.
```

## Testing
- **Unit** (`deploy_bugfix_test.py`): `file://`-git and `path` sources rejected; remote-git and registry pins allowed; missing pyproject tolerated; `deploy` bails before touching the client/CLI. Full suite 484 passed, ruff clean.
- **Manual** (eng-ml-inference): reproduced the original crash (deploy "succeeded" but app was `● Crashed`; logs showed the `file://` clone failure), then confirmed the guard now fails in **1s with exit 1** and the actionable message on the same project.

Fixes ML-69653.


elainewang-db:
replaced by https://github.com/databricks/databricks-ai-bridge/pull/580?timeline_per_page=5

PR #586, opened 2026-09-14
[mason] agent.toml as the source of truth for stores: init declares, deploy reconciles

nihit:
## Summary

Delineates the `init` / `bind` / `dev` / `deploy` responsibilities around memory & session stores so **`agent.toml` is the single source of truth**. Previously `deploy` silently invented and wrote back `<name>-memory` / `<name>-session` bindings, and `bind` provisioned stores. Now:

- **`mason init`** *declares* a memory + session store by default — names derived from the project directory (`<dir>-memory` / `<dir>-session`), overridable with `--memory-store` / `--session-store`. Only for `--server mason`; `--server custom` stays store-free unless a name is passed.
- **`mason memory/sessions bind`** are *declarative-only* — they edit `agent.toml` and make no workspace calls. (`--no-create-stores` removed.)
- **`mason deploy`** is the *only* resource-provisioning verb: it creates any declared-but-missing store (with a per-store notice) and *never* rewrites `agent.toml`.
- **`mason dev`** *warns (not errors)* when a declared store isn't created yet, and stays offline-friendly.

Because `agent.toml` now carries the store **name** only, the resolved memory-store **id** (which the entries API is keyed by) reaches the runtime via the `AGENT_MEMORY_STORE` env var — injected by `deploy` (into `app.yaml`) and `dev` (into the dev-only manifest).

## Testing

`databricks-mason` unit suite: **484 passing** (`uv run --group tests pytest tests/unit_tests/`). 
`ruff check` + `ruff format` clean. 
`ty check` clean apart from one pre-existing diagnostic in the untouched `errors.py`

## Follow-ups (not in this PR)

- **Runtime memory-store fallback:** when neither `AGENT_MEMORY_STORE` nor an `id` in `agent.toml` is present, `resolve_memory_store()` returns the display *name* (used as an id → wrong entries path). Unreachable via `dev`/`deploy` (both inject the env), so deferred — to be resolved fail-closed (return `None`) or via a runtime name→id lookup


qinxinw-db (approved):

elainewang-db (approved):

integrations/mason/src/databricks_mason/deploy.py
```diff
@@ -281,52 +282,30 @@ def _resolve_deployment_name(project, name: Optional[str]) -> str:
...
-        _ensure_session_store(client, f"{base_name}-session")
-        project.bind_session_store(f"{base_name}-session")
-        changed = True
-    if changed:
-        project.write()
-
-
-def validate_stores(client, *, memory_store: Optional[str], session_store: Optional[str]) -> None:
-    """Validate the agent's bound stores exist. Shared by `mason deploy` and `mason dev`.
+def _reconcile_declared_stores(
```
jamesbxwu, line 285: @shivam5 i see the runtime store (currently called durability store) is not using this path, are we planning to unify?

jamesbxwu (commented):
thanks for making this change! Left a small (non-blocking) comment for @shivam5

integrations/mason/src/databricks_mason/deploy.py
shivam5: yes, can make this change in a separate pr

shivam5: Yes—we should unify this at the `mason deploy` reconciliation phase, while keeping separate lifecycle helpers. #586 reconciles named, user-declared Memory/Session Stores. #589 now adds `_reconcile_runtime_store(...)` for the implicit, deployment-scoped Runtime Store. After #586 merges, I will rebase #589 and retain both calls side by side. If #589 lands first, the symmetric conflict resolution applies in #586. This keeps provisioning unified without making Runtime Stores bindable or shareable.

PR #589, opened 2026-09-14
Refactor Mason runtime durability boundary

shivam5:
## Summary

Refactor Mason Runtime so the common invocation runtime is independent from the Lakebase-only
heartbeat and recovery capabilities. This implements the review feedback on construction,
terminology, ownership boundaries, and internal documentation.

## Low-level design

```text
Client
  │ POST /api/invocations
  │ GET  /api/invocations/{id}
  │ GET  /api/invocations/{id}/events
  ▼
runtime/app.py
  AgentApp: HTTP validation, response modes, SSE, @app.invoke / @app.recover
  ▼
runtime/runtime.py
  Runtime: idempotent submit, polling, results, and event replay
  ├────────────────────────────────────────────────────────────┐
  │                                                            │
  ▼                                                            ▼
runtime/store.py                                     runtime/execution.py
RuntimeStore                                         InvocationExecutor
accept/get/claim/complete/fail/events                schedules attempts
  │                                                            │
  │                                                            ▼
  │                                               AttemptExecution
  │                                               calls agent hook and writes
  │                                               events / terminal state
  │
  ├── InMemoryRuntimeStore ◄── LocalInvocationExecutor
  │       process-local          no heartbeat or recovery
  │
  └── DurableRuntimeStore ◄── DurableInvocationExecutor
          durable extension       ├── Heartbeat
          for leases/recovery     └── RecoveryScheduler
                    ▲
                    │
        LakebaseDurableRuntimeStore
```

### Construction and lifecycle

`AgentApp` uses `Runtime.from_environment(...)` for Mason-managed processes and
`Runtime.from_store(...)` when a caller supplies a store explicitly. The selected store determines
the implementation:

- `InMemoryRuntimeStore` selects `Runtime.local()` and `LocalInvocationExecutor`.
- A `DurableRuntimeStore` selects `Runtime.durable()` and `DurableInvocationExecutor`.
- The durable executor always scans persisted `QUEUED` work and starts its first attempt through
  `@app.invoke`, including after process restart. Reads also schedule queued work as a safety net.
- `@app.recover` is optional. Its presence enables a separate scanner for stale `ACTIVE` attempts;
  durable results, events, and queued first-attempt scheduling do not depend on it.

`RuntimeStore.accept(invocation_id, request)` is the idempotency boundary: an identical retry
returns the existing invocation, while a different request using the same ID raises
`InvocationConflictError`.

The same design is documented beside the implementation in
`integrations/mason/src/databricks_mason/runtime/README.md`.

## What changed

- Adds explicit `Runtime.local()`, `Runtime.durable()`, `Runtime.from_store()`, and
  `Runtime.from_environment()` factories. Explicit-store factories no longer perform environment
  discovery.
- Uses `RuntimeStore` / `DurableRuntimeStore`, `InMemoryRuntimeStore`, and
  `LakebaseDurableRuntimeStore` consistently.
- Standardizes state, contexts, errors, and database identifiers on **invocation** terminology.
- Moves durable scheduling to `runtime/durability/execution.py`; the scheduler now depends on a
  small `RecoverySchedulerTarget` protocol instead of a concrete executor.
- Keeps recovery eligibility in the durable store, avoiding duplicated scheduler/store predicates.
- Makes heartbeats durable-only: Lakebase owns `heartbeat_at`; shared invocation state does not.
- Renames Lakebase tables and fields to `invocations`, `invocation_events`, and `invocation_id`.
- Adds method-level store-contract documentation and explicit Runtime Store environment names.
- Removes unreleased compatibility aliases and the old `durable_runtime` constructor flag.
- Updates only the template API references required by the unreleased rename
  (`InvocationContext`, `@app.recover`).

## Scope and compatibility

This PR is the runtime-internal refactor. It does **not** include the generated-template
organization work in #590 or the product configuration, CLI, UI, and developer documentation
changes in #592.

The previous durable Python aliases and constructor flag have not been publicly released, so this
PR intentionally does not retain compatibility code. The invocation HTTP API and its foreground,
background, streaming, polling, and reconnect behavior stay the same. Direct imports of internal
runtime modules remain unsupported.

### Automated

- Focused runtime suite: `66 passed`.
- Full Mason unit suite on the rebased head: `487 passed`.
- Rebase-affected deploy/dev/runtime suite: `152 passed`.
- `mason dev` functional regression: `1 passed`.
- LangGraph template against the exact unreleased wheel: `14 passed, 1 skipped`.
- OpenAI template against the exact unreleased wheel: `15 passed, 1 skipped`.
- Custom LangGraph template route tests: `2 passed`.
- Ruff, formatting, `ty check`, and `git diff --check` passed.
- All current GitHub CI checks are green.

### Authenticated end-to-end

Revalidated on `5da7b2f` with the explicitly selected `e2-dogfood-apps` profile. The test app is
pinned to a wheel whose runtime files exactly match this commit.

- Fresh `mason init` scaffolds succeeded for managed LangGraph, managed OpenAI, and custom
  LangGraph projects.
- Real `mason dev` ran through `databricks apps run-local` and the Apps proxy. Foreground execution,
  idempotent retry, `409` request conflict, background polling, foreground streaming, and SSE replay
  all passed using the in-memory Runtime Store.
- `mason deploy pr589-rb-2005` successfully synced and rolled out the existing isolated test app,
  reusing its Lakebase Runtime Store. The app returned to `RUNNING` / `ACTIVE` after validation.
- Through the deployed Apps URL, foreground execution, durable background polling, live streaming,
  and event replay all passed. Results and events were read back from the durable store.
- Persisted queued work is now covered by unit tests: after runtime restart it begins attempt `1`
  through `@app.invoke` even when no `@app.recover` hook exists; GET/wait also schedules queued work
  accepted by another process.
- Controlled recovery passed again on the final review-fix wheel: after stopping the App while a
  120-second background invocation was active, restarting it detected the stale heartbeat and
  completed the same invocation through `@app.recover` on attempt `2`.

```json
{"mode":"recover","attempt":2}
```


shivam5:
Generated-template cleanup has been split into #590 to keep this runtime/CLI/naming PR focused. #590 should merge after this PR and will then be rebased onto the new runtime API.

integrations/mason/README.md
```diff
@@ -20,7 +20,7 @@ From source:
 pip install 'git+https://github.com/databricks/databricks-ai-bridge.git#subdirectory=integrations/mason'
 ```
 
-For the SDK-hosted durable agent application, install the runtime extra:
+For the SDK-hosted Mason Runtime agent application, install the runtime extra:
```
shivam5: does this need to be extra? why?

integrations/mason/README.md
```diff
@@ -109,22 +109,22 @@ the existing CLI commands remain separate.
 ## Agent application
 
 `AgentApp` provides Mason's invocation HTTP contract, including foreground, streaming, background,
-polling, and event endpoints. By default its state is process-local. Set `durable_runtime=True` to
-use Lakebase persistence, heartbeats, and crash recovery after deployment:
+polling, and event endpoints. `mason dev` uses process-local state; `mason deploy` attaches a
```
shivam5: Link to RUNTIME.md for more details? 

Also somewhere mention if you don't want to use our runtime layer, you can use --server custom.

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -59,30 +61,38 @@ def __init__(
         self,
         *,
         durable_runtime: bool = False,
-        durability_store: DurabilityStore | None = None,
+        durability_store: RuntimeStore | None = None,
```
jamesbxwu: rename to `runtime_store`?

integrations/mason/src/databricks_mason/runtime/durability/recovery.py
jamesbxwu, line 1: in `app.py` and the http contract we are calling it `invocation_id` but here it's called `execution_id` can we standardize on `invocation_id`?  If so, the renames should apply to other areas such as types, errors, etc

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -59,30 +61,38 @@ def __init__(
...
+                execution,
+                runtime_store=store,
+                recovery_enabled=lambda: durable_runtime and self._on_recovery_hook is not None,
+                heartbeat_seconds=3.0,
+                stale_seconds=10.0,
+                scan_seconds=3.0,
+            )
+        else:
+            executor = LocalInvocationExecutor(execution, runtime_store=store)
+        self._runtime = Runtime(runtime_store=store, executor=executor)
```
jamesbxwu: the instantiation logic of the runtime feels really awkward and depends on the combination of `durable_runtime` then `RuntimeStore` and finally `InvocationExecutor`.  Can we simplify this by introduce a factory pattern for constructing the `Runtime` which can then be gated on `durable_runtime` and whether `_on_recovery_hook` is defined.

High level
```
# runtime.py factory methods
@classmethod
def local(
    cls,
    execute_fn: DurableExecutorFn,
    *,
    runtime_store: RuntimeStore | None = None,
    poll_seconds: float = 1.0,
) -> Runtime:
    store = InMemoryRuntimeStore() if runtime_store is None else runtime_store
    execution = AttemptExecution(execute_fn, runtime_store=store)
    return cls(
        runtime_store=store,
        executor=LocalInvocationExecutor(execution, runtime_store=store),
        poll_seconds=poll_seconds,
    )

@classmethod
def durable(
    cls,
    execute_fn: DurableExecutorFn,
    *,
    runtime_store: RuntimeStore | None = None,
    recovery_enabled: Callable[[], bool] = lambda: True,
    heartbeat_seconds: float = 3.0,
    stale_seconds: float = 10.0,
    scan_seconds: float = 3.0,
    poll_seconds: float = 1.0,
) -> Runtime:
    from databricks_mason.runtime.durability.attempt import DurableInvocationExecutor
    from databricks_mason.runtime.durability.store import (
        DurableRuntimeStore,
        default_durability_store,
    )

    store = default_durability_store() if runtime_store is None else runtime_store
    if not isinstance(store, DurableRuntimeStore):
        raise TypeError("durable runtime requires a DurableRuntimeStore")
    execution = AttemptExecution(execute_fn, runtime_store=store)
    return cls(
        runtime_store=store,
        executor=DurableInvocationExecutor(
            execution,
            runtime_store=store,
            recovery_enabled=recovery_enabled,
            heartbeat_seconds=heartbeat_seconds,
            stale_seconds=stale_seconds,
            scan_seconds=scan_seconds,
        ),
        poll_seconds=poll_seconds,
    )

# app.py wiring
if durable_runtime:
    self._runtime = Runtime.durable(
        self._execute,
        runtime_store=durability_store,
        recovery_enabled=lambda: self._on_recovery_hook is not None,
    )
else:
    self._runtime = Runtime.local(self._execute, runtime_store=durability_store)
```

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,125 @@
...
+
+class Runtime:
+    """Coordinate invocation state, execution scheduling, polling, and event replay."""
+
+    def __init__(
+        self,
+        *,
+        runtime_store: RuntimeStore,
+        executor: InvocationExecutor,
+        poll_seconds: float = 1.0,
```
jamesbxwu, line 51: this class needs more documentation.  Can you explain the concept of runtime store, executor and polling?

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,125 @@
...
+    async def stop(self) -> None:
+        """Stop local work and close the Runtime Store."""
+        if not self._started:
+            return
+        await self.executor.stop()
+        self._started = False
+        await self.runtime_store.close()
+
+    async def submit(self, execution_id: str, request: JsonValue) -> DurableExecution:
+        """Accept an idempotent request and schedule it on this worker."""
```
jamesbxwu: what makes the request idempotent? it's not clear from just the function params nor the logic

integrations/mason/src/databricks_mason/runtime/store.py
```diff
@@ -0,0 +1,218 @@
...
+RUNTIME_SCHEMA_ENV = "DATABRICKS_MASON_RUNTIME_SCHEMA"
+RUNTIME_LOCAL_ENV = "DATABRICKS_MASON_RUNTIME_LOCAL"
+
+
+def _validate_execution_id(execution_id: str) -> None:
+    if not execution_id:
+        raise ValueError("execution_id must not be empty")
+
+
+class RuntimeStore(Protocol):
```
jamesbxwu, line 30: let's add docstring for each method here

integrations/mason/src/databricks_mason/runtime/store.py
```diff
@@ -0,0 +1,218 @@
...
+    DurableExecutionStatus,
+    DurableRequestConflictError,
+    JsonObject,
+    JsonValue,
+)
+
+DEFAULT_RUNTIME_SCHEMA = "databricks_mason_runtime"
+RUNTIME_ENDPOINT_ENV = "DATABRICKS_MASON_RUNTIME_ENDPOINT"
+RUNTIME_SCHEMA_ENV = "DATABRICKS_MASON_RUNTIME_SCHEMA"
+RUNTIME_LOCAL_ENV = "DATABRICKS_MASON_RUNTIME_LOCAL"
```
jamesbxwu: let's make these more explicit
```suggestion
RUNTIME_STORE_LOCAL_ENV = "DATABRICKS_MASON_RUNTIME_STORE_LOCAL"
RUNTIME_STORE_LAKEBASE_ENDPOINT_ENV = "DATABRICKS_MASON_RUNTIME_STORE_LAKEBASE_ENDPOINT"
RUNTIME_STORE_SCHEMA_ENV = "DATABRICKS_MASON_RUNTIME_STORE_SCHEMA"
```

integrations/mason/src/databricks_mason/runtime/store.py
```diff
@@ -0,0 +1,218 @@
...
+        return persisted
+
+
+def default_runtime_store() -> RuntimeStore:
+    """Use the attached Lakebase resource when deployed, otherwise process-local state."""
+    if os.getenv(RUNTIME_LOCAL_ENV, "").lower() == "true":
+        return InMemoryRuntimeStore()
+    app_name = os.getenv("DATABRICKS_APP_NAME")
+    if not app_name:
+        return InMemoryRuntimeStore()
```
jamesbxwu: why does the lack of app_name -> InMemoryRuntimeStore?

integrations/mason/src/databricks_mason/runtime/execution.py
```diff
@@ -0,0 +1,154 @@
...
+                await self._runtime_store.fail(execution_id, claimed.attempt)
+            except Exception:
+                logger.exception(
+                    "Failed to persist invocation failure: %s attempt=%d",
+                    execution_id,
+                    claimed.attempt,
+                )
+
+
+class LocalInvocationExecutor:
```
jamesbxwu:
```suggestion
class LocalInvocationExecutor(InvocationExecutor):
```

integrations/mason/src/databricks_mason/runtime/durability/attempt.py
```diff
@@ -1,140 +1,110 @@
...
-def _copy_json_object(value: JsonObject, name: str) -> JsonObject:
-    copied = copy_json_value(value, name)
-    if not isinstance(copied, dict):
-        raise TypeError(f"{name} must be a JSON object")
-    return copied
-
-
-class AttemptRunner:
-    """Claim, execute, heartbeat, and commit one durable attempt."""
+class DurableInvocationExecutor:
```
jamesbxwu:
```suggestion
class DurableInvocationExecutor(InvocationExecutor):
```

integrations/mason/src/databricks_mason/runtime/durability/execution.py
jamesbxwu, line 1: should this be renamed to `execution.py` or `durable_execution.py`?

integrations/mason/src/databricks_mason/runtime/types.py
```diff
@@ -1,30 +1,28 @@
...
 from enum import Enum
-from typing import Protocol, TypeAlias
+from typing import TypeAlias
 
 JsonValue: TypeAlias = None | bool | int | float | str | list["JsonValue"] | dict[str, "JsonValue"]
 JsonObject = dict[str, JsonValue]
 DurableEventEmitter = Callable[[JsonObject], Awaitable[int]]
 
 
 class DurableExecutionStatus(str, Enum):
```
jamesbxwu: is it redundant to call this `DurableExecutionStatus` instead of `ExecutionStatus`?

integrations/mason/src/databricks_mason/runtime/types.py
```diff
@@ -87,7 +85,7 @@ async def emit(self, event: JsonObject) -> int:
 
 @dataclass(frozen=True)
 class DurableAgentContext:
```
jamesbxwu: similarly, is it redundant to have `Durable` in the name?

integrations/mason/src/databricks_mason/runtime/types.py
```diff
@@ -107,76 +105,6 @@ async def emit(self, event: JsonObject) -> int:
...
-        self,
-        execution_id: str,
-        after_sequence: int | None = None,
-    ) -> list[DurableEvent]:
-        """Return execution events after an optional exclusive replay cursor."""
-        ...
-
-
 class DurableRequestConflictError(ValueError):
     """Raised when an execution ID is reused with a different request."""
```
jamesbxwu: we are using request, execution and invocation interchangeably, should this be renamed to either ExecutionConflictError or InvocationConflictError?

integrations/mason/src/databricks_mason/runtime/types.py
jamesbxwu: consider renaming execution to invocation to match the top level app contract?

integrations/mason/src/databricks_mason/runtime/durability/attempt.py
```diff
@@ -1,140 +1,110 @@
...
-        except Exception:
-            logger.exception("Failed to claim durable execution: %s", execution_id)
-            return
-        if claimed is None:
+        self._scan_seconds = scan_seconds
+        self._tasks: dict[str, asyncio.Task[None]] = {}
+        self._recovery_scheduler = None
+
+    async def start(self) -> None:
+        from databricks_mason.runtime.durability.recovery import RecoveryScheduler
```
jamesbxwu: I see a cycle of `DurableInvocationExecutor` importing `RecoveryScheduler` and vice verse.  This is a signal the scheduler should depend on a small protocol (`ensure_recovery_scheduled`) instead of the concrete class—the same pattern as `InvocationExecutor`.

integrations/mason/src/databricks_mason/runtime/durability/recovery.py
```diff
@@ -47,34 +46,21 @@ def start(self, *, recover: bool) -> None:
...
-            return
-        task = asyncio.create_task(
-            self._attempt_runner.run(state.execution_id),
-            name=f"durable-execution-{state.execution_id}",
-        )
-        self._tasks[state.execution_id] = task
-        task.add_done_callback(lambda completed: self._discard_task(state.execution_id, completed))
+        self._executor.ensure_recovery_scheduled(state)
 
     def _is_recoverable(self, state: DurableExecution) -> bool:
```
jamesbxwu: i see the same/similar recovery logic defined in 3 places
1.  here in `RecoveryScheduler._is_recoverable` 
2. `InMemoryDurabilityStore._is_recoverable`
3. `LakebaseDurableRuntimeStore.recoverable_execution_ids`

I worry about drift happening over time.  can we at least consolidate the first two?

jamesbxwu (commented):
thanks for the refactoring, looks much cleaner. 

Can you add the low level design in the PR description to either a README.md or AGENTS.md?

integrations/mason/README.md
shivam5: Yes. This is not a durability switch: `AgentApp` needs FastAPI/Uvicorn and the selected framework integration, while the base package is intentionally CLI-light. #592 now says that explicitly and shows both framework extras; generated projects declare the correct one automatically. If we later support framework-neutral `AgentApp` projects, we can add a separate core-server extra without making the CLI heavier.

integrations/mason/README.md
shivam5: Done in #592: the AgentApp section links to `RUNTIME.md`, and it explains that `mason init --server custom` creates a developer-owned FastAPI server with no Mason Runtime Store provisioning.

integrations/mason/src/databricks_mason/runtime/app.py
shivam5: Done in ef5c06b: the constructor parameter is now `runtime_store`.

integrations/mason/src/databricks_mason/runtime/durability/recovery.py
shivam5: Done in ef5c06b: state, contexts, errors, table/column names, scheduler inputs, and the HTTP-facing terminology now consistently use `invocation` / `invocation_id`.

integrations/mason/src/databricks_mason/runtime/app.py
shivam5: Done in ef5c06b. `Runtime.local()`, `Runtime.durable()`, and `Runtime.from_store()` centralize construction. Since the unreleased durability flag is removed, `AgentApp` delegates to `from_store`; the concrete store selects the executor and the recovery callback is gated by `@app.recover` registration.

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: Done in ef5c06b. The Runtime class now documents store ownership, executor responsibility, polling, idempotency, and factory selection. The full internal flow is also in `runtime/README.md`.

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: Done in ef5c06b. `Runtime.submit()` now documents that `RuntimeStore.accept(invocation_id, request)` owns idempotency: identical retries return existing state and different requests raise `InvocationConflictError`.

integrations/mason/src/databricks_mason/runtime/store.py
shivam5: Done in ef5c06b: every method in the shared and durable store protocols now has a contract docstring.

integrations/mason/src/databricks_mason/runtime/store.py
shivam5: Done in ef5c06b with the explicit `RUNTIME_STORE_LOCAL_ENV`, `RUNTIME_STORE_LAKEBASE_ENDPOINT_ENV`, and `RUNTIME_STORE_SCHEMA_ENV` names.

integrations/mason/src/databricks_mason/runtime/store.py
shivam5: The former app-name predicate is gone. `default_runtime_store()` now has an explicit rule: the local marker wins, a complete endpoint/schema pair creates the Lakebase store, neither selects in-memory, and a partial pair is an error. Its docstring explains the dev/deploy/direct-ASGI cases.

integrations/mason/src/databricks_mason/runtime/execution.py
shivam5: Done in ef5c06b: `LocalInvocationExecutor` implements `InvocationExecutor`.

integrations/mason/src/databricks_mason/runtime/durability/attempt.py
shivam5: Done in ef5c06b: `DurableInvocationExecutor` implements `InvocationExecutor`.

integrations/mason/src/databricks_mason/runtime/durability/execution.py
shivam5: Done in ef5c06b: the durable implementation is now `runtime/durability/execution.py`.

integrations/mason/src/databricks_mason/runtime/types.py
shivam5: Done in ef5c06b: shared types are now `InvocationStatus`, `Invocation`, `InvocationEvent`, and related invocation-named errors/contexts.

integrations/mason/src/databricks_mason/runtime/types.py
shivam5: Done in ef5c06b: the durable prefix is retained only on the durable store extension and executor, where it distinguishes extra heartbeat/recovery capabilities from common runtime behavior.

integrations/mason/src/databricks_mason/runtime/types.py
shivam5: Done in ef5c06b: this is now `InvocationConflictError`.

integrations/mason/src/databricks_mason/runtime/types.py
shivam5: Done in ef5c06b: execution/request terminology was standardized to invocation throughout the public runtime contract.

integrations/mason/src/databricks_mason/runtime/durability/attempt.py
shivam5: Done in ef5c06b. `RecoveryScheduler` now depends on the small `RecoverySchedulerTarget` protocol, so there is no scheduler/executor import cycle.

integrations/mason/src/databricks_mason/runtime/durability/recovery.py
shivam5: Done in ef5c06b. The durable store is now the single authority for recovery eligibility: it lists recoverable IDs and atomically claims them. The scheduler no longer duplicates an eligibility predicate.

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,245 @@
...
+            ),
+            poll_seconds=poll_seconds,
+        )
+
+    @classmethod
+    def from_store(
+        cls,
+        execute_fn: InvocationExecutorFn,
+        *,
+        runtime_store: RuntimeStore | None = None,
```
jamesbxwu: since this is called `from_store()`, we should expect that `runtime_store` cannot be None.  I see that we still delegate to `default_runtime_store()` which actually constructs the store from env variables, can we separate the two by introducing a new `Runtime.from_environment()`? this makes the api contract much cleaner 

Some ideas:
```
# durability/store.py
def runtime_store_from_environment() -> RuntimeStore:
    if os.getenv(RUNTIME_STORE_LOCAL_ENV, "").lower() == "true":
        return InMemoryRuntimeStore()
    endpoint = os.getenv(RUNTIME_STORE_LAKEBASE_ENDPOINT_ENV)
    schema = os.getenv(RUNTIME_STORE_SCHEMA_ENV)
    if endpoint and schema:
        from databricks_mason.runtime.durability.lakebase_runtime_store import (
            LakebaseDurableRuntimeStore,
        )
        return LakebaseDurableRuntimeStore.from_app_resource(...)
    if endpoint or schema:
        raise RuntimeError(...)
    return InMemoryRuntimeStore()
```
```
# runtime.py
@classmethod
def from_store(
    cls,
    execute_fn,
    *,
    runtime_store: RuntimeStore,  # required
    recovery_enabled=None,
    ...
) -> Runtime:
    if isinstance(runtime_store, DurableRuntimeStore):
        return cls.durable(execute_fn, runtime_store=runtime_store, ...)
    return cls.local(execute_fn, runtime_store=runtime_store, ...)

@classmethod
def from_environment(cls, execute_fn, **kwargs) -> Runtime:
    return cls.from_store(
        execute_fn,
        runtime_store=runtime_store_from_environment(),
        **kwargs,
    )
```
```
# app.py
self._runtime = (
    Runtime.from_store(self._execute, runtime_store=runtime_store, ...)
    if runtime_store is not None
    else Runtime.from_environment(self._execute, ...)
)
```

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,245 @@
...
+            executor=LocalInvocationExecutor(execution, runtime_store=store),
+            poll_seconds=poll_seconds,
+        )
+
+    @classmethod
+    def durable(
+        cls,
+        execute_fn: InvocationExecutorFn,
+        *,
+        runtime_store: RuntimeStore | None = None,
```
jamesbxwu: can we tighten this to be `DurableRuntimeStore` and do not accept None?

integrations/mason/src/databricks_mason/runtime/runtime.py
```diff
@@ -0,0 +1,245 @@
...
+        """Return the current state for an invocation ID."""
+        self._require_started()
+        return await self.runtime_store.get(invocation_id)
+
+    async def wait(
+        self,
+        invocation_id: str,
+        *,
+        timeout: float | None = None,
+    ) -> JsonValue:
```
jamesbxwu, line 239: [from my agent review], can you check if this is valid?

get_invocation() / wait() are now read-only. Previously, get_execution() called ensure_scheduled(), and QUEUED was eligible even when recovery was off.

ensure_scheduled() only runs from submit() for status == QUEUED. The durable scanner starts only when recovery_enabled() is true (@app.recover present), and that scanner is also the only path that will pick up a QUEUED row on another replica.

If a worker dies after accept (HTTP 202) and before claim, clients that poll GET/wait never reschedule. With no @app.recover, the invocation stays QUEUED forever (the startup warning only mentions crash recovery, not first-attempt failover). A failed in-process claim() has the same hang.

Please schedule QUEUED work on GET/wait using claim() (not claim_recoverable), independent of the recover hook. Keep the scanner for stale ACTIVE only when a recover hook exists.

integrations/mason/src/databricks_mason/runtime/store.py
```diff
@@ -0,0 +1,236 @@
...
+    InvocationEvent,
+    InvocationStatus,
+    JsonObject,
+    JsonValue,
+)
+
+DEFAULT_RUNTIME_STORE_SCHEMA = "databricks_mason_runtime"
+RUNTIME_STORE_LAKEBASE_ENDPOINT_ENV = "DATABRICKS_MASON_RUNTIME_STORE_LAKEBASE_ENDPOINT"
+RUNTIME_STORE_SCHEMA_ENV = "DATABRICKS_MASON_RUNTIME_STORE_SCHEMA"
+RUNTIME_STORE_LOCAL_ENV = "DATABRICKS_MASON_RUNTIME_STORE_LOCAL"
```
jamesbxwu, line 22: any mason CLI changes needed due to the rename?

shivam5: yes, the internal cli changes have been made.
mason dev writes DATABRICKS_MASON_RUNTIME_STORE_LOCAL into its temporary app.masondev.yaml, and mason deploy writes the Lakebase endpoint and Runtime Store schema variables into the deployable app.yaml

integrations/mason/src/databricks_mason/runtime/runtime.py
shivam5: this makes sense. looks like a narrow but valid failure case.

I will split the scanner into 2. one scanner which scans for queued tasks, which will always execute.
the recovery scanner will only look for active and stale heartbeat ones, and only fire if @app.recover is registered

integrations/mason/src/databricks_mason/runtime/store.py
shivam5: The required internal CLI wiring is already included in #589:  injects  through its temporary , and  injects the Lakebase endpoint/schema variables through the deployable . Unit and cross-process functional tests cover these names, and I revalidated both real dev and deploy flows. No user-facing CLI command rename is needed in this refactor; the broader product/config terminology remains isolated in #592.

integrations/mason/src/databricks_mason/runtime/store.py
```diff
@@ -0,0 +1,236 @@
...
+        self.persisted_events.append(persisted)
+        return persisted
+
+
+def runtime_store_from_environment() -> RuntimeStore:
+    """Construct the Runtime Store selected by the process environment.
+
+    ``mason dev`` sets the local marker to override any inherited Lakebase variables. A deployed
+    Mason server receives both Lakebase variables from ``mason deploy``. Direct ASGI use and tests
+    have neither, and intentionally use process-local state.
```
jamesbxwu: let's avoid mentioning the mason cli here so we can better decouple
```suggestion
```

jamesbxwu (approved):
thanks for addressing all of the comments!

integrations/mason/src/databricks_mason/runtime/store.py
shivam5: Removed the CLI-specific explanation. The docstring now describes only the environment contract: the local marker takes precedence, complete Lakebase configuration selects durable storage, and otherwise the runtime uses process-local state. Addressed in 5da7b2f.

PR #590, opened 2026-09-14
Separate Mason runtime adapters from agent code

shivam5:
## Summary

Make the generated Mason templates tell a simple developer story:

```text
client
  -> runtime/main.py      # construct AgentApp and register hooks
  -> runtime/adapter.py   # translate Mason input/events/output
  -> agent/agent.py       # run the framework-native agent
```

Developers can bring an existing agent by keeping its normal execution code in `agent/agent.py` and exposing a small `run_agent(...)` entrypoint. Mason-specific request types, event translation, and hook registration stay outside the portable agent code.

## Low-level design

| Component | Owns | Calls / depends on |
|---|---|---|
| `runtime/main.py` | `AgentApp` construction, `invoke`/`recover` registration, optional UI/custom FastAPI routes | `runtime.adapter.invoke`, `runtime.adapter.recover` |
| `runtime/adapter.py` | Mason payload validation, session/actor/model selection, HITL input translation, streaming event translation, result shaping, recovery policy | `agent.agent.run_agent` and the framework-specific recovery helper |
| `agent/agent.py` | Framework-native agent, model, instructions, tools, Session/Memory integrations, MCP lifecycle, HITL state, tracing, and `run_agent` | Framework SDKs and Mason's framework helpers, but no `InvocationContext` or HTTP request types |

### Normal invocation

```text
POST /api/invocations
  -> AgentApp invokes runtime.adapter.invoke
  -> adapter translates the application payload
  -> adapter calls agent.run_agent
  -> native framework events are translated and emitted through InvocationContext
  -> adapter returns the Mason invocation result
```

### Recovery invocation

```text
stale invocation
  -> AgentApp invokes runtime.adapter.recover with the original input
  -> adapter chooses framework-specific recovery input
  -> adapter calls the same agent.run_agent entrypoint
```

- LangGraph recovery behavior is unchanged: continue from the invocation's checkpoint when available, otherwise replay the original input.
- OpenAI Agents SDK has no checkpoint-continuation primitive. Recovery reuses the same session and replays the original messages with a leading `developer` instruction that identifies the run as a recovery attempt, warns that tool calls or external side effects may already have completed or remain in progress, and asks the agent to check state before repeating them.
- The normal OpenAI `invoke` path is unchanged, and the adapter does not mutate the caller's message list.

## Why

The previous templates mixed framework-native agent code with Mason's HTTP/runtime protocol in one large `agent.py`. That made existing agents harder to bring to Mason and obscured which code developers were expected to own.

This split keeps the adapter intentionally small: it translates input/output and calls `run_agent`; it does not own the agent's model, tools, sessions, MCP lifecycle, or business logic.

### LangGraph template

- Keeps agent construction, tools, MCP lifecycle, checkpointing, tracing, and `run_agent` in `agent/agent.py`.
- Moves Mason payload handling, event translation, result shaping, and hook implementations to `runtime/adapter.py`.
- Keeps the existing checkpoint-aware recovery semantics.

### OpenAI template

- Keeps agent construction, tools, MCP lifecycle, sessions, HITL state, tracing, and `run_agent` in `agent/agent.py`.
- Moves Mason payload handling, event translation, result shaping, and hook implementations to `runtime/adapter.py`.
- Marks replayed recovery input so the model understands the at-least-once recovery context.

### Generated guidance and UI composition

- Updates template `README.md` and `AGENTS.md` files to identify `agent/agent.py` as the normal customization point.
- Keeps `runtime/main.py` focused on server construction and hook registration while allowing custom FastAPI routes.
- Applies the same entrypoint wiring to templates composed with the default chat UI.

## External behavior and compatibility

- No Mason invocation endpoint, request, response, CLI, `agent.toml`, or Runtime Store API changes are introduced here.
- Existing applications are not rewritten. Newly generated templates use the clearer file boundary.
- The only intentional runtime behavior change is on the OpenAI recovery path: a recovery-only developer message is prepended to the replayed input.
- Related work: #589 owns the runtime/durability package boundary; #592 owns product terminology, CLI/configuration, provisioning, and `RUNTIME.md`.

## Validation

Validated on September 16, 2026 after rebasing onto current `main`:

- Full Mason suite: `502 passed`.
- OpenAI template suite: `15 passed, 1 skipped`.
- LangGraph template suite: `14 passed, 1 skipped`.
- Fresh LangGraph template with the default chat UI: `27 passed, 1 skipped`.
- Fresh OpenAI template with the default chat UI: `28 passed, 1 skipped`.
- Ruff check and `git diff --check` passed.

### Local HTTP end-to-end

Exercised both generated templates with real model calls:

- Foreground invocation returned `200`.
- Background invocation returned `202` and polling reached `completed`.
- Live SSE streaming and stored event replay both returned events.
- Reusing the same invocation ID and body returned the persisted result.
- Reusing the same invocation ID with a different body returned `409`.
- A direct OpenAI `recover(...)` real-model test observed the recovery developer instruction and returned the expected recovery-aware response.

### Deployed end-to-end

Tested against the `e2-dogfood-apps` workspace:

- LangGraph foreground, background, polling, and SSE passed.
- OpenAI foreground, background, polling, SSE, and idempotency passed.
- Invocation results and events for both templates remained available after redeploy, verifying Runtime Store reuse.

The generated projects were tested with a wheel built from this PR. An unmodified dependency resolution selected the already-published `databricks-mason==0.1.5.dev0`, which predates this branch's `start_trace` API and is therefore not a valid way to exercise the PR source.

The full stale-heartbeat takeover path was not separately fault-injected in this template PR; the framework recovery hooks and persisted result/replay behavior were exercised directly.


elainewang-db (commented):
qq: I thought `invoke` and `recover` were meant to be implemented by the agent author, but now they are moved to adapter?

shivam5:
@elainewang-db  the adapter code is also managed by agent author. but the agent code should be the unchanged agent code that the developer ports from anywhere.
The runtime code is the glue code they write to plumb with our runtime

elainewang-db:
> @elainewang-db the adapter code is also managed by agent author. but the agent code should be the unchanged agent code that the developer ports from anywhere. The runtime code is the glue code they write to plumb with our runtime

yes, that makes sense, i think the language `most applications should not edit this file.` is a bit confusing because I thought that `invoke` and `recover` were the agent author entrypoints.

elainewang-db (approved):
thanks!

PR #593, opened 2026-09-15
[mason] Provision Runtime Store through managed API

shivam5:
## Summary

Add the Conversation Store Runtime Store API as a private, internally controlled backend for Mason's durable invocation state, while keeping the existing per-app Lakebase project path as the default.

- `_USE_MANAGED_RUNTIME_STORE = False` is a private code constant. There is no CLI option or environment flag for customers.
- With the switch off, deploy continues to create/reuse the same per-app Lakebase project, attach the same `postgres-runtime-store` App resource, and configure the same endpoint/schema environment variables.
- With the switch off, deletion remains app-only and deliberately leaves the Lakebase project, matching existing behavior.
- With the switch on in a future internal release, Mason creates/reuses a Runtime Store after the App exists, validates its App/SP owner, and uses the API-returned branch, database, and username. Managed deletion removes the store before the App and retains the App if cleanup fails.
- No migration or automatic backend conversion is included.

Design: [Runtime Store API design](https://docs.google.com/document/d/14Tt7ygDTCISit_lN2vkJw6nS9h1IiaDeE_JzRofzUKU/edit?tab=t.vfhiphco0q79#heading=h.izj82tgkuzja)
Backend rollout: https://github.com/databricks-eng/universe/commit/b0e44eb9d04b2da2dce9f44b76bfe83f6712ff1b

## Validation at head `17ac8f6`

| Path | Environment | LiteSwap | Result |
| --- | --- | --- | --- |
| Legacy create/reuse + Postgres resource | `eng-ml-inference` staging | No | Pass |
| Legacy invocation persistence + redeploy readback | `eng-ml-inference` staging | No | Pass |
| Legacy full stop/start + readback | `eng-ml-inference` staging | No | Pass |
| Legacy delete | `eng-ml-inference` staging | No | Pass: App deleted; Lakebase project remained, then the test project was explicitly purged during cleanup |
| Managed canonical Get | `eng-ml-inference` staging | No | Pass |
| Managed canonical Create | `eng-ml-inference` staging | No | Reaches Conversation Store, then is rollout-blocked by `cp/conversation-store is not allowed to call RbacService.CheckPolicy` |
| Managed create/invoke/redeploy/stop/start/delete | staging | Yes, Conversation Store + Apps | Pass, including database/store cleanup |

The native managed Create failure is an Access Control rollout dependency, not a Mason failure. The private switch remains off, so it does not affect released behavior. Before enabling the switch, repeat the native managed flow after Access Control finishes rolling out and validate a non-admin/cross-manager deletion case.

Local and CI validation:

- Mason unit suite: **925 passed**
- Focused managed/legacy/runtime paths: **148 passed**
- Bridge Lakebase suite: **110 passed**
- Hermetic final-wheel functional suite: **36 passed**
- Ruff, format, type checks, and `git diff --check`: pass
- GitHub CI: **52/52 checks passed**
- Review: approved; **0 unresolved threads**

All temporary `rt593` staging Apps, Lakebase projects, workspace sources, and the LiteSwap unit were removed after validation.


integrations/mason/src/databricks_mason/cli/deploy.py
```diff
@@ -448,16 +449,49 @@ def _grant_store_access(
 def _reconcile_runtime_store(
```
elainewang-db: qq: should we introduce a `mason runtime` cli command mirroring `mason sessions / mason runtime`, and we can move some of these common methods there? can be a follow up, non blocking

shivam5: Currently runtime stores are implementation detail (of our runtime) and not exposed to the user, so I think we should not do that for now.

elainewang-db (approved):

integrations/mason/src/databricks_mason/_api_client.py
```diff
@@ -168,6 +177,44 @@ def ensure_workspace_dir(self, path: str) -> None:
...
+        )
+
+    def get_runtime_store(self, name: str) -> models.RuntimeStore:
+        """Resolve the service-managed backend and app owner before reusing a store."""
+        return _as(
+            models.RuntimeStore,
+            self._do("GET", f"{_BASE}/{runtime_store_path(name)}"),
+        )
+
+    def delete_runtime_store(self, name: str) -> dict:
```
shivam5: does delete work without a expected id?

integrations/mason/src/databricks_mason/cli/deploy.py
shivam5: Kept Runtime Store as an internal deployment detail, so I did not add a user-facing  command. The follow-up in 5a57626 separates managed and legacy provisioning into dedicated internal modules, which gives us the reusable boundary without exposing the resource.

integrations/mason/src/databricks_mason/_api_client.py
shivam5: Yes. The public DELETE request intentionally sends only . Conversation Store fetches the current metadata row and passes its persisted  to the DAO as ; that internal generation guard prevents a stale delete from removing metadata for a recreated store. Mason should not send the physical database ID.

integrations/mason/src/databricks_mason/cli/deploy.py
shivam5: Correction (inline formatting): Kept Runtime Store as an internal deployment detail, so I did not add a user-facing `mason runtime` command. Commit 5a57626 separates managed and legacy provisioning into dedicated internal modules, which gives us the reusable boundary without exposing the resource.

integrations/mason/src/databricks_mason/_api_client.py
shivam5: Correction (inline formatting): Yes. The public DELETE request intentionally sends only `runtime-stores/{runtime_store_id}`. Conversation Store fetches the current metadata row and passes its persisted `database_id` to the DAO as `expectedDatabaseId`; that internal generation guard prevents a stale delete from removing metadata for a recreated store. Mason should not send the physical database ID.

integrations/mason/src/databricks_mason/cli/deploy.py
```diff
@@ -625,13 +629,51 @@ def deploy(
...
+    runtime_manifest: dict[str, Optional[str]] = {}
+    if project is not None and project.server == AgentServer.MASON:
+        if use_managed_runtime_store:
+            app_service_principal_id = _app_service_principal(name, obj.profile)
+            with render.status("Reconciling Runtime Store…"):
+                runtime_backend = managed_runtime_store.get_or_create_backend(
+                    client, name, app_service_principal_id
+                )
+            assert app_service_principal_id is not None
+            resource_error = remove_app_resources(
```
shivam5: what is this? why do we need to remove app resources?

integrations/mason/src/databricks_mason/cli/deploy.py
```diff
@@ -625,13 +629,51 @@ def deploy(
...
+                name, {runtime_backend.resource_name}, obj.profile
             )
+            if resource_error:
+                raise AgentCliError(
+                    "Could not detach the legacy Lakebase resource for the Runtime Store.",
```
shivam5: why do we need to detach? we don't want to do any migrations.

integrations/mason/src/databricks_mason/cli/deploy.py
```diff
@@ -854,9 +896,24 @@ def deployments_stop(obj, name, yes) -> None:
 @click.option("--yes", "-y", is_flag=True, help="Skip the confirmation prompt.")
 @click.pass_obj
 def deployments_delete(obj, name, yes) -> None:
-    """Delete a deployment."""
+    """Delete a deployment and, when managed provisioning is enabled, its Runtime Store."""
```
shivam5, line 871: shouldn't we delete runtime store in both cases, because runtime store is implementation detail - who provisions and delete depends on use_managed_runtime_store (that is through api or through sdk - thats it)

integrations/mason/src/databricks_mason/runtime/durability/lakebase_runtime_store.py
```diff
@@ -125,6 +128,109 @@ def _get_token(self) -> str:
             return token
 
 
+def _managed_lakebase_connection(
```
shivam5: why is this new code?

integrations/mason/src/databricks_mason/_api_client.py
```diff
@@ -78,6 +79,14 @@ def session_store_path(name: str) -> str:
     return f"session-stores/{raw}"
 
 
+def runtime_store_path(name: str) -> str:
+    """Normalize a logical ID or resource name for the internal Runtime Store API."""
+    raw = name.removeprefix("runtime-stores/")
```
shivam5: whats this manual hardcoding? why is this needed?

integrations/mason/src/databricks_mason/app_resources.py
```diff
@@ -120,6 +121,24 @@ def apply_postgres_resources(
     return (result.stderr or result.stdout or "").strip() or "unknown error"
 
 
+def remove_app_resources(
```
shivam5: why do we need this?

integrations/mason/src/databricks_mason/lakebase_runtime_store.py
```diff
@@ -1,75 +1,118 @@
...
-    if not normalized[0].isalpha():
-        normalized = f"mason-{normalized}"
-    return f"{normalized}-runtime-store"[:63].rstrip("-")
+def validate_owner(
+    app: str, store_id: str, app_service_principal_id: str, runtime_store: Any
+) -> None:
+    """Refuse to reuse or delete a store belonging to a different app identity."""
+    if (
+        not isinstance(runtime_store, dict)
+        or runtime_store.get("name") != f"runtime-stores/{store_id}"
```
shivam5: please get rid of these. harcodings. why do we need this - we have all information in the apis

integrations/mason/src/databricks_mason/cli/deploy.py
shivam5: Agreed. This was migration cleanup, which we do not need. I removed the app-resource removal path; managed deploy leaves existing app resources untouched.

integrations/mason/src/databricks_mason/cli/deploy.py
shivam5: Agreed. I removed detachment entirely. Backend selection is now a private code-only rollout switch; this PR does not expose migration behavior or a customer-controlled flag.

integrations/mason/src/databricks_mason/cli/deploy.py
shivam5: With the internal switch off, deletion remains exactly as before: Mason deletes only the App and leaves the legacy per-app project. With it on, Mason deletes the managed Runtime Store first and retains the App on cleanup failures. I removed a proposed legacy hard purge because the deterministic project name is not sufficient proof of ownership.

integrations/mason/src/databricks_mason/runtime/durability/lakebase_runtime_store.py
shivam5: The duplicate existed because the shared connector could not accept the API-returned database and app SP. I added optional database/username support to AsyncLakebaseSQLAlchemy and removed the duplicate managed connection code; the legacy Apps-resource connector remains separate.

integrations/mason/src/databricks_mason/_api_client.py
shivam5: Agreed. I removed the custom regex/path normalizer. The validated deployment name is the Runtime Store ID, and the API client only URL-escapes it for item requests.

integrations/mason/src/databricks_mason/app_resources.py
shivam5: It was only used for migration detachment. Since this PR does not migrate existing deployments, I removed the helper and its tests.

integrations/mason/src/databricks_mason/lakebase_runtime_store.py
shivam5: Agreed. I removed the derived store ID plus the endpoint/resource/schema assumptions. Managed mode uses the deployment ID and the branch, database, and owner SP returned by the Runtime Store API; Mason keeps only its fixed internal table schema.

PR #594, opened 2026-09-15
[mason] ExtractMemories: client binding + Session.extract_memories SDK wrapper

qinxinw-db:
## Summary

Adds mason support for the new **`ExtractMemories`** RPC (`conversation-store/mason.proto`), which synchronously distills a session's transcript into memory entries and returns them.

**Client binding** (`_api_client.py`)
- **`_MasonApiClient.extract_memories(store, session_id, memory_store, *, instructions=None, dry_run=False)`** → `POST /api/2.0/agents/session-stores/{store}/sessions/{session_id}/extractions`.
  - `ExtractMemories` only binds the `/2.0/agents` path (no `/agents/v1` alias like the item ops), so a `_BASE_2_0 = "/api/2.0/agents"` base is added for it.
  - `memory_store` is normalized to its `memory-stores/<name>` resource form; `instructions` optional; `dry_run` omitted when false, sent when true.
- **`ExtractMemoriesResponse`** model (`.name` correlation id, `.entries` → list of `MemoryEntry`).

**Typed SDK wrapper** (`session_store.py`)
- **`Session.extract_memories(*, memory_store, instructions=None, dry_run=False)`** → `List[ExtractedMemory]`, delegating through `SessionStores._extract_memories`.
- **`ExtractedMemory`** — a frozen, read-only result view of a written entry (name/actor_id/path/content/description/session_id/source_type/timestamps), exported from the package. It's deliberately not the mutable `Memory` resource handle, since the entries live in the target memory store, not the session.

Not included: a `mason sessions extract-memories` CLI command (easy follow-up if wanted).

## Test plan
- [x] `pytest tests/unit_tests/` — 481 passed. Added client tests (path/body, dry-run + bare-name normalization, response wrapping) and SDK tests (typed entries + timestamp parsing, dry-run/empty-entries), and updated the public-surface test for the new export.
- [x] `ruff check` / `ruff format --check` — clean
- [x] `ty check` — clean

Note: the server RPC is `stage: DEVELOPMENT`; this works once the endpoint is live in the target workspace.


integrations/mason/src/databricks_mason/_api_client.py
```diff
@@ -590,3 +592,31 @@ def clear_session_items(self, store: str, session_id: str) -> dict:
         return self._do(
             "POST", f"{_BASE}/session-stores/{store}/sessions/{session_id}/items:clear", body={}
         )
+
+    def extract_memories(
```
shaotong-db, line 596: Do we want to add a CLI command for this too?

qinxinw-db: i believe a CLI won't be super useful -- this is mainly for agent application to call programtically at the end of each session

integrations/mason/src/databricks_mason/_api_client.py
```diff
@@ -21,6 +21,8 @@
     from databricks.sdk import WorkspaceClient
 
 _BASE = "/api/agents/v1"
+# ExtractMemories only binds the 2.0 path (no /agents/v1 alias), so it needs its own base.
+_BASE_2_0 = "/api/2.0/agents"
```
shaotong-db, line 25: nit: we probably want to use this for all API calls. We can do it in a follow-up.

shaotong-db (approved):
LGTM

PR #597, opened 2026-09-15
[mason] CLI facelift: semantic color, cargo/uv diagnostics, and a scannable --help

sirui-sun:
## What

Gives the Mason CLI a visual facelift so its output reads as scannable structure, with color reserved for meaning. Implements the design in "Mason CLI: Facelifting Design Decisions." **CLI-only** — no changes to agent runtime or app behavior.

### Color & output (`render.py`, new `theme.py`)
- Semantic palette where color carries meaning, never decoration: green = success, blue = links/URLs, cyan = the actionable command, amber = warning, red = error. No decorative brand accent.
- cargo/uv-style diagnostics: a color-coded severity keyword, a plain message, and an indented `help:` line with the fix. Only the keyword/label is styled — the message and fix stay at the terminal's default foreground so they're fully legible. Errors use cargo's `error[CODE]:` form.
- Success boxes with copy-safe next steps (cyan commands, no `$` prefix).

### `--help` (`help.py`, new `_group.py`)
- Commands grouped by intent (SETUP / DEVELOP / SHIP) with a numbered "Getting started" path, replacing the flat alphabetical dump.
- Section headings and description text recede via the terminal's adaptive dim attribute (SGR 2 — legible on light and dark backgrounds); command/option names and the auth/links footer stay at full weight. The capabilities block reads as secondary.
- Unknown commands and Click usage errors (bad option, missing/extra argument) render in the diagnostic grammar with "did you mean?" suggestions.

## Why

Mason inherited Click's default output: a flat command dump, red used as a section title rather than an error signal, dead-end errors with no suggestions, and raw pagination cursors in the human's face. This reserves color and glyphs for meaning and makes each screen scannable.

## Notes

- Honors `-o json` and `NO_COLOR` / non-TTY (colors are stripped).
- Deferred to `main`'s newer interactive pagination (kept the `--page-token` hint) rather than the earlier facelift tweak.

## Testing

- `python -m pytest tests/unit_tests/` — 492 pass. The one failure (`tool_runtime_test::…custom_headers`) is the known pre-existing failure, in `runtime/workspace.py`, unrelated to this change.
- `ruff check`, `ruff format --check`, `ty check` — all clean.


integrations/mason/src/databricks_mason/_group.py
```diff
@@ -0,0 +1,216 @@
+"""A `click.Group` subclass carrying the facelift's `--help` upgrades (design §1, §4, §6).
```
elainewang-db: nit: can we remove design doc references from the comments?

integrations/mason/src/databricks_mason/_group.py
```diff
@@ -0,0 +1,216 @@
...
+    def write_heading(self, heading: str) -> None:
+        self.write(f"{'':>{self.current_indent}}{dim(heading.upper())}\n")
+
+    def write_dl(
+        self,
+        rows: Iterable[tuple[str, str]],
+        col_max: int = 30,
+        col_spacing: int = 2,
+    ) -> None:
+        # Gray only the description column. Click ≥8.4 measures column widths and wraps with
```
elainewang-db, line 149: nit: can we bump the mason pyproject.toml click dependency to >=8.4 then too?

elainewang-db (approved):

integrations/mason/src/databricks_mason/_group.py
elainewang-db: btw my claude thinks this will fix the failing test too

integrations/mason/src/databricks_mason/_group.py
sirui-sun: Done — removed the design-doc section references (and the internal doc name/author) from the comments across theme.py, errors.py, help.py, render.py, and _group.py.

integrations/mason/src/databricks_mason/_group.py
sirui-sun: Bumped to `click>=8.4`. It's needed anyway: the --help graying styles the description column with ANSI, and 8.4+ measures/wraps columns with ANSI-aware `term_len`, so the color doesn't break table alignment. And yes — with the floor at 8.4, lowest-direct now resolves to a Click new enough that the message-format mismatch can't recur; I also kept the test assertion version-agnostic for robustness.

PR #601, opened 2026-09-16
[mason] Guide migration of existing LangGraph agents

ankit-db:
Existing LangGraph users currently have to infer how their graph should connect to Mason's server, tools, managed state, tracing, and recovery. `mason init --framework langgraph --existing .` now prepares a Claude Code migration bundle in the existing project without changing application source or configuration.

The bundle contains a focused migration skill, a ready-to-run prompt, migration settings, and a candidate project generated from the templates bundled with the installed CLI. The skill explicitly handles existing checkpoint history, custom graph state and outputs, identity/session mapping, and at-least-once recovery decisions.

Shared runtime requirements now live in `MASON_CONTRACT.md`, which is included in normal LangGraph scaffolds and migration references. The generated README and AGENTS guide link to that contract instead of repeating its details. `mason doctor` remains separate follow-up work.

Validation:

- `pytest tests/unit_tests/init_test.py -q` — 36 passed
- Mason unit suite — 516 passed; the sole failure was an existing ANSI assertion under this agent environment's `NO_COLOR=1`, and that test passes with color enabled
- `ruff check`, `ruff format --check`, and `ty check`
- Built the wheel and verified it contains the migration skill and shared contract
- Generated a real migration bundle and verified its local file/anchor links


junchoi-db (approved):

integrations/mason/src/databricks_mason/cli/init.py
```diff
@@ -101,8 +104,109 @@ def _write_env(dest: pathlib.Path, profile: str) -> bool:
...
+                "durable_runtime": durable_runtime,
+            }
+        )
+        return
+    render.success(
+        "Prepared migration instructions (agent conversion is still required)",
+        fields={"Directory": str(dest), "Skill": str(skill / "SKILL.md")},
+        next_steps=[
+            (f"cd {shlex.quote(str(dest))}", "Enter the existing project"),
+            "Open Claude Code and paste the prompt from .claude/skills/mason-migrate/PROMPT.md:",
```
junchoi-db, line 197: super nit: should we do .agent/skills for codex users?

shivam5:
This generates .claude in the same folder if I do `mason init --existing`. Should we create a new folder like we do for `mason init` commands and instantiate `.claude` in it. Will not collide with existing projects.

Also the next steps/command help is centered around claude. Similar to Jun's command can we make it provider agnostic? (say something like your coding agent (claude/codex/pi))

shivam5 (approved):

integrations/mason/src/databricks_mason/cli/init.py
ankit-db: Good call — done. The bundle now lives in a tool-neutral `mason-migrate/` directory, and both `.claude/skills/` and `.agent/skills/` get a small pointer skill that forwards to it. That way Codex and Claude Code share one reference project instead of duplicating ~30 files.

ankit-db:
@shivam5 both addressed. The bundle no longer goes into `.claude/` — it is a `mason-migrate/` folder at the project root, with only a one-file pointer skill in each agent's config dir, so nothing collides with existing agent setup. Wording is now provider-agnostic ("Open your coding agent and paste the prompt from `mason-migrate/PROMPT.md`").

I kept it inside the project rather than a separate directory so the agent can still read and edit the project it is migrating; the README notes the bundle is disposable once conversion is done.

PR #602, opened 2026-09-16
Mason: migrate all APIs to /api/2.0/agents

annzhang-db:
## Summary
- Move every Mason-owned agents API caller to the canonical `/api/2.0/agents` namespace: managed memory, session stores, sessions, items, resource clients, runtime helpers, generated UI templates, and CLI examples.
- Keep the transitional session-store create request compatible with both backends by sending `session_store_name` and `session_store_id`, and accept both legacy and canonical store response identities.
- Remove the private store-less session lookup because the canonical API requires the parent session store; Mason's public resource API and CLI already require the store.

## Why
The canonical namespace is already available on the existing backend, including production, and the pending backend contract uses it for all Mason resources. Migrating only session-store creation leaves memory and the rest of the session lifecycle on the legacy compatibility namespace.

## Rollout
1. Merge and publish a new databricks-mason release containing this change.
2. Upgrade affected consumers before rolling out the backend identity changes.
3. Retain both session-store create query parameters until older backends no longer require `session_store_name`.

This PR does not publish a package or change the package version.

## Validation
- 70 focused client and resource tests passed.
- Ruff check and format passed; changed template/helper modules passed Python syntax checks; `git diff --check` passed.
- Repository-wide scan confirms no `/api/agents/v1` references remain.
- Existing production backend: canonical session-store/session/item create, get, list, update, append, pop, and delete routes worked, and canonical managed-memory store/entry CRUD and search passed. Temporary resources were cleaned up.
- A pre-existing production consistency issue remains: a session-store metadata update reads back immediately but later reverts after session activity. It reproduced with both the previous mixed-route client and this all-canonical client, so it is not caused by the namespace migration.


annzhang-db:
Validated this PR’s proposed client (`cfb4bc0`) on **September 16, 2026 (UTC)**:

- **Current staging APIs — PASS:** session-store create/get/list/update/delete; session create/get/list/update/delete; item append/list/pop.
- **Pending backend (`59b89a9`) via staging LiteSwap — PASS:** the same lifecycle, plus memory-store and memory-entry CRUD and search.
- The client handles both legacy and canonical session-store responses while preserving compatibility for subsequent operations.
- Tests used the actual PR client. LiteSwap added only the routing header; swap-pod logs confirmed requests reached the pending backend.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

integrations/mason/README.md
```diff
@@ -392,6 +392,24 @@ sync/streaming/background transport selector is manual.
 
 ## Developing Mason
 
+### Session-store API transition
```
smurching: This doesn't seem important to put in the public README right?

integrations/mason/src/databricks_mason/_api_client.py
```diff
@@ -386,8 +385,8 @@ def create_session_store(
             models.SessionStore,
             self._do(
                 "POST",
-                f"{_BASE}/session-stores",
-                query={"session_store_name": name},
+                f"{_BASE_2_0}/session-stores",
```
smurching: Just to sanity check, is this new route rolled out everywhere including prod?

smurching (approved):
Mostly looks good! I'll approve to unblock merging but can we first just validate this updated logic works against the proposed new backend + existing prod backend?

integrations/mason/src/databricks_mason/_api_client.py
jennsun: is there a similar update we need to make here for memory-stores?

annzhang-db:
Production validation completed against ai-oss-ecosystem-integration-testing using PR head 0fd650f.

29/29 unique HTTP method/path shapes under /api/2.0/agents succeeded:

- Memory stores: create, get, list, update, delete, grant permission
- Memory entries: create, get, list, search, update, delete
- Session stores: create, get, list, update, delete, grant permission
- Sessions: create, get, list, update, delete, fork
- Session items: append, list, pop, clear
- Session extraction route

The test used live temporary resources and an existing app service principal for permission grants. Updated metadata read back correctly, and all resources were cleaned up without failures.

PR #603, opened 2026-09-16
[Mason] Support request-user auth for first-class Genie tools

junchoi-db:
> **Stacked PR** — parent: #619; parallel with #620 and #621.
> **New code in this PR:** through `7f72453` — first-class Genie user auth and consent.
> **Review only this PR's diff.** Shared OBO/runtime plumbing belongs to #618 and #619.

## Stack order

|---|---|---|---|
| #618 | `main` @ `406b652` | `cf4b3ac` | Request-scoped OBO context composed with the existing Runtime |
| #619 | #618 @ `cf4b3ac` | `490393c` | Managed-MCP user/App auth plumbing and baseline `ai-gateway` |
| #603 | #619 @ `490393c` | `7f72453` | First-class Genie user auth and `genie` scope |
| #620 | #619 @ `490393c` | `c17e7a0` | DBSQL user auth and `sql` scope |
| #621 | #619 @ `490393c` | `6e622ed` | Sandbox Volume user auth and `files` scope |

## What

- Adds `auth = "user" | "app"` to first-class Genie One and Genie Agent bindings introduced by #613.
- Resolves first-class Genie clients from the active request-auth context.
- Requests `genie` consent only when a first-class Genie binding uses user auth.
- Drains in-flight blocking Genie SDK calls before request credential cleanup, including cancellation paths.

## Scope boundary

This leaf adds Genie authentication and scope policy only. It does not add execution, background, storage, or recovery behavior.

## Why

First-class Genie supports both identities. A deployed App can query as its service principal when that principal has the required grants, or query as the requesting user without granting the App service principal access to the underlying data. The CLI therefore keeps explicit `auth = "app" | "user"` selection instead of making Genie universally OBO.

## Design

Genie leaf of the final runtime-composition contract in `2026-09-18-mason-request-user-auth-sync-execution.md`.

## Deployed E2E authentication matrix

Tested on df1 from one fresh wheel built at PR head `7f72453` (SHA-256 `944a61b88aad67df47eee791e00f22cf6a31f7d2e939363e6b1117368272319e`). Both manifests were generated through the real `mason tools add genie-agent|genie-one --auth app|user` commands.

| Authentication | Native Genie | Genie One | Identity and authorization control |
|---|---|---|---|
| App service principal | PASS | PASS | Native Genie reported App SP `78495718526781`; Genie One reported `request_user_id=null` and `app_user_id=78495718526781`; the App SP had table `SELECT` |
| Request-user OBO | PASS | PASS | Native Genie reported Jun `6357521728819386`; Genie One reported `request_user_id=6357521728819386`; the App SP was deliberately denied table access while Jun retained access |

- `6/6` scored cases passed: controls, Native Genie, and Genie One in both authentication lanes.
- Both Genie surfaces returned the exact governed table marker in both lanes.
- Freshness logs proved deployed commit `7f72453` and the exact wheel hash above.
- Ingress logs observed forwarded `ai-gateway` and `genie` scopes.
- Cleanup removed both Apps, workspace sources, Genie Space, table, Volume, and schema; shared catalog grants were unchanged.

**Conclusion:** App-SP and OBO are both supported. Keep this PR's explicit auth model and its current dependency on shared request-auth plumbing; do not convert Genie to an OBO-only standalone leaf.

## Additional proof of work

- Cumulative all-leaves gate on `c88efeb`: `1100 passed, 11 skipped`; Ruff format/lint and `ty check` passed.
- [Previous deployed CUJ capability report](https://metric-reports-viewer-6051921418418893.staging.aws.databricksapps.com/20260918_mason-request-user-auth-cuj-e2e.html) includes native Genie and Genie One positive/negative identity controls.

## Testing

- Covers user/App client resolution, `genie` scope derivation, missing auth, adapter construction, timeout, cancellation, and cleanup.


integrations/mason/src/databricks_mason/cli/app_auth.py
```diff
@@ -0,0 +1,184 @@
...
+        return
+    effective = set(app.effective_user_api_scopes or [])
+    if effective - _IDENTITY_DEFAULT_SCOPES:
+        raise AgentCliError(
+            "Apps did not return configured scopes for an App with unexplained effective grants.",
+            hint="Inspect the App's scope configuration with its owner before retrying adoption.",
+        )
+
+
+def requires_user_auth(project: AgentProject | None) -> bool:
```
junchoi-db, line 41: how do we know this in advance? which tools requre user auth?

integrations/mason/src/databricks_mason/cli/app_auth.py
```diff
@@ -0,0 +1,184 @@
...
+
+@dataclass(frozen=True)
+class AppAuthPlan:
+    apps: AppsAPI
+    name: str
+    existing_scopes: tuple[str, ...] | None
+    scopes: tuple[str, ...]
+
+
+def required_user_scopes(project: AgentProject | None) -> set[str]:
```
junchoi-db, line 81: same here -- given which tool -> how do we knwo user is the only available scopes?

integrations/mason/src/databricks_mason/langgraph/genie.py
```diff
@@ -16,7 +28,21 @@ def genie_tools() -> list[BaseTool]:
...
-        agent = GenieAgent(record.space_id or "")
+        if (
+            record.auth == "user"
+            and workspace_client_for is None
+            and os.getenv("DATABRICKS_APP_NAME")
+        ):
+            raise AuthError(
+                "MCP_USER_AUTHORIZATION_MISSING",
+                "This deployed Genie integration requires request-user authorization.",
+                integration_id=record.id,
```
junchoi-db, line 39: I don't get this 

auth is user right? so what is wrong?

integrations/mason/src/databricks_mason/langgraph/mcp.py
```diff
@@ -13,24 +13,44 @@
...
+def _server_from_tool(
+    tool: ToolRecord,
+    *,
+    workspace_client_for: Callable[[str], WorkspaceClient] | None = None,
+) -> DatabricksMCPServer | None:
     if tool.kind not in {"sandbox", "mcp", "uc_function", "genie_one"}:
         return None
-    client = workspace_client()
+    mode = tool.auth or "app"
+    if workspace_client_for is None and mode == "user" and os.getenv("DATABRICKS_APP_NAME"):
```
junchoi-db, line 42: what is relation ship between

workspace cient
env var
autho mode user?

integrations/mason/src/databricks_mason/openai/genie.py
```diff
@@ -16,7 +28,21 @@ def genie_tools() -> list[FunctionTool]:
     for record in load_tools(expected_framework="openai"):
         if record.kind != "genie_agent":
             continue
-        agent = GenieAgent(record.space_id or "")
+        if (
+            record.auth == "user"
```
junchoi-db, line 32: same here

integrations/mason/src/databricks_mason/langgraph/genie.py
junchoi-db: Yes, `auth = "user"` is the requested identity mode, but it is not itself a credential. The request path must also pass the active request-owned `RequestAuthContext.client_for` resolver into `genie_tools()`. `GenieAgent._client()` then calls that resolver with `"user"` and receives a `WorkspaceClient` configured from the forwarded user token. This guard covers the broken deployed-App case where the manifest says user but invocation plumbing did not provide the resolver: rather than silently fall back to the App SP, it fails closed. Local development retains the environment-authenticated fallback.

integrations/mason/src/databricks_mason/langgraph/mcp.py
junchoi-db: The flow is: `auth = "user"` in `agent.toml` selects the identity mode; the request-owned resolver turns that mode into the actual `WorkspaceClient`; and that client carries the forwarded user token used for the API call. `DATABRICKS_APP_NAME` is only a deployment-boundary signal set by Databricks Apps. If it is present and a user-auth binding has no request resolver, we fail closed instead of falling back to the environment/App-SP client. If `auth = "app"`, the resolver returns the environment-authenticated App client. In local development, where `DATABRICKS_APP_NAME` is absent, the normal environment-authenticated client fallback remains available.

PR #604, opened 2026-09-16
Show default experiment URL in mason dev output

jamesbxwu:
## What

With tracing on by default, `mason dev` now prints the default MLflow experiment's **Traces** URL in its startup panel — the same `Traces` line `mason deploy` already shows. Previously `mason dev` wired the experiment into env silently, so a local run gave no hint of where its traces landed.

```
Flow:
  resolve_trace_experiment_id ──► experiment_id
        ├─► env wiring (unchanged)
        └─► experiment_url(host, id) ──► trace_url ──► _announce_local_url
                                                          │
                                                 render.success(fields={..., "Traces": <url>})
```

## Details

- The URL is computed inside the existing best-effort tracing block, falling back to the bare experiment id when the host is unavailable (offline).
- Threaded into `_announce_local_url`, which adds a `Traces` field for both the chat-UI and API-only startup panels.
- When tracing is disabled, no `Traces` line is shown.

## Testing

- New unit tests: `test_dev_shows_default_experiment_url_when_tracing_on` and `test_dev_omits_experiment_url_when_tracing_off`.
- Full `dev_test.py` (32), `deploy_test.py` + `tracing_test.py` (77) pass; `ruff` + `ty` clean.
- Recently-added functional/e2e tests (`dev_runtime_test.py`, `cli_smoke_test.py`) intentionally bypass the announce/output layer, so they are unaffected.
- Reviewed with Isaac Review: 0 final findings.

image


elainewang-db (approved):
thanks! can we also change the port to all use 8001 (I can do in a follow up but if its easy here that would be great). also on mason deploy do we show the experiment link?

jamesbxwu:
> thanks! can we also change the port to all use 8001 (I can do in a follow up but if its easy here that would be great). also on mason deploy do we show the experiment link?

@elainewang-db let's do that as a quick follow up since i want to understand how this is integrated with `--app-port` as well

PR #606, opened 2026-09-16
[mason] Hide internal Runtime Store from deploy output

shivam5:
## Summary

- stop showing the internal Runtime Store database path in `mason deploy` output
- retain Runtime Store provisioning and deployment wiring
- add regression coverage for the user-facing output

## Testing

- `uv run pytest tests/unit_tests/deploy_test.py -q` (64 passed)
- `uv run ruff check src/databricks_mason/cli/deploy.py tests/unit_tests/deploy_test.py`
- `uv run ruff format --check src/databricks_mason/cli/deploy.py tests/unit_tests/deploy_test.py`


shivam5:
Folded into https://github.com/databricks/databricks-ai-bridge/pull/607

PR #608, opened 2026-09-16
[mason] deploy: preserve user_api_scopes by masking resource-grant updates (ML-69759)

elainewang-db:
`mason deploy`'s resource-grant step (`apply_postgres_resources` / `apply_experiment_resource`) called `apps update --json {"resources": …}` with no field mask. `apps update` replaces the whole app spec, so every field not in the payload — notably `user_api_scopes` — reset to the iam-only default on **every** deploy, silently breaking OBO until scopes were manually re-set.

Both grants now go through one `_update_app_resources` helper that uses `apps create-update` with `update_mask: "resources"` (the pattern the compute path already uses), so the write touches only the resources array and leaves `user_api_scopes` (and all other fields) intact.

Note: this fixes the **reset** (blocker #1). Applying a scope *change* still requires an app restart to re-mint the OBO token — that's inherent (the token is minted at app startup).

## Verification

Unit: 552 tests, ruff, ty; added `test_resource_update_is_masked_to_resources` as a regression guard.

Live check against the Databricks Apps API (e2-dogfood, disposable app), which the unit tests can't cover since they mock the CLI:

| Step | `effective_user_api_scopes` |
| --- | --- |
| Baseline | `[iam.access-control:read, iam.current-user:read]` |
| Set `user_api_scopes: ["sql"]` | `[iam.access-control:read, iam.current-user:read, sql]` |
| **Fix** — masked `create-update` (`update_mask: "resources"`) | `[…, sql]` — **preserved** ✅ |
| **Control** — old unmasked `apps update {resources}` | `[iam.access-control:read, iam.current-user:read]` — **reset** (reproduces the bug) ✅ |

Fixes ML-69759.


integrations/mason/src/databricks_mason/app_resources.py
```diff
@@ -116,10 +114,26 @@ def apply_postgres_resources(
...
-    result = _databricks(
-        ["apps", "update", app, "--json", json.dumps(payload)], profile, capture=True, check=False
-    )
+    result = _update_app_resources(app, preserved + ours, profile)
     if result.returncode == 0:
         return None
     return (result.stderr or result.stdout or "").strip() or "unknown error"
+
+
+def _update_app_resources(
```
junchoi-db, line 123: qq: what are resouces we currently pass here ?

junchoi-db (approved):
LGTM % few questions

integrations/mason/src/databricks_mason/app_resources.py
junchoi-db: would this for memory/session store?

I wonder if this would apply for tools
* mcp
* uc function  etc

elainewang-db: memory and session stores go through the conversation store API. MCP and uc functions aren't app resources, so this shouldn't apply to them either. the resources here are just the runtime durability store and the trace experiment

PR #609, opened 2026-09-17
fix(mason): validate MCP services before adding tool bindings

junchoi-db:
## What
Validate managed MCP services before `mason tools add mcp` writes `agent.toml`.

## Why
The command previously validated only the three-part name, so a nonexistent service was reported as successfully added and failed later at runtime.

The CLI now performs a read-only Unity Catalog service lookup using the selected profile. Lookup failures preserve the project files and surface the original error code with an actionable hint. Local project/name/conflict checks still run first; listing and removing bindings remain offline. Metadata validation does not guarantee runtime EXECUTE permissions or invoke any tool.

## Design doc
n/a — focused bug fix.

## Testing
- 76 focused unit and installed-CLI functional tests passed.
- Regression tests first failed on the old implementation: nonexistent services and lookup errors incorrectly succeeded.
- Changed-file Ruff lint, formatting and type checks passed.
- Added opt-in live CLI E2E coverage for LangGraph and OpenAI: reject nonexistent services without writes, add valid services, repeat idempotently, list, and remove.
- Updated the existing full tool matrix to expect rejection rather than successful registration of an unavailable service.

## Proof of work
- **Live built-wheel CLI E2E: 2 passed, 0 skipped** (LangGraph and OpenAI). Real workspace API, no mocked responses; project files are compared byte-for-byte on rejection. No tools were invoked or remote resources created.
- **Installed-wheel, workspace-free functional smoke tests: 8 passed.**
- **Full Mason unit suite: 608 passed, no exclusions or skips.** Color tests ran with `TERM=xterm` and `NO_COLOR` unset.
- The existing full matrix's updated CLI-authoring step also passed against the live workspace, including the explicit profile and `NOT_FOUND` assertion. The full deployment/runtime matrix was not run for this metadata-only CLI fix.
- Built artifact: `databricks_mason-0.1.6.dev0-py3-none-any.whl`, SHA-256 `fce2a16b6937fc62ed3d034b5b7d9ff84181af12d0132be202f6cbd1abc7e9e1`.

Example from the live run (profile and temporary project path omitted):

```console
$ mason --profile <profile> --output json tools add mcp system.ai.missing_service --source <project>
exit=1
{
  "error": {
    "message": "Could not validate MCP service 'system.ai.missing_service': Resource not found",
    "code": "NOT_FOUND",
    "hint": "Run `mason mcp list --schema catalog.schema` with the same --profile to find available services."
  }
}
```

Adding `system.ai.web_search` returned `changed: true`; adding it again returned `changed: false`, with only `agent.toml` changed by the first add.

## CI compatibility repair
Upstream `main` at `14e3bcd` already had broken CLI tests: [baseline CI run](https://github.com/databricks/databricks-ai-bridge/actions/runs/35158817235). A separate follow-up commit repairs those tests so this PR can run the complete suite:

- Update memory/session CLI imports to `databricks_mason.cli.*` and supply the required `server="mason"` fixture argument.
- Remove obsolete provisioning mocks: binding is now local-only, and the test context explicitly rejects workspace access.
- Update the removed `--no-create-stores` option test to assert the actual unknown-option error and an unchanged manifest.

Repository-wide Ruff, full-package type checking with all extras, and all 608 unit tests pass locally. **All 52 GitHub checks pass** on `1fb3006`: [tests](https://github.com/databricks/databricks-ai-bridge/actions/runs/35177680492), [lint and type checks](https://github.com/databricks/databricks-ai-bridge/actions/runs/35177680487). No tests are skipped or disabled.


integrations/mason/src/databricks_mason/_api_client.py
```diff
@@ -195,6 +196,10 @@ def _do(
 
     # --- Unity Catalog MCP Services -----------------------------------------
 
+    def get_mcp_service(self, service: str) -> dict:
+        """Look up a managed MCP service visible to the authenticated user."""
+        return self._do("GET", f"{_MCP_SERVICES_PATH}/{quote(service, safe='')}")
```
junchoi-db, line 201: _MCP_SERVICES_PATH -- why is this hard coded?

integrations/mason/src/databricks_mason/cli/tools.py
```diff
@@ -75,6 +75,18 @@ def _add_spec(obj: Any, source: pathlib.Path, spec: ToolSpec) -> None:
...
+    if spec.source.kind == "mcp":
+        service = spec.source.service
+        try:
+            obj.client().get_mcp_service(service)
+        except AgentCliError as exc:
+            raise AgentCliError(
+                f"Could not validate MCP service {service!r}: {exc.message}",
+                error_code=exc.error_code,
+                hint=exc.hint
+                or "Run `mason mcp list --schema catalog.schema` with the same --profile "
```
elainewang-db: can we update this help message with `mason tools list --kind mcp`?

integrations/mason/src/databricks_mason/cli/tools.py
```diff
@@ -179,7 +191,11 @@ def add_mcp(
     tool_id: str | None,
     source: pathlib.Path,
 ) -> None:
-    """Add a Databricks-managed MCP service as a tool (see `mason mcp list` for available services)."""
+    """Validate and add a Databricks-managed MCP service as a tool.
+
+    Checks the service in the selected workspace before changing agent.toml.
+    See `mason mcp list` for available services.
```
elainewang-db: same here

elainewang-db (approved):

PR #611, opened 2026-09-17
[mason] Facelift the prebuilt agent chat UI

sirui-sun:
Redesign Mason's bundled chat UI (both the langgraph and openai templates) to match the "Agent Workbench" design prototype, and add a memory browser and a tracing card. The UI stays zero-build vanilla HTML/CSS/JS served by the existing FastAPI runtime.

Visual + layout:
- Cool lavender shell with white card panels, a Google-blue accent, and the Mason logo mark in the header.
- Chat: user messages as bubbles, agent responses as plain text with a name + timestamp header and a hover "Copy" action.
- Composer: model picker beside a circular Send button (disabled until there is text).
- Left rail: a single card divided by lines, with click-to-expand capabilities (outline icons) and a searchable Sessions list.
- Right rail: a full-height Events panel with a Logs/List view toggle, filter pills, a payload detail pane, and JSON syntax highlighting.

Memory browser:
- Clicking the Memory capability opens a pane in the left rail (store name + resource, a labeled actor-ID filter, a search box, an entry count, and cards with path/preview/actor/date). Clicking a card opens a modal with the full entry. The memory list/search endpoints accept an optional actor override (defaulting to the viewer).

Traces:
- A Traces capability reports whether MLflow tracing is configured and, when on, links to the experiment; the config route now returns a `tracing` object (enabled + best-effort experiment URL).

Both templates' UI assets stay in sync (styles.css is identical; index.html/app.js differ only in the framework label and the framework-specific session-history handling). Cached asset links are versioned so a redeploy never serves a stale stylesheet. CHAT_APP.md is updated to match, and the demo-UI unit tests cover the actor override and the tracing config.


integrations/mason/src/databricks_mason/templates/ui/agent-langgraph/runtime/ui.py
```diff
@@ -443,17 +488,20 @@ async def create_memory_entry(request: Request, payload: MemoryEntryRequest) ->
     async def list_memory_entries(
         request: Request,
         path_prefix: str | None = Query(default=None),
+        actor: str | None = Query(default=None),
     ) -> dict:
+        # The UI can browse another actor's memories by passing ?actor=; default to the viewer.
         _require_memory()
         return await _managed_call(
-            _state_client().list_memory_entries, _request_actor(request), path_prefix
+            _state_client().list_memory_entries, actor or _request_actor(request), path_prefix
```
shivam5, line 496: will this allow user A to read user B's memories if they supplied different actor? do we want that?

shivam5 (approved):
looks great, left a minor comment regarding querying memory

integrations/mason/src/databricks_mason/templates/ui/agent-langgraph/runtime/ui.py
sirui-sun: I think for a toy implementation like this, this is fine. Generally actors is anyways not a security boundary. cc: @nihit

PR #615, opened 2026-09-17
mason: give scaffolded default store names a unique token

elainewang-db:
## What

When you scaffold an agent with `mason init`, it now creates its own uniquely-named memory and session stores by default (unless you pass `--memory-store` / `--session-store`).

Previously the default names came only from the directory (`<dir>-memory` / `<dir>-session`). Display names aren't unique, so two `my-agent` projects would get the same store name and could bind to each other's store. Now each scaffold gets a short random token shared by its two stores:

my-agent-abcxyz-memory
my-agent-abcxyz-sessions

The store kind (`memory` / `sessions`) stays at the end, and names never start or end with a digit.


integrations/mason/src/databricks_mason/agent_project.py
```diff
@@ -199,29 +185,25 @@ def uc_function(cls, tool_id: str, *, function: str) -> "ToolSpec":
...
-    reduces to nothing (e.g. a directory of only punctuation) falls back to ``agent``.
+    Sanitized to the store display-name charset (lower-case alphanumerics and hyphens). Store names
+    must not start with a digit, so leading digits/hyphens are dropped; a name that reduces to
+    nothing (only punctuation, or all-numeric) falls back to ``agent``. An optional per-scaffold
+    ``token`` is inserted before the suffix (``my-agent-<token>-memory``) so fresh scaffolds get
+    distinct stores while the name still ends with the store kind.
     """
-    slug = re.sub(r"[^a-z0-9-]+", "-", project_name.lower()).strip("-") or "agent"
-    return f"{slug}-{suffix}"
+    slug = re.sub(r"[^a-z0-9-]+", "-", project_name.lower()).strip("-").lstrip("0123456789-")
```
jennsun: Store names should be able to start with a digit now since we are deterministically prefixing session and memory stores with "session-" and "memory-" respectively. therefore the "must not start with a digit" do longer needs to be enforced on the project name. can we update this regex/documentation around this to be more flexible as a result? should be something like:
```
slug = re.sub(r"[^a-z0-9-]+", "-", project_name.lower()).strip("-")
```

integrations/mason/src/databricks_mason/cli/init.py
jennsun, line 1: Non-blocking naming/observability suggestion: could we include the chosen memory/session names in both the text and JSON output from `mason init`? With the new random suffix, neither name can be inferred from the directory, and neither appears in the success output today.

Could we also update the two option help strings and `integrations/mason/README.md` (the default-store paragraph)? They still describe `<dir>-memory` / `<dir>-session`, while this now produces `<dir>-<token>-memory` / `<dir>-<token>-sessions`.

integrations/mason/tests/unit_tests/agent_project_test.py
```diff
@@ -398,10 +242,23 @@ def test_required_project_selections_name_agent_manifest(
 
 @pytest.mark.parametrize(
     ("raw", "expected"),
-    [("My_Agent", "my-agent-memory"), ("a.b c", "a-b-c-memory"), ("___", "agent-memory")],
+    [
+        ("My_Agent", "my-agent-memory"),
+        ("a.b c", "a-b-c-memory"),
+        ("___", "agent-memory"),
+        ("2048-game", "game-memory"),  # leading digits dropped: names must not start with a digit
```
jennsun: Update 2048 test and make sure stores starting with digits can successfully be created - merged PRs for context: 
- memory: https://github.com/databricks-eng/universe/pull/2629585
- session: https://github.com/databricks-eng/universe/pull/2629587

integrations/mason/tests/unit_tests/init_test.py
```diff
@@ -27,6 +28,20 @@ def __init__(self, output: str = "text", profile=None):
...
+def _pop_default_stores(manifest: dict, slug: str = "proj") -> dict:
+    """Pop the scaffold's default store tables, asserting each is `<slug>-<token>-<kind>`.
+
+    Default names carry a per-scaffold random token so fresh scaffolds don't collide, so they can't
+    be compared literally. Check the shape and that both stores share the one token, then return the
+    manifest without them for an exact-equality check on the rest.
+    """
+    mem = re.fullmatch(rf"{slug}-([a-z]{{6}})-memory", manifest.pop("memory_store")["name"])
+    sess = re.fullmatch(rf"{slug}-([a-z]{{6}})-sessions", manifest.pop("session_store")["name"])
+    assert mem and sess, "default store names must be <slug>-<kind>-<token>"
```
jennsun: this says "<slug>-<kind>-<token>", but shouldn't it be `<slug>-<token>-<kind>`?

integrations/mason/tests/unit_tests/init_test.py
```diff
@@ -27,6 +28,20 @@ def __init__(self, output: str = "text", profile=None):
...
+    assert mem.group(1) == sess.group(1), "memory and session stores must share the scaffold token"
+    return manifest
+
```
jennsun, line 45: can we avoid mutating the manifest `pop` here? Can we do something like this to validate the random portion while still asserting over the generated manifest:
```
def _default_store_token(manifest: dict, slug: str = "proj") -> str:
    mem = re.fullmatch(
        rf"{re.escape(slug)}-([a-z]{{6}})-memory",
        manifest["memory_store"]["name"],
    )
    sess = re.fullmatch(
        rf"{re.escape(slug)}-([a-z]{{6}})-sessions",
        manifest["session_store"]["name"],
    )
    assert mem and sess
    assert mem.group(1) == sess.group(1)
    return mem.group(1)
```
then
```
token = _default_store_token(manifest)
assert manifest == {
    "schema_version": 1,
    "agent": {...},
    "memory_store": {"name": f"proj-{token}-memory"},
    "session_store": {"name": f"proj-{token}-sessions"},
}
```

jennsun (changes requested):
left some feedback to make sure the updated scaffolding for store names matches the actual session/memory store creation behavior!

integrations/mason/README.md
```diff
@@ -81,8 +81,8 @@ stores declared in `agent.toml`, and grants the app's service principal access t
 deployments list` shows what you have deployed, and `mason deployments get my-agent` prints its
 URL and status.
 
-`mason init` declares default memory and session stores in `agent.toml` (named `<name>-memory` and
-`<name>-session`), so the deployed agent has long-term memory and durable conversation history —
+`mason init` declares default memory and session stores in `agent.toml`, so the deployed agent has
+long-term memory and durable conversation history —
 `mason deploy` creates them if they don't exist yet. Point the agent at stores you already have with
```
jennsun, line 87: Now that we've removed the names here, could we document the naming shape and explain how to find it? For example
```
mason init creates <name>-<6-letter-token>-memory and <name>-<6-letter-token>-sessions, and records both names in agent.toml.
```

jennsun (approved):
had one small doc update in the README, thanks for addressing other feedback and adding the memory/session store names in the output of mason init! this will be super helpful :)

PR #618, opened 2026-09-18
[Mason] Add request-scoped OBO auth to agent runtime

junchoi-db:
> **Stack root** — review and merge this PR first.
> **Current head:** `02a14f6` — rebased onto `main` at `de096fb`, composing transient request auth with Mason's existing Runtime and recognizing the real `mason dev` local-runtime marker.
> Capability bindings and OAuth scope policy remain in child PRs.

## Stack order

|---|---|---|---|
| #618 | `main` @ `de096fb` | `02a14f6` | Request-scoped OBO context composed with the existing Runtime |
| #619 | #618 @ `010bf6a` | `70d4ec2` | Managed-MCP user/App auth plumbing and baseline `ai-gateway` |
| #603 | #619 @ `70d4ec2` | `4f915a0` | First-class Genie user auth and `genie` scope |
| #620 | #619 @ `70d4ec2` | `0bd3967` | DBSQL user auth and `sql` scope |
| #621 | #619 @ `70d4ec2` | `72eb002` | Sandbox Volume user auth and `files` scope |

#603, #620, and #621 are parallel leaves.

This update rebases only the stack root; descendant rows retain their pre-rebase parent SHAs until those PRs are rebased separately.

## What

- Creates a non-serializable `RequestAuthContext` from trusted Databricks Apps forwarded-user headers.
- Composes that transient context with Mason's existing synchronous, streaming, and background Runtime paths.
- Passes the live context to the first execution attempt through `InvocationContext`, then closes and clears its credential material.
- Namespaces internal invocation and session identifiers by App and user while preserving public API identifiers.
- Rejects a recovered request-user attempt with `MCP_USER_AUTH_RECOVERY_UNSUPPORTED` before agent code runs.
- Treats `DATABRICKS_MASON_RUNTIME_STORE_LOCAL=true` as local development even though `mason dev` also sets `DATABRICKS_APP_NAME`.

## Scope boundary

This PR does **not** implement a background subsystem, Runtime Store, saver, checkpointer, scheduler, or second execution path. Those Runtime capabilities already exist on `main` (the durability boundary landed in #589 and managed Runtime Store provisioning in #593). OBO only supplies transient request credentials to the existing Runtime's active first attempt.

Invocation inputs, status, events, and results may use the existing Runtime Store. The bearer token, `RequestAuthContext`, and credentialed clients never do. Failure recovery cannot reconstruct the original forwarded credential, so a recovered request-user attempt fails closed.

## Why

Managed tools need the requesting user's Databricks identity without serializing or persisting bearer credentials. Keeping authentication as an attempt-local concern preserves the Runtime's existing execution modes and avoids creating a second runtime solely for OBO.

## Design

Final runtime-composition contract for `2026-09-18-mason-request-user-auth-sync-execution.md`.

## Proof of work

- Post-rebase focused auth/runtime suite, including the real scaffolded `start-server` / Mason-dev path: `49 passed`.
- Changed-file Ruff and format checks pass; Mason's exact CI typecheck command passes with all extras.
- GitHub Actions on `02a14f6`: all lint, typecheck, package, and Mason-dev functional checks passed on Linux and macOS.
- [PR 618 OBO runtime-mode E2E report](https://metric-reports-viewer-6051921418418893.staging.aws.databricksapps.com/20260921_mason-pr618-obo-runtime-modes-e2e.html): `5/5` expected outcomes across real `mason dev` and `mason deploy`.
  - Foreground OBO: success in dev and deploy.
  - Direct-background OBO: accepted, status/events readable, and completed in dev and deploy.
  - Crash recovery: persisted work reached replacement attempt 2 and failed closed before the recovery handler with `MCP_USER_AUTH_RECOVERY_UNSUPPORTED`.
- Pre-rebase E2E wheel SHA-256: `d9ccf4aeaa76d5fef56a6c6b245b2876ad3ec2e160a54ad4bea7fb523945ded4`; packaged auth/runtime modules matched the reviewed `010bf6a` head byte-for-byte.
- [Previous deployed capability report](https://metric-reports-viewer-6051921418418893.staging.aws.databricksapps.com/20260918_mason-request-user-auth-cuj-e2e.html) covers the broader user-vs-App identity and scope matrix.

## Testing

- Verifies sync, streaming, and background first attempts receive isolated request auth.
- Verifies status/events/results can be retained without persisting the token.
- Verifies success, failure, cancellation, and shutdown close credential material.
- Verifies recovery fails before either invocation handler executes.
- Verifies the actual `mason dev` manifest works when Apps CLI sets both the App name and local-runtime marker.
- Preserves existing App-auth Runtime behavior.


integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -56,10 +57,17 @@ def __init__(
...
+        auth_policy: InvocationAuthPolicy | None = None,
     ) -> None:
+        self.auth_policy = auth_policy or InvocationAuthPolicy()
         self._invoke_hook: InvocationHook | None = None
         self._recovery_hook: InvocationHook | None = None
-        if runtime_store is None:
+        self._runtime: Runtime | None
+        if self.auth_policy.requires_user:
+            if runtime_store is not None:
+                raise ValueError("Request-user execution cannot use an explicit Runtime Store")
```
shivam5: lets not say request user execution cannot user an explicit runtime store.
lets say something like - "request-user execution (or OBO?) do not survive failure recovery yet."

shivam5: Also to support that, there has been some discussion which we can revive - https://databricks.slack.com/archives/C0BL69HA4DQ/p1787592444426839

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -56,10 +57,17 @@ def __init__(
...
         runtime_store: RuntimeStore | None = None,
+        auth_policy: InvocationAuthPolicy | None = None,
     ) -> None:
+        self.auth_policy = auth_policy or InvocationAuthPolicy()
         self._invoke_hook: InvocationHook | None = None
         self._recovery_hook: InvocationHook | None = None
-        if runtime_store is None:
+        self._runtime: Runtime | None
+        if self.auth_policy.requires_user:
+            if runtime_store is not None:
```
shivam5: why is there a check on runtime_store?

If we want to check whether failure recovery is started or not, the hook should be `self._recovery_hook is None`
Runtime store does more than just failure recovery.

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -145,13 +160,126 @@ async def _execute(
             raise RuntimeError(f"no {handler} handler is registered")
         return await function(copy.deepcopy(invocation_request["input"]), context)
 
+    async def _invoke_request_user(
+        self, request: Request, body: _InvocationRequest, invocation_id: str
+    ) -> Response:
+        if body.background:
```
shivam5: why is there any check on background?
the restart of execution after a crash, is agnostic whether the invocation was background or not

junchoi-db: @shivam5 

If we assume that background mode is direclty triggered from user execution -- when it is safe to assume OBO would work

I think issue is when there is crash recovery -> resumed execution does not have user context

integrations/mason/src/databricks_mason/runtime/app.py
junchoi-db: Addressed in `cf4b3ac`. Request-user/OBO execution can use either an explicit Runtime Store or the environment-selected store. The failure now reads `Request-user execution does not survive failure recovery yet`, which describes the actual boundary: invocation state may be stored, but the forwarded credential is transient and cannot be reconstructed for a recovered attempt.

integrations/mason/src/databricks_mason/runtime/app.py
junchoi-db: Addressed in `cf4b3ac`. I removed the `runtime_store` restriction entirely. `AgentApp` now always composes request auth with the existing Runtime, using either `Runtime.from_store(...)` or `Runtime.from_environment(...)`; recovery enablement remains driven by the existing recovery-hook callback. The OBO-specific check is now at attempt execution, where `attempt_context.is_recovery` fails before agent code because the transient credential cannot survive recovery.

integrations/mason/src/databricks_mason/runtime/app.py
junchoi-db: Addressed in `cf4b3ac`. I removed the request-user background rejection. Sync, streaming, and background first attempts now all use the existing Runtime; OBO only hands the transient request auth to the active first attempt. A recovered attempt fails with `MCP_USER_AUTH_RECOVERY_UNSUPPORTED` regardless of the original request mode. Added focused coverage for background composition and recovery-before-handler; the current auth gate is `42 passed` with Ruff clean.

integrations/mason/src/databricks_mason/runtime/auth.py
```diff
@@ -0,0 +1,130 @@
...
+        self._host = os.getenv("DATABRICKS_HOST", "").rstrip("/")
+        self._app = os.getenv("DATABRICKS_APP_NAME", "local")
+        self._workspace = os.getenv("DATABRICKS_WORKSPACE_ID", "")
+        self._local = local
+        self._clients: dict[str, WorkspaceClient] = {}
+        self._closed = False
+
+    @classmethod
+    def from_headers(cls, headers: Mapping[str, str]) -> RequestAuthContext:
+        if not os.getenv("DATABRICKS_APP_NAME"):
```
shivam5: Are we using databricks app name to check if the deployment is local/remote?

I think mason dev sets this as well:
DATABRICKS_APP_NAME
DATABRICKS_MASON_RUNTIME_STORE_LOCAL=true

Therefore auth thinks it is deployed and requires x-forwarded-access-token. Local requests do not have that header, so every OBO invocation returns 401 before the handler runs.

Can we test this?

shivam5 (commented):
Does a disconnected foreground request lose credentials?
Execution and http request are separate for us.

Can we test all these scenarios e2e?

shivam5:
I think this might also fail with horizontal scaling.
1. Request arrives at instance A.
2. A keeps the token in its local _request_auth dictionary and writes a credential-free invocation to the Runtime Store.
3. Every app instance scans that shared store for queued work.
4. Instance B can win the database claim and execute attempt 1.
5. B has no token - its _request_auth dictionary is separate - so the invocation fails

Proposed flow:
1. Instance A receives the request and stores the token only in memory.
2. In one database transaction, A creates the invocation directly as ACTIVE, attempt 1 - never globally visible as QUEUED.
3. A schedules that already - claimed attempt locally.
4. Other instances see ACTIVE and cannot claim it.
5. If A crashes, normal stale-attempt recovery occurs and fails closed, as intended, because the token was never persisted.

Let's actually think if it just makes sense to store the user credentials in our runtime instead of doing all of the different hacks, and restricting and re-execution.

shivam5:
Adding discussion thread on the same issue with apps team for reference - https://databricks.slack.com/archives/C0BL69HA4DQ/p1790027498055889?thread_ts=1787592444.426839&cid=C0BL69HA4DQ

shivam5:
Adding the test matrix that we discussed (if we can ask the agent to test, and we can verify):

All invocation modes (foreground, streaming, background, background+streaming) * mason dev/deploy * user cred change (this pr) / no user cred * failure_recovery added/not * horizontal scaled app instance / not
All test cases should either work, or return expected errors.

Sorry for the the exhaustive test list ask, but might be required since the auth changes is affecting all different invocation ways permutations.

integrations/mason/src/databricks_mason/runtime/app.py
```diff
@@ -132,57 +138,113 @@ async def _execute(
...
-            output = await self._runtime.invoke(invocation_id, invocation_request)
+            output = await self._runtime.invoke(runtime_invocation_id, invocation_request)
             return JSONResponse({"id": invocation_id, "status": "completed", "output": output})
         except InvocationConflictError as exc:
             raise HTTPException(409, "id was already used for another request") from exc
         except InvocationFailedError as exc:
             raise HTTPException(500, "agent invocation failed") from exc
+        finally:
+            if registered_auth and not execution_owns_auth and request_auth is not None:
+                self._discard_request_auth(runtime_invocation_id, request_auth)
```
junchoi-db, line 225: cc: @shivam5

integrations/mason/src/databricks_mason/runtime/auth.py
```diff
@@ -0,0 +1,133 @@
...
+    def namespace(self, kind: str, value: str) -> str:
+        material = json.dumps(
+            [self._host, self._workspace, self._app, self._principal, kind, value],
+            separators=(",", ":"),
+        )
+        return hashlib.sha256(material.encode()).hexdigest()
+
+    def client_for(self, mode: str) -> WorkspaceClient:
+        if self._closed:
+            raise AuthError(
```
junchoi-db, line 104: cc: @shivam5

shivam5 (approved):

PR #619, opened 2026-09-18
[Mason] Add explicit user and app auth for managed MCP tools

junchoi-db:
> **Stacked PR** — parent: #618 (review and merge that first).
> **New code in this PR:** through `490393c` — managed-MCP identity selection and deployment auth plumbing.
> **Review only this PR's diff.** Runtime/OBO composition belongs to #618.

## Stack order

|---|---|---|---|
| #618 | `main` @ `406b652` | `cf4b3ac` | Request-scoped OBO context composed with the existing Runtime |
| #619 | #618 @ `cf4b3ac` | `490393c` | Managed-MCP user/App auth plumbing and baseline `ai-gateway` |
| #603 | #619 @ `490393c` | `7f72453` | First-class Genie user auth and `genie` scope |
| #620 | #619 @ `490393c` | `c17e7a0` | DBSQL user auth and `sql` scope |
| #621 | #619 @ `490393c` | `6e622ed` | Sandbox Volume user auth and `files` scope |

#603, #620, and #621 are parallel leaves. The all-leaves validation branch is `c88efeb`.

## What

- Adds explicit `auth = "user" | "app"` selection for managed MCP bindings.
- Builds LangGraph and OpenAI MCP clients from the active invocation's client resolver.
- Requests baseline `ai-gateway` consent only when a managed MCP binding uses user auth.
- Reconciles Databricks Apps OAuth scopes without removing unrelated existing scopes.
- Keeps individual MCP integration failures isolated for App-auth and request-user tools.

## Scope boundary

This PR adds tool-auth plumbing only. It does not implement sync, streaming, background, storage, or recovery. All execution modes come from the existing Runtime; #618 only injects transient OBO auth into its active first attempt.

## Why

Application authors must choose whether each tool runs as the requesting user or the App service principal. Centralizing that choice here lets capability leaves add only their required OAuth scopes.

## Design

Managed-tool and CLI layer for the final runtime-composition contract in `2026-09-18-mason-request-user-auth-sync-execution.md`.

## Proof of work

- Cumulative all-leaves gate on `c88efeb`: `1100 passed, 11 skipped`; Ruff format/lint and `ty check` passed.
- [Previous deployed CUJ capability report](https://metric-reports-viewer-6051921418418893.staging.aws.databricksapps.com/20260918_mason-request-user-auth-cuj-e2e.html) covers managed MCP user/App identity controls and provider consent.

## Testing

- Covers explicit and mixed auth configuration, missing runtime auth, CLI rendering, scope reconciliation, templates, and App-auth fault isolation.


integrations/mason/src/databricks_mason/cli/deploy.py
```diff
@@ -517,9 +529,27 @@ def deploy(
     project = _load_project(source_dir)
     if project is not None and project.tools:
         require_managed_tool_support(source_dir)
+    user_auth = requires_user_auth(project)
     base_name = _resolve_deployment_name(project, name)
     name = _prefixed_name(base_name)
     _validate_deployment_name(name)
+    if adopt_user_auth and not user_auth:
+        raise AgentCliError("--adopt-user-auth requires explicit user-auth tools in agent.toml.")
```
junchoi-db: fyi what is the default in cli?

integrations/mason/src/databricks_mason/cli/app_auth.py
```diff
@@ -0,0 +1,171 @@
...
+        if plan.existing_scopes is None:
+            plan.apps.create(desired)
+        else:
+            current = plan.apps.get(plan.name)
+            _validate_forwarding(current)
+            _validate_implicit_identity_scopes(current)
+            if tuple(sorted(set(current.user_api_scopes or []))) != plan.existing_scopes:
+                raise AgentCliError("Apps user scopes changed since preflight; review and retry.")
+            if plan.scopes != plan.existing_scopes or instances is not None:
+                plan.apps.create_update(plan.name, update_mask=",".join(mask), app=desired).result(
```
junchoi-db, line 160: does this invoke api/rpc?

integrations/mason/src/databricks_mason/cli/app_auth.py
```diff
@@ -0,0 +1,171 @@
...
+            desired_scopes = set(plan.scopes)
+            effective = set(current.effective_user_api_scopes or [])
+            if (
+                set(current.user_api_scopes or []) == desired_scopes
+                and desired_scopes.issubset(effective)
+                and effective.issubset(desired_scopes | _IDENTITY_DEFAULT_SCOPES)
+            ):
+                return
+            if attempt + 1 < attempts:
+                time.sleep(5)
```
junchoi-db, line 175: why is this needed

integrations/mason/src/databricks_mason/cli/app_auth.py
```diff
@@ -0,0 +1,171 @@
...
+                "Request-auth contract migration requires explicit auth on every managed "
+                f"MCP/sandbox binding: {', '.join(unspecified)}.",
+                hint="Choose auth = 'app' to preserve legacy identity, or explicitly choose 'user'.",
+            )
+    return user_auth
+
+
+@dataclass(frozen=True)
+class AppAuthPlan:
+    apps: AppsAPI
```
junchoi-db, line 67: explain thi apps api

integrations/mason/src/databricks_mason/templates/agent-langgraph/runtime/adapter.py
```diff
@@ -59,32 +68,58 @@ async def recover(value: Any, context: InvocationContext) -> dict:
...
+            )
         await context.emit(event)
         if event.get("type") in ("message", "interrupt"):
             outputs.append(event)
 
     interrupted = bool(outputs and outputs[-1].get("type") == "interrupt")
     return {
         "output": [event["message"] if event["type"] == "message" else event for event in outputs],
-        "session_id": session_id,
+        **({"session_id": session_id} if not user_auth or payload.get("session_id") else {}),
```
junchoi-db, line 122: how is user auth reslated to session id injection?

integrations/mason/src/databricks_mason/cli/deploy.py
junchoi-db: The default is false: Click sets `adopt_user_auth=False` unless `--adopt-user-auth` is passed. The flag is only required when an existing App must adopt Mason-managed user scopes. A new App with user-auth tools gets its required scopes during creation. Separately, `mason tools add mcp/sandbox` defaults the tool identity to `auth = "user"`.

integrations/mason/src/databricks_mason/cli/app_auth.py
junchoi-db: Yes. `AppsAPI.create_update(...)` invokes the Databricks Apps update API through the generated SDK client, with an update mask for `user_api_scopes` and optionally the compute instance fields. It returns a long-running-operation waiter; `.result(timeout=...)` waits for that update operation to finish before scope verification begins.

integrations/mason/src/databricks_mason/cli/app_auth.py
junchoi-db: The Apps update operation can finish before `effective_user_api_scopes` reflects the new consent configuration. This loop re-reads the App up to 12 times and sleeps five seconds between attempts so we do not busy-poll while control-plane propagation converges. If configured and effective scopes still do not match, deployment stops before source rollout.

integrations/mason/src/databricks_mason/cli/app_auth.py
junchoi-db: `AppsAPI` is the generated Databricks SDK service client exposed by `WorkspaceClient(profile=...).apps`; it is not an App resource object. `AppAuthPlan` carries that authenticated API handle, the preflight snapshot of existing scopes, and the desired scope union into the apply phase. The handle is then used for `get`, `create`, and `create_update` calls with the same selected workspace profile.

integrations/mason/src/databricks_mason/templates/agent-langgraph/runtime/adapter.py
junchoi-db: This is not general session-ID injection; it is request-user isolation. Two users may send the same public `session_id` or actor, so an explicitly supplied ID is namespaced by the authenticated principal before checkpoint/session-store lookup. The Runtime-generated `context.session_id` is already namespaced in `AgentApp`, so the adapter only re-namespaces a payload-provided ID. The response returns the caller-facing ID and never exposes the internal namespaced value.

integrations/mason/src/databricks_mason/cli/app_auth.py
```diff
@@ -0,0 +1,171 @@
...
+
+from databricks_mason.agent_project import AgentProject
+from databricks_mason.errors import AgentCliError
+from databricks_mason.project_config import load_project_metadata
+from databricks_mason.project_types import AgentServer
+
+_IDENTITY_DEFAULT_SCOPES = frozenset({"iam.access-control:read", "iam.current-user:read"})
+
+
+def _validate_forwarding(app: App) -> None:
```
junchoi-db, line 20: qq: `Enable user access token forwarding` -> can we enable this by default in `mason deploy`?

integrations/mason/src/databricks_mason/cli/app_auth.py
```diff
@@ -0,0 +1,171 @@
...
+        return False
+    managed = [tool for tool in project.tools if tool.source.kind in ("mcp", "sandbox")]
+    user_auth = any(tool.auth == "user" for tool in managed)
+    metadata = None
+    if (project.root / ".mason/project.toml").is_file():
+        metadata = load_project_metadata(project.root)
+    contract = metadata.request_auth_contract_version if metadata else None
+    if user_auth and (contract != 1 or project.server != AgentServer.MASON):
+        raise AgentCliError(
+            "User auth requires request_auth_contract_version = 1 in .mason/project.toml.",
```
junchoi-db: what does request_auth_contract_version = 1 mean ? is it per tool user scope? explain

integrations/mason/src/databricks_mason/cli/app_auth.py
```diff
@@ -0,0 +1,171 @@
...
+    name: str
+    existing_scopes: tuple[str, ...] | None
+    scopes: tuple[str, ...]
+
+
+def required_user_scopes(project: AgentProject | None) -> set[str]:
+    """Return baseline Apps scopes for request-user managed tools."""
+    if project is None:
+        return set()
+    return {
```
junchoi-db, line 78: add todo to extend this

integrations/mason/src/databricks_mason/cli/app_auth.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+    if existing is not None:
+        _validate_implicit_identity_scopes(existing)
+    configured = tuple(sorted(set(existing.user_api_scopes or []))) if existing else None
+    requested = {"ai-gateway"} if required_scopes is None else required_scopes
+    scopes = tuple(sorted({*(configured or ()), *requested}))
+    return AppAuthPlan(apps=apps, name=name, existing_scopes=configured, scopes=scopes)
+
+
+def apply_app_auth(plan: AppAuthPlan, *, instances: int | None = None, attempts: int = 12) -> None:
```
junchoi-db: this adds scope to existing app?

integrations/mason/src/databricks_mason/cli/app_auth.py
```diff
@@ -0,0 +1,171 @@
...
+            current = plan.apps.get(plan.name)
+            _validate_forwarding(current)
+            _validate_implicit_identity_scopes(current)
+            if tuple(sorted(set(current.user_api_scopes or []))) != plan.existing_scopes:
+                raise AgentCliError("Apps user scopes changed since preflight; review and retry.")
+            if plan.scopes != plan.existing_scopes or instances is not None:
+                plan.apps.create_update(plan.name, update_mask=",".join(mask), app=desired).result(
+                    timeout=timedelta(minutes=5)
+                )
+        for attempt in range(attempts):
```
junchoi-db, line 163: qq: can we use tenacity?

integrations/mason/src/databricks_mason/cli/deploy.py
```diff
@@ -482,6 +488,11 @@ def _grant_store_access(
     default=None,
     help="Number of deployment instances.",
 )
+@click.option(
+    "--adopt-user-auth",
```
junchoi-db: explain what this is -- why is this needed in deploy?

integrations/mason/src/databricks_mason/cli/deploy.py
junchoi-db: so this is when user update existing app via agent.toml and does deploy -> we update existing app?

junchoi-db: still not sure why this option is needed?

integrations/mason/src/databricks_mason/langgraph/mcp.py
```diff
@@ -13,24 +13,44 @@
...
 from databricks_mason.runtime.workspace import workspace_client, workspace_headers
 
 logger = logging.getLogger(__name__)
 
 
-def _server_from_tool(tool: ToolRecord) -> DatabricksMCPServer | None:
+def _server_from_tool(
+    tool: ToolRecord,
+    *,
+    workspace_client_for: Callable[[str], WorkspaceClient] | None = None,
```
junchoi-db: explain this callable -- what is this doing -- how is this paeed

integrations/mason/src/databricks_mason/langgraph/mcp.py
```diff
@@ -13,24 +13,44 @@
...
+) -> DatabricksMCPServer | None:
     if tool.kind not in {"sandbox", "mcp", "uc_function", "genie_one"}:
         return None
-    client = workspace_client()
+    mode = tool.auth or "app"
+    if workspace_client_for is None and mode == "user" and os.getenv("DATABRICKS_APP_NAME"):
```
junchoi-db: who sets DATABRICKS_APP_NAME ?

integrations/mason/src/databricks_mason/openai/mcp.py
```diff
@@ -15,25 +15,142 @@
...
 
-logger = logging.getLogger(__name__)
+if TYPE_CHECKING:
+    from databricks.sdk import WorkspaceClient
 
 _FRAMEWORK = "openai"
 
 
-class _DownscopedMcpServer(McpServer):
+def _auth_error(
```
junchoi-db: is this logic duplicated? integrations/mason/src/databricks_mason/langgraph/mcp.py

integrations/mason/src/databricks_mason/cli/deploy.py
junchoi-db: Yes. This path is for an existing App named by the project when agent.toml introduces request-user tools. I changed it in cecec4b so adoption is one-time: --adopt-user-auth is required only when Mason must add a missing scope; later deploys need no flag once the scope is present. The explicit opt-in remains for the first scope mutation because it changes the App user-consent contract, may require users to re-consent, and the Apps API has no compare-and-swap guarantee. New Apps are configured automatically.

integrations/mason/src/databricks_mason/cli/app_auth.py
junchoi-db: For new Apps, yes: cecec4b now explicitly creates the App with forward_user_access_token=true instead of depending on a platform default. An existing App that explicitly disabled forwarding still fails closed because changing this setting requires stopping and restarting its compute; mason deploy should not silently restart an existing App. The error tells the owner to enable forwarding and restart before retrying.

integrations/mason/src/databricks_mason/cli/app_auth.py
junchoi-db: It is a project-level template compatibility marker, not a per-tool scope. Version 1 means the generated Mason AgentApp request path owns a transient RequestAuthContext for the active attempt, passes its user/app client resolver into managed-tool adapters, never persists the token, and rejects recovery without fresh request auth. Each tool still selects auth = user or auth = app independently in agent.toml. cecec4b adds this explanation in code and the README.

integrations/mason/src/databricks_mason/cli/app_auth.py
junchoi-db: Added in cecec4b. The TODO now calls out extending required_user_scopes to return the least-privilege Apps scope for each supported request-user tool kind/service; the current baseline remains ai-gateway.

integrations/mason/src/databricks_mason/cli/app_auth.py
junchoi-db: Yes, when the required scope is missing. Preflight snapshots the existing configured scopes and forms a union with the required scopes. Apply re-reads the App to detect drift, updates only user_api_scopes plus explicitly requested instance fields, preserves unrelated scopes, and verifies configured and effective scopes before source rollout. cecec4b also makes adoption one-time: no flag is needed when the existing App already has the scope.

integrations/mason/src/databricks_mason/cli/app_auth.py
junchoi-db: We could, but I kept the bounded loop intentionally. Mason does not directly depend on tenacity, and this is one fixed propagation check: at most 12 reads with a five-second interval. Adding a runtime dependency for that does not simplify the behavior. The SDK waiter already covers the update long-running operation; this separate loop verifies effective_user_api_scopes after control-plane propagation. The bounded failure path is unit-tested.

integrations/mason/src/databricks_mason/cli/deploy.py
junchoi-db: This is in deploy because forwarding and user API scopes are Databricks App control-plane configuration, not agent source. Request-user code cannot work until that contract exists, so deploy reconciles and verifies it before stores, manifests, or source rollout; otherwise Mason could publish code that is guaranteed to fail OBO at runtime. App-only projects bypass this path. cecec4b adds the rationale next to the preflight. Verification: 1116 Mason unit tests passed, 11 skipped; Ruff and ty are clean.

integrations/mason/src/databricks_mason/langgraph/mcp.py
junchoi-db: workspace_client_for is the request-owned identity resolver. The generated runtime adapter passes RequestAuthContext.client_for when an invocation has request-user auth. The callable accepts user or app and returns either the forwarded-token WorkspaceClient or the environment-authenticated App client for that binding. cecec4b names the shared type WorkspaceClientResolver and documents this flow in runtime/mcp_auth.py.

integrations/mason/src/databricks_mason/langgraph/mcp.py
junchoi-db: Databricks Apps sets DATABRICKS_APP_NAME in deployed App compute. Mason does not set it for ordinary local execution. The adapter uses it only as a deployment boundary: if an agent.toml binding asks for user auth but no request-owned resolver was passed, deployed code fails closed instead of silently falling back to App identity. cecec4b moves and documents that check in the shared MCP auth helper.

integrations/mason/src/databricks_mason/openai/mcp.py
junchoi-db: Yes. The LangGraph and OpenAI adapters duplicated client resolution plus nested auth-error and structured tool-error classification. cecec4b moves that shared behavior into runtime/mcp_auth.py; each adapter now retains only framework-specific server construction, discovery, and sandbox downscoping. The parameterized adapter tests assert that both frameworks use the same classifiers.

integrations/mason/src/databricks_mason/cli/app_auth.py
junchoi-db: let's make it very explicit

junchoi-db: add to function name and also add to comments

junchoi-db: Made this explicit in `25ac7c5`: removed `request_auth_contract_version`, so `AgentApp` now derives request-user behavior directly from tool bindings with `auth = "user"`; explicit `auth_policy=` remains the override. The deploy path now describes exactly when an existing App needs its user API scopes updated. Focused verification: 118 passed, 2 skipped; targeted Ruff clean.

junchoi-db: Addressed in `25ac7c5`: renamed the two-phase helpers to `plan_app_user_scope_update` / `apply_app_user_scope_update`, renamed the plan type and fields around user API scopes, renamed the CLI gate to `--allow-user-scope-update`, and expanded function comments/docstrings to distinguish new-App creation from adding missing scopes to an existing App. Focused verification: 118 passed, 2 skipped; targeted Ruff clean.

integrations/mason/src/databricks_mason/openai/mcp.py
```diff
@@ -15,25 +15,77 @@
...
+
+    async def connect(self):
+        if not self._mason_request_user:
+            return await super().connect()
+        try:
+            return await super().connect()
+        except Exception as error:
+            raise _auth_error(error, self.name) or AuthError(
+                "MCP_TOOL_FAILED",
+                "Could not connect to the configured MCP service.",
```
junchoi-db, line 58: why is error handling diff depending  onself._mason_request_user?

integrations/mason/src/databricks_mason/openai/mcp.py
```diff
@@ -15,25 +15,77 @@
...
+                502,
+                self.name,
+            ) from None
+
+    async def list_tools(self, *args, **kwargs):
```
junchoi-db, line 63: same here

elainewang-db (approved):

PR #627, opened 2026-09-18
[mason] Bump to 0.1.7

elainewang-db:


shivam5:
should we do 0.1.7.dev0 or a stable release?
If it's field probably dev0 is fine?

shivam5 (approved):

PR #639, opened 2026-09-21
[mason] dev: trace to a local sqlite-backed MLflow server

jamesbxwu:
## What

`mason dev` now records traces to a **local, sqlite-backed MLflow tracking server** it launches for the run, instead of provisioning a workspace MLflow experiment. Local tracing becomes zero-setup: no workspace experiment, no auth, no username, no permissions - traces stay on the dev machine and are viewable in the local MLflow UI (its URL is printed in the startup panel).

## Why

Local tracing previously required a workspace experiment under `/Users/<user>/...`, which needs the current username resolved online plus write permissions - awkward for a purely local dev loop - and it baked MLflow env into the committed `app.yaml`. Mirroring how `mason dev` already derives its local manifest and how session/memory stores default to local in dev, tracing now defaults to a local backend.

## How

- For Mason-server projects, `mason dev` launches `uvx mlflow server` backed by `sqlite:///<project>/.mason/mlflow.db` on a free localhost port. Using `uvx` keeps it off both the (skinny) CLI env and the agent venv, and the server owns the sqlite schema (no client/server migration mismatch).
- It injects `MLFLOW_TRACKING_URI` (the local server) + a bare `MLFLOW_EXPERIMENT_NAME` (the project dir name - valid locally, no workspace path) into the **dev-only manifest** (`app.masondev.yaml`), never the deployable `app.yaml`. The agent enables tracing through the same runtime gate a deployment uses, so **no agent code differs between dev and deploy**.
- The local store persists across runs (only `app.masondev.yaml` is cleaned up), so traces accumulate in the same `<project>` experiment. The server is best-effort (dev still runs without traces if it can't start) and is torn down when dev exits.
- The startup panel's existing "Traces" line now shows the local UI URL. `.mason/` and `app.masondev.yaml` are added to the template `.gitignore`s.

`mason deploy` is unchanged - it still provisions/uses the managed workspace experiment.

## Testing

- **Unit**: reworked `dev_test.py` (local-tracing manifest wiring, Traces URL, teardown, best-effort degrade) plus new tests for the `_start_local_tracing` helper. Full mason unit suite: **929 pass**; ruff + ty clean.
- **Manual Testing**: ran mason init and mason dev and verified the local mlflow tracing working e2e with UI
```
 mason init mlflow-local-tracing --framework openai

╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ✔ Scaffolded 'agent-openai'                                                                                                                                                              │
│ Framework           openai                                                                                                                                                               │
│ Server              Mason AgentApp                                                                                                                                                       │
│ Template ref        bundled (databricks-mason 0.2.0)                                                                                                                                     │
│ Directory           mlflow-local-tracing                                                                                                                                                 │
│ Chat app            enabled                                                                                                                                                              │
│ Memory store        mlflow-local-tracing-ajoavz-memory                                                                                                                                   │
│ Session store       mlflow-local-tracing-ajoavz-sessions                                                                                                                                 │
│ Traces experiment   /Shared/mason_traces/mlflow-local-tracing-ajoavz                                                                                                                     │
│ Profile (.env)      DEFAULT                                                                                                                                                              │
│ Next steps                                                                                                                                                                               │
│ cd mlflow-local-tracing                     Enter the project directory                                                                                                                  │
│ mason dev                                   Run the agent locally                                                                                                                        │
│ Open http://localhost:8000 to chat with it                                                                                                                                               │
│ mason deploy mlflow-local-tracing           Deploy it to Databricks (from the project dir)                                                                                               │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯


mason dev
⚠ Memory store 'mlflow-local-tracing-ajoavz-memory' is declared but not created yet — long-term memory is disabled locally. Run `mason deploy` to create it.
⚠ Session store 'mlflow-local-tracing-ajoavz-sessions' is declared but not created yet — conversation history is in-memory (not durable). Run `mason deploy` to create it.

╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ✔ Starting agent                                                                                                                                                                         │
│ Chat UI   http://localhost:8000                                                                                                                                                          │
│ Traces    http://127.0.0.1:59441                                                                                                                                                         │
│ Next steps                                                                                                                                                                               │
│ Open http://localhost:8000 to chat with your agent                                                                                                                                       │
│ mason tools add mcp <service>                       Give the agent a tool                                                                                                                │
│ mason memory bind <store>                           Attach a memory / session store                                                                                                      │
│ mason deploy mlflow-local-tracing                   Deploy it to Databricks                                                                                                              │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```
Screenshot 2026-09-21 at 10 15 11 AM
Screenshot 2026-09-21 at 10 15 19 AM


integrations/mason/src/databricks_mason/cli/dev.py
```diff
@@ -155,28 +151,20 @@ def dev(
...
-    except Exception as exc:  # noqa: BLE001 - tracing must never block a local run
-        render.diagnostic(
-            "warning",
-            f"tracing not enabled — {exc}",
-            help="running without traces; set it up later with `mason tracing configure`",
-        )
-    if env_updates:
-        _upsert_manifest_env(source_dir, env_updates)
+    if project is not None and project.server == AgentServer.MASON:
+        tracing_server, tracing_env = start_local_tracing_server(source_dir)
```
jamesbxwu: why gate this to only mason server and not custom server? We should set up the MLFlow tracking server regardless.

jamesbxwu: Good call — done in 11d5d84. Removed the `project.server == MASON` gate so `mason dev` starts the local MLflow tracking server and injects its env for every project (any framework/server, and even a raw dir). The server setup is orthogonal to whether the agent uses it: an agent that traces via MLflow (autolog or `start_trace`) picks it up, and it's a harmless no-op for one that doesn't. Added a custom-server test to lock it in.

integrations/mason/src/databricks_mason/cli/dev.py
```diff
@@ -155,28 +151,23 @@ def dev(
...
-    # local run.
+    # Local tracing: start a local MLflow tracking server backed by sqlite under .mason/ and point the
+    # agent at it via the dev-only manifest — for any project, regardless of framework/server. An agent
+    # that uses MLflow (autolog or `start_trace`) then traces to it; it's harmless for one that doesn't.
+    # Traces stay on the machine — no workspace experiment, no auth, no username needed — and the same
+    # server serves the trace UI. Launched via `uvx mlflow` so it needs neither the (skinny) CLI env nor
+    # the agent venv, and it owns the sqlite schema (so there's no client/server migration mismatch).
+    # Best-effort: any launch failure degrades to running without traces. `mason deploy` handles the
+    # managed workspace experiment instead.
+    tracing_server, tracing_env = start_local_tracing_server(source_dir)
```
elainewang-db, line 164: qq: does `mason dev` respect `mason tracing disable`? should we start the local tracing server in that case?

integrations/mason/src/databricks_mason/cli/tracing.py
```diff
@@ -146,6 +147,86 @@ def _trace_to_json(trace: Any) -> dict:
...
+                stderr=subprocess.STDOUT,
+            )
+    except OSError as exc:
+        render.diagnostic(
+            "warning",
+            f"local tracing not started — {exc}",
+            help="running without traces (is `uv` installed?)",
+        )
+        return None, {}
+    return server, {
```
elainewang-db, line 214: do we know that the server is up and ready to serve requests immediately here?

elainewang-db (commented):
can we make sure integrations/mason/cli.md is updated with the changed behavior?

integrations/mason/src/databricks_mason/cli/tracing.py
jamesbxwu: in my local testing, it was up immediately and able to take requests

integrations/mason/src/databricks_mason/cli/dev.py
jamesbxwu: good question, I have a follow up PR to unify the behavior between tracing, session and memory. The flow would be as follows:

1. `mason init` -> default experiment name is written to `agent.toml`
2. `mason dev` -> tracing runs locally since the default experiment name is not provisioned yet on the target workspace
3. `mason xxx bind` -> updates `agent.toml` but does not provision anything
4. `mason deploy` -> if experiment name is present in `agent.toml`, provision the experiment
5. `mason dev` -> the experiment is now provisioned, start exporting to it instead of local (consistent with session/memory)
6. `mason xxx unbind` and then `mason dev` -> fall back to local for tracing (consistent with session/memory)

This means tracing is default on for `mason dev` and we will only start the local mlflow server when no workspace experiment is provisioned. 

I plan to rename `mason tracing configure/disable` to `mason tracing bind/unbind` for consistency. 

For this PR , it is aligned such that `mason tracing disable` only affects `mason deploy` and not `mason dev`. I've updated the cli.md accordingly

integrations/mason/src/databricks_mason/cli/dev.py
```diff
@@ -155,28 +153,23 @@ def dev(
...
```
elainewang-db, line 164: should we wrap all steps after server start with the try/finally to ensure cleanup?

jamesbxwu: Good call — done in 8427746. Moved the `try/finally` up to right after `start_local_tracing_server`, so every step after the server starts (building the dev manifest via `_dev_entry_point`, which parses app.yaml and can raise; announcing the URL; run-local) is covered and the `finally` always stops the server + removes the dev-only manifest. Previously a failure before run-local (e.g. a malformed app.yaml) orphaned the local MLflow server. Added `test_dev_stops_local_tracing_server_when_setup_fails` to lock it in.

elainewang-db (approved):

PR #640, opened 2026-09-21
[mason] tracing: bind by experiment name (not a workspace-local id) across the CLI

jamesbxwu:
## What

Adopt the MLflow experiment **name** (portable across workspaces) as the tracing identifier
everywhere in the `mason` CLI, replacing the workspace-local experiment **id**.  The issue with using the id is that a profile switch via `mason login` will lead to failure due to the `experiment_id` being a workspace bound entity. 

- **agent.toml / `mason init`**: stores/writes `[tracing] experiment_name` (a `/Shared` path), never an id.
- **`mason deploy`**: get-or-creates that named experiment in the active workspace, resolves its id,
  wires `MLFLOW_EXPERIMENT_ID` + grants the SP. agent.toml is never mutated. Fixes the stale-id bug
  (a pinned id broke after re-login to a different-workspace profile; a name just get-or-creates the
  same-named experiment in the new workspace).
- **`mason tracing configure`**: `--experiment-name` (primary) + `--experiment-id` (convenience,
  resolved to a name and stored as a name), mutually exclusive. A not-yet-created name is accepted
  (deploy creates it); a missing/UC-backed **id** is rejected.
- **`mason tracing list`/`get`**: `--experiment-name` + `--experiment-id` (mutually exclusive),
  replacing `list`'s id-only `--experiment`. An explicit identifier must exist - an unknown name or
  id errors ("No MLflow experiment named/found ...") rather than silently listing nothing, so a typo
  isn't mistaken for an empty experiment. With neither flag, reads this project's workspace experiment
  if provisioned, else the local `mason dev` store (`.mason/mlflow.db`, tagged `(local dev)`), so a
  not-yet-deployed dev run's traces are still inspectable. Adds `alembic` (lets mlflow-skinny open the
  sqlite store).
- **`mason dev`**: strips the deploy-written workspace tracing env (`MLFLOW_TRACKING_URI` /
  `MLFLOW_EXPERIMENT_ID` / `MLFLOW_TRACING_DESTINATION`) from the dev-only manifest, so a stale
  workspace id (which MLflow resolves ahead of the name) can't point local dev tracing at an
  experiment that doesn't exist on the local sqlite server.

## Out of scope (follow-up)

Unify the `mason dev`, `mason xxx bind` and `mason deploy` across session, memory and tracing

## Test

`ruff`, `ty`, and the full mason unit suite pass. Unit coverage for list/get/configure spans explicit
id, explicit name→id, mutual exclusion, error-on-missing (name + id), workspace default, and the
local-dev-store fallback.

### Manual verification (live workspace, langgraph agent)

Ran the full lifecycle end to end. Steps 9-12 required freeing workspace app quota (an environment
limit, unrelated to this change); step 7 initially surfaced the `--experiment-id` raw-exception bug,
which is fixed in this PR and re-verified.

| # | Command | Result |
|---|---------|--------|
| 1 | `mason init --framework langgraph --disable-chat-app <dir>` | agent.toml wrote `[tracing] experiment_name = /Shared/mason_traces/<slug>` (a name, never an id) |
| 2 | `mason dev` | Local sqlite MLflow server started (`Traces http://127.0.0.1:<port>`), agent served |
| 3 | `curl POST /api/invocations` (local) | Agent responded; `.mason/mlflow.db` written |
| 4 | `mason tracing list` | Read the **local** store; title tagged `(local dev)`; the trace listed |
| 5 | `mason tracing get <id>` | Read the local trace (status / latency / spans / request / response) |
| 6 | `mason tracing list --experiment-name /Shared/nope` | Clean error `No MLflow experiment named '…' in this workspace` |
| 7 | `mason tracing configure --experiment-id <bad>` | Clean error `No MLflow experiment found with id '…'` |
| 8 | `mason tracing list --experiment-id <bad>` | Clean error (same) |
| 9 | `mason deploy <name>` | Created memory + session stores; provisioned the experiment **by name** → id; wired `MLFLOW_EXPERIMENT_ID` into `app.yaml`; deployed the app; granted the SP store + trace access |
| 10 | `mason endpoint invoke <app> --path /api/invocations` (deployed) | Deployed agent responded (`completed`) |
| 11 | `mason tracing list` | Now read the **workspace** experiment (no `(local dev)` tag); the deployed trace listed |
| 12 | `mason tracing get <id>`, `list --experiment-name <path>`, `list --experiment-id <id>` | All three read the deployed trace from the workspace |
| 13 | `mason dev` (after deploy) | `app.masondev.yaml` **stripped** the deploy-written `MLFLOW_EXPERIMENT_ID` + `databricks` URI, and re-added local `MLFLOW_TRACKING_URI` (`http://127.0.0.1:<port>`) + `MLFLOW_EXPERIMENT_NAME` |
| 14 | `mason tracing disable` → `mason dev` | `disabled = true` written; dev **still traced locally** (deploy-only semantics) |
| 15 | cleanup | app, both stores, and the experiment deleted |

## Breaking change (experimental)

Tracing config in `agent.toml` moves from `[tracing] experiment_id` to `[tracing] experiment_name`, and
`AgentProject.load` reads only `experiment_name`. A project that ran a pre-this-PR `mason deploy` (which
pinned `experiment_id` into `agent.toml`) will, after upgrading, fall back to the default experiment and
bind a fresh one on the next deploy - its earlier traces stay in the old experiment. The default name
also moved (`/Users/<you>/mason-traces/<project>` -> `/Shared/mason_traces/<project>`).

Mason is experimental (pre-GA; the CLI and config may change in backward-incompatible ways), so this is
an accepted break rather than a silent-migration bug: after upgrading, re-bind the experiment you want
with `mason tracing configure` (surfaced by Isaac review as a MAJOR; accepted and documented here).


integrations/mason/src/databricks_mason/cli/deploy.py
```diff
@@ -723,7 +719,13 @@ def deploy(
             f"Cause: {grant_error}",
         )
     if trace_setup_error is not None:
-        steps.insert(0, f"Tracing wasn't set up (deployed without it). Cause: {trace_setup_error}")
+        steps.insert(
+            0,
+            "Tracing wasn't set up (deployed without it) — the experiment couldn't be created "
+            "(is /Shared writable in this workspace?). Point at a writable experiment with "
+            "`mason tracing configure --experiment-name <path>` (e.g. under your /Users/<you>/…) and "
+            f"redeploy. Cause: {trace_setup_error}",
```
jamesbxwu: this is too verbose.  We shouldn't assume it's related to permission error with /shared.  Let's keep the old msg and assume the actionable msg will be in `trace_setup_error`

integrations/mason/src/databricks_mason/cli/help.py
```diff
@@ -208,14 +208,14 @@
...
+        ("mason tracing configure --experiment-id 12345", "trace to a specific experiment id"),
         ("mason tracing disable", "turn tracing off"),
     ),
     ("tracing", "configure"): (
-        ("mason tracing configure --experiment 12345", "trace to a specific experiment id"),
+        ("mason tracing configure --experiment-id 12345", "trace to a specific experiment id"),
     ),
     ("tracing", "disable"): (("mason tracing disable", "turn tracing off"),),
-    ("tracing", "list"): (("mason tracing list --experiment 12345", "list recent traces"),),
+    ("tracing", "list"): (("mason tracing list --experiment-id 12345", "list recent traces"),),
```
jamesbxwu: shouldn't we show the --experiment-name variant too?

integrations/mason/src/databricks_mason/cli/deploy.py
```diff
@@ -370,19 +370,22 @@ def _reconcile_declared_stores(
...
-
-    Shared by `mason dev` and `mason deploy` so both trace to the same experiment for a given project.
+    - otherwise -> the project's bound ``experiment_name`` (written into agent.toml at `mason init`),
+      else the default ``/Shared/mason_traces/<project>`` -> get-or-create it in the workspace the
+      current profile targets, and return that workspace's id.
+
+    Nothing is written back to agent.toml (no id is pinned), so re-deploying under a
+    different-workspace profile just get-or-creates the same-named experiment there. Raises if the
+    experiment can't be created (e.g. `/Shared` isn't writable); the caller degrades to deploying
+    without tracing and points the user at `mason tracing configure`.
```
jamesbxwu: you are bleeding too much implementation detail of other components such as mason init and mason dev here. Keep docstring specific to the input params and do not make assumptions about other components.

integrations/mason/src/databricks_mason/cli/deploy.py
```diff
@@ -392,17 +395,10 @@ def resolve_trace_experiment_id(
...
-        profile, client, default_experiment_name(client.current_user, project_name)
-    )
-    # Pin the resolved default so subsequent runs reuse it by id (removes the special-cased "recompute
-    # the default" path). No agent.toml (raw dir) just means nowhere to pin — still trace this run.
-    if project is not None and project.configure_tracing(experiment_id):
-        project.write()
-    return experiment_id
+    name = project.trace_experiment_name if project is not None else None
+    if not name:
+        name = default_experiment_name(project_name)
```
jamesbxwu, line 393: add a TODO that this fall back will go away when we consolidate tracing/session/memory s.t. deploy only reads from what's configured in `agent.toml`

integrations/mason/src/databricks_mason/cli/init.py
```diff
@@ -385,19 +386,24 @@ def init(
...
             # Store display names aren't unique and projects often share a directory name, so a
-            # per-scaffold token keeps default stores from colliding. Memory and session share the
-            # token; an explicit --memory/session-store wins.
+            # per-scaffold token keeps default stores from colliding. Memory, session, and the default
+            # tracing experiment share the token; an explicit --memory/session-store wins.
             token = "".join(secrets.choice(string.ascii_lowercase) for _ in range(6))
             memory_store = memory_store or default_store_name(dest.name, "memory", token)
             session_store = session_store or default_store_name(dest.name, "sessions", token)
+            # A default /Shared experiment name (username-free, so this stays offline); `mason deploy`
+            # get-or-creates it in the target workspace. Stored as a name, not an id, so it's portable.
```
jamesbxwu: you shouldn't know here that the default experiment name is in `/Shared`
```suggestion
```

integrations/mason/src/databricks_mason/cli/tracing.py
```diff
@@ -99,24 +134,88 @@ def create_experiment_idempotent(profile: Optional[str], client, name: str) -> s
...
-    return experiment.experiment_id if experiment else None
+    if project is not None and project.trace_disabled:
+        return None
+    name = project.trace_experiment_name if project is not None else None
+    return name or default_experiment_name(pathlib.Path(source).resolve().name)
+
+
+def _trace_read_target(
+    source: pathlib.Path | str, profile: Optional[str]
+) -> tuple[Optional[str], Optional[str]]:
```
jamesbxwu: document the return type tuple? Better yet, define a python dataclass for readability

integrations/mason/src/databricks_mason/cli/tracing.py
```diff
@@ -99,24 +134,88 @@ def create_experiment_idempotent(profile: Optional[str], client, name: str) -> s
...
+    if db.exists():
+        uri = f"sqlite:///{db}"
+        mlflow.set_tracking_uri(uri)
+        local = mlflow.get_experiment_by_name(pathlib.Path(source).resolve().name)
+        if local is not None:
+            return uri, local.experiment_id
+    return None, None
+
+
+def _explicit_experiment_target(
```
jamesbxwu: ### 
```suggestion
def _workspace_experiment_target(
```

integrations/mason/src/databricks_mason/cli/tracing.py
```diff
@@ -99,24 +135,90 @@ def create_experiment_idempotent(profile: Optional[str], client, name: str) -> s
...
         project = AgentProject.load(source)
     except AgentCliError:
         project = None
-    if project is not None and project.trace_experiment_id:
-        return project.trace_experiment_id
-    client = obj.client()
-    name = default_experiment_name(client.current_user, pathlib.Path(source).resolve().name)
-    experiment = mlflow.get_experiment_by_name(name)
-    return experiment.experiment_id if experiment else None
+    if project is not None and project.trace_disabled:
```
elainewang-db, line 162: does `_resolve_experiment_name`  also serve `mason tracing list/get`? if so, should we ignore `project.trace_disabled` on read resolution so that list and get still work on local traces?

integrations/mason/src/databricks_mason/cli/tracing.py
elainewang-db, line 136: should we check to make sure the resolved experiment is not uc backed here? in case an experiment is set without going through `mason tracing configure`?

integrations/mason/src/databricks_mason/cli/tracing.py
jamesbxwu: Good catch - `_resolve_experiment_name` does feed `mason tracing list`/`get`, and on this PR a disabled project short-circuits reads, so local dev traces become invisible once tracing is disabled.

The stacked #643 cleans this up more fundamentally: it drops the `disabled` flag entirely (presence of `experiment_name` is the on/off switch) and makes `list`/`get` always fall through to the local `.mason/mlflow.db` store regardless of binding, so local dev traces are always readable. Since #643 rewrites this exact path, would it be OK to defer the fix to that PR rather than patch it here?

integrations/mason/src/databricks_mason/cli/tracing.py
jamesbxwu: Good call - done in 995101d. `mason tracing configure`/`bind` already rejects a UC-backed experiment, but a hand-edited `agent.toml` could name one directly and slip past that check, so `create_experiment_idempotent` (the deploy provisioning path) now re-checks the resolved experiment and refuses a UC-backed one with the same error. Added a unit test for it.

integrations/mason/src/databricks_mason/cli/tracing.py
elainewang-db: defering sgtm!

elainewang-db (approved):

PR #642, opened 2026-09-21
[mason] Back the chat UI's model picker with the Unity Catalog AI Gateway

ankit-db:
## Summary

The chat UI's model picker listed Model Serving endpoints, so it could only offer models a workspace had already deployed. This points it at the Unity Catalog AI Gateway instead, which exposes Databricks-managed models as `system.ai.*` model services that every workspace can reach.

- New `databricks_mason.runtime.models.list_ai_gateway_models` lists the chat-capable `system.ai` model services from `GET /api/2.1/unity-catalog/model-services?parent=schemas/system.ai` — embeddings-only services filtered out, paged, transient list failures retried. It is framework-neutral and re-exported lazily from `databricks_mason` and `databricks_mason.runtime`, so the light CLI import path stays free of the agent stack.
- Both templates' `runtime/ui.py` discover through it, replacing the duplicated `serving_endpoints.list()` plumbing (net -112 lines).
- Both agents call the chosen model with `use_ai_gateway=True`, routing to `<host>/ai-gateway/mlflow/v1`.
- The templates' default `MODEL` becomes `system.ai.claude-sonnet-4-5`. This is required, not a preference: the gateway resolves `system.ai.*` names, so a serving-endpoint default would no longer resolve. Picking a different default is a one-line change if another model is preferred.
- `/api/demo/models` keeps its `{default, available}` response shape, so **no frontend changes** — the existing `#model-select` / `renderModels()` keep working. Ranking simplifies to default-first-then-alphabetical since the gateway returns a single namespace; the `_MODEL_LIMIT` cap still never drops the default.
- `CHAT_APP.md` updated for both templates.

This supersedes the picker half of #547, which cannot be rebased: it predates the template move under `src/databricks_mason/templates/`, the chat UI facelift (#611), and the runtime adapter split (#590). The reasoning/"Thinking" UI from that branch is not included here and remains to be ported separately.

## Test plan

- [x] `pytest tests/unit_tests/` — 946 passed, including 6 new tests for `list_ai_gateway_models` (paging, embeddings filtering, retry, error propagation)
- [x] `pytest tests/functional/` against the built wheel — 36 passed, confirming `mason dev` still gets a model reply end to end after the URL change
- [x] Scaffolded both templates from these sources and ran their own suites — 39 passed each, covering the rewritten picker tests (default pinned, dedup, cap, graceful fallback when `system.ai` is unreadable)
- [x] `ruff check`, `ruff format --check`, `ty check`
- [x] Verified `runtime/models.py` ships in the wheel and that the lazy export resolves without importing `mlflow`
- [ ] Try the picker against a live workspace to confirm `system.ai` listing and a gateway-routed turn


integrations/mason/src/databricks_mason/runtime/__init__.py
```diff
@@ -31,6 +32,8 @@
     # Workspace SDK client construction (account-host / run-local routing handled).
     "workspace_client",
     "workspace_headers",
+    # Unity Catalog AI Gateway model discovery (system.ai chat models, for the demo UI's picker).
+    "list_ai_gateway_models",
```
akansal2: nit: shall we use "model services" instead of "models" everywhere instead?

integrations/mason/src/databricks_mason/runtime/models.py
```diff
@@ -0,0 +1,89 @@
...
+        try:
+            return client.api_client.do("GET", _MODEL_SERVICES_PATH, query=query)
+        except Exception as exc:  # noqa: BLE001 - retry any list failure, then surface the last one
+            last_error = exc
+            if attempt < _LIST_ATTEMPTS - 1:
+                time.sleep(_RETRY_DELAY_S)
+    raise last_error if last_error else RuntimeError("model-services list returned nothing")
+
+
+def list_ai_gateway_models(client: WorkspaceClient) -> list[str]:
```
akansal2: ditto - `list_ai_gateway_model_services`? and update all the other sites as well?

integrations/mason/src/databricks_mason/templates/ui/agent-langgraph/tests/test_demo_ui.py
```diff
@@ -275,14 +275,14 @@ def test_demo_config_does_not_wait_for_model_discovery(monkeypatch):
     monkeypatch.setattr(
         ui,
         "_discover_chat_models",
-        lambda: calls.append(True) or ["databricks-gpt-5-2", "databricks-gpt-5-5"],
+        lambda: calls.append(True) or ["system.ai.claude-sonnet-4-5", "system.ai.another-chat"],
```
akansal2: should we use a real model service name instead of "another-chat"? same for rest of this file's edits

integrations/mason/tests/unit_tests/runtime_model_services_test.py
```diff
@@ -0,0 +1,121 @@
...
+    result = list_ai_gateway_models(cast(Any, client))
+    assert result == ["system.ai.a", "system.ai.b"]
+    # First call has no page_token; the second carries the token from page one.
+    assert client.api_client.calls[0].get("parent") == "schemas/system.ai"
+    assert "page_token" not in client.api_client.calls[0]
+    assert client.api_client.calls[1].get("page_token") == "1"
+
+
+def test_empty_listing():
+    assert _list([{"model_services": []}]) == []
```
akansal2, line 93: if the list of model services is empty (for ex. the user doesn't have permissions on any model services, or the workspace is CSP-compliant or any other reason to not have any model services - do we want to include logic to instead show the FMAPI list of "databricks-..." endpoints? iirc this is what AI Gateway does right now on their list page. this may also depend on when the AI Gateway v2/FMAPI PPT endpoint deprecation is scheduled to be + not an urgent ask since our product is only going to PrPr first

akansal2 (commented):
few comments, thanks for making the changes!

integrations/mason/src/databricks_mason/runtime/models.py
ankit-db: Done — renamed to `list_ai_gateway_model_services`, and the module with it (`runtime/models.py` → `runtime/model_services.py`). The module rename has a second benefit: `runtime/models.py` sat awkwardly next to the unrelated `databricks_mason/models.py` (the typed API-response wrappers), and that ambiguity is now gone.

All call sites updated: both `__init__.py` re-exports, both templates' `runtime/ui.py`, the tests, and `CHAT_APP.md`.

integrations/mason/src/databricks_mason/runtime/__init__.py
ankit-db: Switched the discovery side to "model services" throughout — the helper, the module, and this export comment.

I did stop short of renaming the picker's own surface, and want to flag that explicitly in case you disagree: `/api/demo/models`, the `models` key in `/api/ui/config`, `_discover_chat_models`, and the `model` field in the invocation body all stay as they are. Two reasons — those name the model the agent actually calls (the string handed to the OpenAI-compatible API), not the Unity Catalog resource; and that shape is already consumed by the shipped frontend, so renaming it would mean touching `app.js` and the config contract for no functional gain. Happy to go further if you would rather the whole path speak UC's vocabulary.

integrations/mason/src/databricks_mason/templates/ui/agent-langgraph/tests/test_demo_ui.py
ankit-db: Done. Fixtures now use real service names — `system.ai.claude-opus-4-8` and `system.ai.llama-4-maverick` (the latter already appears in the dspy integration), alongside the templates' default `system.ai.claude-sonnet-4-5` and `system.ai.gte-large` for the embeddings-filtering case. I applied the same pass to `runtime_model_services_test.py`, which had `system.ai.a`/`system.ai.b` placeholders in the paging and sorting tests.

One deliberate exception: the cap test still generates `system.ai.test-model-000…029`, since it needs more services than the 20-slot cap and no real catalog is that long. I left a comment there saying so.

integrations/mason/tests/unit_tests/runtime_model_services_test.py
ankit-db: Not doing this here, per your "not urgent" — but I want to write down the gap precisely so it does not get lost.

Current behaviour on an empty or unreadable list: `_discover_chat_models` catches the failure, `_rank_models` still pins the default, so the picker renders exactly one option and `app.js` disables the select (`available.length <= 1`). So the UI degrades rather than breaking. The sharper problem is the one you are pointing at: in a workspace with no `system.ai` model services, that single remaining option is the template default, which is itself a `system.ai.*` name and would not resolve through the gateway either. So "picker shows the default" is not actually a working fallback there — the turn would fail at call time, not at discovery.

A real fix is what you describe: on an empty result, fall back to listing `databricks-*` FMAPI endpoints and route those without `use_ai_gateway`, which means the fallback has to carry the routing choice per model rather than just the name. That is more than a listing change, and it depends on the AI Gateway v2 / PPT deprecation timing you mention, so it seems right to decide it after PrPr rather than guess now.

Want me to file an issue capturing this so it is tracked against the deprecation timeline?

ankit-db: Yeah, I agree with the agent mostly - can defer this

integrations/mason/src/databricks_mason/templates/agent-langgraph/agent/agent.py
```diff
@@ -21,7 +21,10 @@
     thread_config,
 )
 
-MODEL = "databricks-gpt-5-2"
+# A Unity Catalog AI Gateway model, served from the `system.ai` schema and queried through the
+# gateway (see `use_ai_gateway=True` below). Swap for any `system.ai.*` model your workspace
+# exposes — the demo chat app's picker lists what's available.
+MODEL = "system.ai.claude-sonnet-4-5"
```
akansal2, line 27: shall we update this section with `model service` instead of `model` as well?

integrations/mason/src/databricks_mason/templates/agent-openai/agent/agent.py
```diff
@@ -25,7 +25,10 @@
 
 logger = logging.getLogger(__name__)
 
-MODEL = "databricks-gpt-5-2"
```
akansal2, line 28: same

integrations/mason/src/databricks_mason/templates/ui/agent-langgraph/CHAT_APP.md
```diff
@@ -20,17 +20,19 @@ entries (searchable, filterable by actor, with a modal for each entry); the othe
...
-`llm/v1/chat` task), with `agent.agent.MODEL` pinned as the default. Each request sends the selected
-endpoint as `model` in the invocation body; the agent is rebuilt per turn, so the picker changes the
-model for the next turn without a restart. Discovery is best-effort: if listing is unavailable (e.g.
-no permission), the picker falls back to just the default. Omitting `model` uses `MODEL`.
-
-A large workspace can expose thousands of endpoints, so the picker is capped (`_MODEL_LIMIT`, 20)
-and ranked — default first, then `databricks-*` foundation models, then other custom/external chat
-endpoints — so the cap keeps the canonical choices. The list call is unpaginated and occasionally
-fails on big workspaces, so it is retried a few times before falling back to the default.
+The composer's model picker lists the workspace's Unity Catalog AI Gateway chat models — the
```
akansal2:
```suggestion
The composer's model picker lists the workspace's Unity Catalog AI Gateway chat model services — the
```

integrations/mason/src/databricks_mason/templates/ui/agent-openai/CHAT_APP.md
```diff
@@ -20,17 +20,19 @@ entries (searchable, filterable by actor, with a modal for each entry); the othe
...
-`llm/v1/chat` task), with `agent.agent.MODEL` pinned as the default. Each request sends the selected
-endpoint as `model` in the invocation body; the agent is rebuilt per turn, so the picker changes the
-model for the next turn without a restart. Discovery is best-effort: if listing is unavailable (e.g.
-no permission), the picker falls back to just the default. Omitting `model` uses `MODEL`.
-
-A large workspace can expose thousands of endpoints, so the picker is capped (`_MODEL_LIMIT`, 20)
-and ranked — default first, then `databricks-*` foundation models, then other custom/external chat
-endpoints — so the cap keeps the canonical choices. The list call is unpaginated and occasionally
-fails on big workspaces, so it is retried a few times before falling back to the default.
+The composer's model picker lists the workspace's Unity Catalog AI Gateway chat models — the
```
akansal2: same

integrations/mason/src/databricks_mason/templates/agent-langgraph/agent/agent.py
ankit-db: Done in 5b9c004 — the comment now reads "A Unity Catalog AI Gateway model service, served from the `system.ai` schema … Swap for any `system.ai.*` model service your workspace exposes."

integrations/mason/src/databricks_mason/templates/agent-openai/agent/agent.py
ankit-db: Done in 5b9c004 — same wording applied to the openai template.

integrations/mason/src/databricks_mason/templates/ui/agent-langgraph/CHAT_APP.md
ankit-db: Applied in 5b9c004. I also updated the matching line lower in the section — "so `MODEL` is a `system.ai.*` model service name" — so the doc is consistent. Left the `model` field / `/api/demo/models` references as-is, since those name the model the agent calls (the frontend contract), per the thread above.

integrations/mason/src/databricks_mason/templates/ui/agent-openai/CHAT_APP.md
ankit-db: Done in 5b9c004 — same change applied to the openai template's CHAT_APP.md.

integrations/mason/tests/unit_tests/runtime_model_services_test.py
ankit-db: Thanks. Leaving the behaviour as-is for this PR then. I can file a tracking issue with the note above (empty-catalog fallback to `databricks-*` FMAPI endpoints, gated on the AI Gateway v2 / PPT deprecation timing) so it is not lost after PrPr — say the word and I will open it and link it here.

akansal2 (approved):
lgtm, thank you!

PR #644, opened 2026-09-22
[mason] Onboard existing OpenAI Agents SDK agents via mason init --existing

ankit-db:
Enables `mason init --existing` to onboard **OpenAI Agents SDK** agents, not just LangGraph.

- Removes the LangGraph-only guard; the migration bundle (contract, skill, reference project) now follows the chosen framework.
- Adds an OpenAI `MASON_CONTRACT.md` and makes the shared `mason-migrate` skill framework-aware.
- Drive-by fix: caps the model-services list `page_size` at 100 (the API max), unbreaking the chat UI's model picker.

Verified with unit tests and an end-to-end migration of a real OpenAI agent, including a live gateway call.


integrations/mason/src/databricks_mason/cli/init.py
```diff
@@ -268,8 +274,8 @@ def _prepare_migration(
 @click.option(
     "--existing",
     is_flag=True,
-    help="Prepare a coding-agent migration bundle for an existing LangGraph project "
-    "(defaults to .).",
+    help="Prepare a coding-agent migration bundle for an existing LangGraph or OpenAI "
```
jamesbxwu: instead of mentioning `LangGraph or OpenAI` here and have to keep adding to this list, can we just print out the items from `_FRAMEWORK_LABELS`?

integrations/mason/src/databricks_mason/cli/init.py
```diff
@@ -374,8 +381,8 @@ def init(
     if dest.exists():
         raise AgentCliError(
             f"Destination '{dest}' already exists.",
-            hint="Use --existing --framework langgraph to prepare a migration, "
-            "or choose a new directory to scaffold.",
+            hint="Use --existing to prepare a migration from an existing LangGraph or OpenAI "
```
jamesbxwu: same common as before

integrations/mason/src/databricks_mason/cli/init.py
```diff
@@ -356,13 +362,14 @@ def init(
     template_name = template.mason_server if mason_server else template.custom_server
     chat_app_enabled = mason_server and not disable_chat_app
     if existing:
-        if selected_framework != "langgraph" or not mason_server:
+        if not mason_server:
             raise click.UsageError(
-                "--existing currently supports --framework langgraph --server mason"
+                "--existing requires --server mason (supported frameworks: langgraph, openai)."
```
jamesbxwu: i thought we had kept a list of support framework as an enum somewhere, can we use that directly?

jamesbxwu (approved):
LGTM overall, left a few minor nits. Feel free to merge once they are addressed

integrations/mason/src/databricks_mason/cli/init.py
ankit-db: Good call — done in f6e95b9. The `--existing` help now renders from `_FRAMEWORK_LABELS` via an `_or_join` helper (`_FRAMEWORK_LABEL_PHRASE`), so the list grows automatically and never needs manual edits.

integrations/mason/src/databricks_mason/cli/init.py
ankit-db: Fixed here too in f6e95b9 — the hint now uses the same `_FRAMEWORK_LABEL_PHRASE`.

integrations/mason/src/databricks_mason/cli/init.py
ankit-db: Yep, `AgentFramework` — the guard message now derives from it: `_FRAMEWORK_VALUE_PHRASE = ", ".join(f.value for f in AgentFramework)` (f6e95b9).

